PR #1, opened 2017-08-15
adding changes to make new instances independent

abit2:
@DanRyanIrish @Cadair 
new arguements gave error in creation of new instances

### DIFF
--- a/ndcube/cube_utils.py
+++ b/ndcube/cube_utils.py
@@ -79,8 +79,8 @@ def get_cube_from_sequence(cubesequence, item):
                     data[i] = cube[item[2]]
                 else:
                     data[i] = cube[item[2::]]
-            result = cubesequence._new_instance(
-                data, meta=cubesequence.meta, common_axis=cubesequence.common_axis)
+            result = deepcopy(cubesequence)
+            result.data = data
     return result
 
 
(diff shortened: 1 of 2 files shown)


DanRyanIrish:
Wait, I have a question.  The docstring says this function handles the ```NDCubeSequence.__getitem__```.  Doesn't that mean it should be a shallow copy?  Since we want the actually ```NDCubeSequence```/```NDCube``` in the parent ```NDCubeSequence``` to be returned?  

Just trying to remember what we wanted from this function.

ndcube/cube_utils.py
```diff
@@ -47,27 +47,35 @@ def get_cube_from_sequence(cubesequence, item):
...
+        elif isinstance(item[0], slice) and type_slice is None:
+            if len(item[1::]) is 1:
+                result.data = result.data[item[0]]
+                for i, data in enumerate(result.data):
+                    result.data[i] = data[item[1]]
+            else:
+                result.data = result.data[item[0]]
+                for i, data in enumerate(result.data):
+                    result.data[i] = data[item[1::]]
+        elif isinstance(item[0], slice) and type_slice is 'sequence_as_cube':
```
abit2: using type slice to make the getitem more understandable and this if is used for the unique purpose

DanRyanIrish: Add that as a comment in the code.

PR #3, opened 2017-08-16
changing name coords to extra_coords and adding tests for extra_coords slicing

abit2:
@DanRyanIrish @Cadair 
tests for extra_coords slicing :)


DanRyanIrish:
Looking good @abit2.  Two comments so far.
1. We want the extra_coords attribute hidden.  So it should be ```_extra_coords```, not ```extra_coords```.
2. Why don't you take the block of code in each test where you check the _extra_coords are as expected, and turn it into a function and import like you did with ```assert_wcs_are_equal```.  This time you could put it in cube_utils and call it something like ```assert_extra_coords_equal```.  This will make it easier to maintain and to change in the future.

PR #14, opened 2017-09-22
Bugfix: Consistent NDCubeSequence slicing

DanRyanIrish:
**Problem fixed by this PR**:
Consider slicing an ```NDCubeSequence``` via ```NDCubeSequence.index_as_cube``` to get a (N-1)D slice from a single constituent ```NDCube```.  If the slice came from the first ```NDCube``` in the sequence, another ```NDCube``` was returned.  However, if the slice came from a difference ```NDCube``` in the sequence, and ```NDCubeSequence``` was returned.

This PR ensures that no matter what single ```NDCube``` the (N-1)D slice is taken from it is always returned as an ```NDCube```.  Behaviour for when slices straddle multiple ```NDCube```s in sequence, etc. is unchanged.


ndcube/cube_utils.py
```diff
@@ -183,9 +184,9 @@ def index_sequence_as_cube(cubesequence, item):
     if item is None or (isinstance(item, tuple) and None in item):
         raise IndexError("None indices not supported")
     try:
```
abit2: A better approach would be to just check the first item i.e item[0] of tuple should be slice.
So something like - 
```
if isinstance(item_tuple, tuple):
     if isinstance(item_tuple[0], slice):
        return get_cube_from_sequence(cubesequence, item_tuple, type_slice='sequence_as_cube')

return get_cube_from_sequence(cubesequence, item_tuple)

abit2: This will remove the try and catch as it is vague in telling us which part of the code has failed.

DanRyanIrish: Hi @abit2.  Thanks for taking a look at this.  To be honest, I don't fully understand the difference of what ```slice_type=None``` and ```slice_type='sequence_as_cube``` mean.  Can you explain it to me?

abit2: https://github.com/sunpy/ndcube/pull/16 
so it needed some cleanup that's all :)
both of them were trying to do the same thing.

abit2:
https://github.com/sunpy/ndcube/pull/16

PR #16, opened 2017-09-24
code cleanup, making it more readable

abit2:
@DanRyanIrish please look into this code I think this should work I tried this on the tests and it worked fine.

### DIFF
--- a/ndcube/cube_utils.py
+++ b/ndcube/cube_utils.py
@@ -61,19 +61,15 @@ def get_cube_from_sequence(cubesequence, item, type_slice=None):
                 result = result.data[item[0]][item[1]]
             else:
                 result = result.data[item[0]][item[1::]]
-        # if the 0th index is slice.
-        # used for the index_sequence_as_cube function. Slicing across cubes.
-        # item represents (slice(start_cube_index, end_cube_index, None),
-        # [slice_of_start_cube, slice_of_end_cube]) if end cube is not sliced then length is 1.
-        elif isinstance(item[0], slice) and type_slice is None:
-            if len(item[1::]) is 1:
-                result.data = result.data[item[0]]
-                for i, data in enumerate(result.data):
-                    result.data[i] = data[item[1]]
-            else:
-                result.data = result.data[item[0]]
-                for i, data in enumerate(result.data):
-                    result.data[i] = data[item[1::]]
+        # the 0th index of tuple will have the slice that will be applied across cubes
+        # the 1st index of tuple contains the information of 1st cube to slice and last
+        # cube to slice.
+        # and the rest of the tuple contains the rest of the slice information applied to
+        # all cubes.
+        # example - (slice(0, 3, None), [slice(0, 2, None), slice(0, 1, None)], 1, 0)
+        # this part of the code should not be called directly by the __getitem__ method
+        # of the NDCubesequence as having slice in 0th index is not possible.
+        # so having this sequence_as_cube parameter helps.
         elif isinstance(item[0], slice) and type_slice is 'sequence_as_cube':
             data = result.data[item[0]]
             # applying the slice in the start of cube.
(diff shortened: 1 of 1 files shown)


abit2:
https://github.com/sunpy/ndcube/pull/14

DanRyanIrish:
Hi @abit2.  Thanks so much for this PR to help explain the point of the 'sequence_as_cube' kwarg.

ndcube/cube_utils.py
```diff
@@ -61,19 +61,15 @@ def get_cube_from_sequence(cubesequence, item, type_slice=None):
...
-                    result.data[i] = data[item[1::]]
+        # the 0th index of tuple will have the slice that will be applied across cubes
+        # the 1st index of tuple contains the information of 1st cube to slice and last
+        # cube to slice.
+        # and the rest of the tuple contains the rest of the slice information applied to
+        # all cubes.
+        # example - (slice(0, 3, None), [slice(0, 2, None), slice(0, 1, None)], 1, 0)
+        # this part of the code should not be called directly by the __getitem__ method
+        # of the NDCubesequence as having slice in 0th index is not possible.
+        # so having this sequence_as_cube parameter helps.
```
DanRyanIrish: These comments are very helpful!  Thanks!

ndcube/cube_utils.py
```diff
@@ -182,10 +178,13 @@ def index_sequence_as_cube(cubesequence, item):
...
-        return get_cube_from_sequence(cubesequence, item_tuple, type_slice='sequence_as_cube')
-    except:
-        return get_cube_from_sequence(cubesequence, item_tuple)
+
+    # if the slicing is done across one cube. Then we can pass the item_tuple as
+    # (int, rest of the information) not (int, int , rest of the information)
+    if isinstance(item_tuple[0], int) and isinstance(item_tuple[1], int):
+        if item_tuple[0] == item_tuple[1]:
+            return get_cube_from_sequence(cubesequence, item_tuple[1::])
+    return get_cube_from_sequence(cubesequence, item_tuple, type_slice='sequence_as_cube')
```
DanRyanIrish: Shouldn't there be an else preceding the above return statement here?

ndcube/cube_utils.py
```diff
@@ -182,10 +178,13 @@ def index_sequence_as_cube(cubesequence, item):
...
-    try:
-        return get_cube_from_sequence(cubesequence, item_tuple, type_slice='sequence_as_cube')
-    except:
-        return get_cube_from_sequence(cubesequence, item_tuple)
+
+    # if the slicing is done across one cube. Then we can pass the item_tuple as
+    # (int, rest of the information) not (int, int , rest of the information)
+    if isinstance(item_tuple[0], int) and isinstance(item_tuple[1], int):
+        if item_tuple[0] == item_tuple[1]:
+            return get_cube_from_sequence(cubesequence, item_tuple[1::])
```
DanRyanIrish: As best as I can tell, ```item_tuple``` is either 
(sequence_index_or_slice, cube_index_or_slice_dim0, cube_index_or_slice_dim1, cube_index_or_slice_dim2, etc) 
OR
(sequence_index_or_slice, [first_cube_index_or_slice_dim0, last_cube_index_or_slice_dim0], cube_index_or_slice_dim1, cube_index_or_slice_dim2, etc.) 

Is this correct so far?

Assuming it is, I don't understand a few things about the above if statement:
* Why it must be the case that ```item_tuple[0] == item_tuple[1]```
* Why ```item_tuple[1]``` must also be an ```int``` for the return line below to be executed.  Isn't all that matters is the ```item_tuple[0]``` is an ```int```, i.e. on one cube is being sliced?  So couldn't ```item_tuple[1]``` also be a slice?

abit2:
So what I remember is that when we are slicing in the same cube using index_as_cube we are getting something like (0, 0, other information of further dimensions) so I think for this is the only case that needs to be handled seperately. 
So we cannot have 0th index value as integer and 1st index value as slice.

ndcube/cube_utils.py
```diff
@@ -182,10 +178,13 @@ def index_sequence_as_cube(cubesequence, item):
...
         raise IndexError("None indices not supported")
-    try:
-        return get_cube_from_sequence(cubesequence, item_tuple, type_slice='sequence_as_cube')
-    except:
-        return get_cube_from_sequence(cubesequence, item_tuple)
+
+    # if the slicing is done across one cube. Then we can pass the item_tuple as
+    # (int, rest of the information) not (int, int , rest of the information)
+    if isinstance(item_tuple[0], int) and isinstance(item_tuple[1], int):
+        if item_tuple[0] == item_tuple[1]:
```
DanRyanIrish: I did some checking manual testing and the above condition (```if item_tuple[0] == item_tuple[1]:```) needs to be removed. 

At the moment, if ```item_tuple[0] == item_tuple[1]:```, a single slice of an (N-1)DCube is returned and a new ```NDCube``` object.  But if ```item_tuple[0] != item_tuple[1]``` a single slice of an (N-1)DCube is returned wrapped up in an ```NDCubeSequence```.

I'll merge this PR and then remove it as part of my current PR.

Issue #22, opened 2017-10-29
NDCubeSequence Indexing Bug


DanRyanIrish:
```
# Imports
from astropy.io import fits
from astropy.wcs import WCS
from ndcube import NDCube, NDCubeSequence

# Define files with which bug was found.
file0 = 'iris_l2_20141211_191222_3803105278_raster_t000_r00000.fits'
file1 = 'iris_l2_20141211_191222_3803105278_raster_t000_r00001.fits'

# Build NDCubes from C II 1336 data
ext_num = 1
hdulist0 = fits.open(file0)
ndc0 = NDCube(hdulist0[ext_num].data, WCS(hdulist0[ext_num].header))
hdulist0.close()
hdulist1 = fits.open(file1)
ndc1 = NDCube(hdulist1[ext_num].data, WCS(hdulist1[ext_num].header))
hdulist1.close()

# Build NDCubeSequence
nds = NDCubeSequence([ndc0, ndc1], common_axis=0)
```
```nds.index_as_cube[2]``` errors with the following traceback.
```
---------------------------------------------------------------------------
IndexError                                Traceback (most recent call last)
<ipython-input-56-f3fe4c7f3673> in <module>()
----> 1 nds.index_as_cube[2]

~/sunpy_dev/ndcube/ndcube/ndcube.py in __getitem__(self, item)
    695 
    696     def __getitem__(self, item):
--> 697         return cube_utils.index_sequence_as_cube(self.seq, item)
    698 
    699 

~/sunpy_dev/ndcube/ndcube/cube_utils.py in index_sequence_as_cube(cubesequence, item)
    183     # (int, rest of the information) not (int, int , rest of the information)
    184     if isinstance(item_tuple[0], int) and isinstance(item_tuple[1], int):
--> 185         return get_cube_from_sequence(cubesequence, item_tuple[1::])
    186     else:
    187         return get_cube_from_sequence(cubesequence, item_tuple, type_slice='sequence_as_cube')

~/sunpy_dev/ndcube/ndcube/cube_utils.py in get_cube_from_sequence(cubesequence, item, type_slice)
     61                 result = result.data[item[0]][item[1]]
     62             else:
---> 63                 result = result.data[item[0]][item[1::]]
     64         # the 0th index of tuple will have the slice that will be applied across cubes
     65         # the 1st index of tuple contains the information of 1st cube to slice and last

IndexError: list index out of range
```

DanRyanIrish:
[...]

PR #23, opened 2017-10-30
NDCubeSequence Slicing Bugfix

DanRyanIrish:
Eliminated type_slice kwarg from get_cube_from_sequence() as apparently not needed and causes bugs.

### DIFF
--- a/ndcube/cube_utils.py
+++ b/ndcube/cube_utils.py
@@ -36,7 +36,7 @@ def select_order(axtypes):
     return result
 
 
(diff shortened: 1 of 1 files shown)


DanRyanIrish:
@abit2, further to my recent comment on issue #22, I've remade this PR.  Can you confirm that these changes make sense to you?  If so, I can merge them and (presumably) fix this indexing bug.

Thanks!

abit2:
Seems good considering that seq[0:2, 0:2, 0] kind of operations is wrong and should be done with index_as_cube method instead.

PR #28, opened 2017-12-04
Move NDCubeSequence to a seperate file

Cadair:
This is just some light refactoring to move some of the code out of the `ndcube.py` file as it was getting a little large.


Cadair:
I tend to lean away from very long files and towards smaller groupings of related functionality. I am happy to move things about in this layout, but I think at least having NDCube and NDCubeSequence in separate files is a good idea.

ndcube/ndcube_sequence.py
```diff
@@ -0,0 +1,170 @@
...
+                axis_type=self.dimensions.axis_types[1::]))
+
+    @property
+    def dimensions(self):
+        return SequenceDimensionPair(
+            shape=tuple([len(self.data)]+list(self.data[0].dimensions.shape)),
+            axis_types=tuple(["Sequence Axis"]+self.data[0].dimensions.axis_types))
+
+    @property
+    def _common_axis_extra_coords(self):
```
DanRyanIrish: For simplicity, could you make this property public as part of this PR?  Same for ```NDCube._extra_coords```.

DanRyanIrish: On second thoughts, I realised this may have impact on tests and will definitely have a knock onto IRISpy.  So I'll do it in a separate PR.

PR #29, opened 2017-12-04
[WIP] Implement an Abstract Base Class to define the NDData API

Cadair:
This refactors the `NDCube` class a little so that the API is defined in a `NDCubeBase` object, which can not be instantiated itself but allows tests like `isinstance(inst, NDCubeBase)` to ensure API compatibility with things that look like `NDCube` 

This is now based on #34

### DIFF
--- a/ndcube/ndcube.py
+++ b/ndcube/ndcube.py
@@ -185,8 +185,8 @@ class NDCube(NDCubeSlicingMixin, NDCubePlotMixin, NDCubeBase):
         else:
             self._extra_coords_wcs_axis = None
         # Initialize NDCube.
-        super(NDCube, self).__init__(data, wcs=wcs, uncertainty=uncertainty, mask=mask,
-                                     meta=meta, unit=unit, copy=copy, **kwargs)
+        super().__init__(data, wcs=wcs, uncertainty=uncertainty, mask=mask,
+                         meta=meta, unit=unit, copy=copy, **kwargs)
 
     def pixel_to_world(self, quantity_axis_list, origin=0):
         list_arg = []
--- a/ndcube/ndcube.py
+++ b/ndcube/ndcube.py
@@ -85,6 +85,11 @@ class NDCubeBase(astropy.nddata.NDData, metaclass=NDCubeMetaClass):
         e.g. WAVE or HPLT-TAN for wavelength of helioprojected latitude.
         """
 
+    @abc.abstractproperty
+    def extra_coords(self):
+        """
+        """
+
     @abc.abstractmethod
     def crop_by_coords(self, lower_left_corner, dimension_widths):
         """
--- a/ndcube/ndcube.py
+++ b/ndcube/ndcube.py
@@ -77,18 +77,14 @@ class NDCubeBase(astropy.nddata.NDData, metaclass=NDCubeMetaClass):
             reverse of the wcs axis order.
         """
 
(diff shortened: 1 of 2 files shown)


ndcube/ndcube.py
```diff
@@ -1,40 +1,131 @@
...
+    Extra arguments are passed on to `~astropy.nddata.NDData`.
 
     Parameters
     ----------
     data: `numpy.ndarray`
         The array holding the actual data in this object.
 
     wcs: `ndcube.wcs.wcs.WCS`
-        The WCS object containing the axes' information
+        The WCS object containing the axes information
```
DanRyanIrish: I think the original version is actually right.  axes' information as in information of the axes.

ndcube/ndcube.py
```diff
@@ -1,40 +1,131 @@
...
 
+        Returns
+        -------
+        result: NDCube
 
-class NDCube(astropy.nddata.NDData):
+        """
+
+
+class NDCube(NDCubeBase):
```
DanRyanIrish: I think it might be a good idea to document somewhere in the ```NDCube``` attributes that the docstrings can be found in the ```NDCubeBase``` class so people ca easily find where to look when searching the codebase.  What do you think?

ndcube/ndcube.py
Cadair: It wasn't a grammatical fix I was trying to work out a sphinx bug. Turns out this wasn't it I will add it back.

ndcube/ndcube.py
```diff
@@ -1,29 +1,123 @@
...
+    def dimensions(self):
+        """
+        Returns a named tuple with two attributes: 'shape' gives the shape
+        of the data dimensions; 'axis_types' gives the WCS axis type of each dimension,
+        e.g. WAVE or HPLT-TAN for wavelength of helioprojected latitude.
+        """
+
+    @abc.abstractproperty
+    def extra_coords(self):
+        """
```
Cadair: @DanRyanIrish can you suggest a docstring for this?

DanRyanIrish: How about something like:
```
"""
Returns extra coordinates not included in the NDCube's WCS object.

Returns
-------
result: `dict`
    Dictionary of extra coords where each key is the name of an extra
    coordinate supplied by user during instantiation of the NDCube.
    The value of each key is itself a dictionary with the following keys:
    'axis': `int`
        The number of the data axis to which the extra coordinate corresponds.
    'value': `astropy.units.Quantity` or array-like
        The value of the extra coordinate at each pixel/array element along the
        corresponding axis (given by the 'axis' key, above).  Note this means
        that the length of 'value' must be equal to the length of the data axis
        to which is corresponds.
        
        """
```
?

ndcube/ndcube.py
```diff
@@ -12,13 +14,106 @@
...
+        origin : `int`.
+            Origin of the top-left corner. i.e. count from 0 or 1.
+            Normally, origin should be 0 when passing numpy indices, or 1 if
+            passing values from FITS header or map attributes.
+            See `~astropy.wcs.WCS.wcs_pix2world` for more information.
+            Default is 0.
+
+        Returns
+        -------
+
```
DanRyanIrish: Delete this empty line.

DanRyanIrish:
I approve of this PR pending it passing Travis.

PR #31, opened 2017-12-06
Make extra_coord properties public

DanRyanIrish:
This PR makes NDCube._extra_coords and NDCubeSequence._common_axis_extra_coords public attributes.


DanRyanIrish:
Thanks!

PR #36, opened 2018-01-04
NDCubeSequence Slicing

DanRyanIrish:
This PR devises a new framework for how ```NDCubeSequence``` are sliced under-the-hood.  It does not remove any essential functionality from the previous framework.  Instead it makes the code more transparent and gives more scope for future more powerful slicing capabilities in the future.

The framework defined two named tuples, ```SequenceSlice``` and ```SequenceItem```.  ```SequenceSlice``` contains a ```sequence_index```  (type=```int```) and ```common_axis_item``` (type=```int``` or ```slice```) attribute.  This describes a slice along the common axis of a particular ```NDCube``` within and ```NDCubeSequence```.  ```SequenceItem``` has attributes ``sequence_index```  (type=```int```) and ```cube_item``` (type=```int``` or ```slice``` or ```tuple```).  It is the same as ```SequenceSlice``` except that it can also describe slicing along non-common cube axes.

Various functions convert ```NDCubeSequence.__getitem__``` items and ```NDCubeSequence.index_as_cube``` items to a ```list``` of ```SequenceItem```, one for each cube included by the slicing items.  Once the ```list``` of ```SequenceItem``` objects has been derived, a function, ```slice_sequence```, applies these slices to the various cubes in the sequence and returns a new ```NDCubeSequence``` or ```NDCube``` based on the original slicing items.


DanRyanIrish:
@Cadair This PR is currently raising errors to with comparing ints with None types from slice objects.  It would be great to discuss the best way to handle this.  Perhaps there's a better way than converting all the Nones to the appropriate ints via if statements based on the slice's step size and sign?

DanRyanIrish:
@Cadair.  This PR seems to be passing except for some failures due to what seems to be a change to Travis.  Do you know about these errors?

PR #41, opened 2018-02-06
Added method to NDCube to crop by extra coordinate.

DanRyanIrish:


ndcube/ndcube.py
```diff
@@ -85,18 +85,19 @@ def extra_coords(self):
...
-        Crops an NDCube given a lower left corner and widths of region of interest.
+        Crops an NDCube given minimum values and interval widths along axes.
 
         Parameters
         ----------
-        lower_left_corner: `list` of `astropy.units.Quantity`
-            The lower left corner of the region of interest described in physical units
-            consistent with the NDCube's wcs object.  The length of the iterable must
-            equal the number of data dimensions and must have the same order as the data.
+        min_coord_values: `list` of `astropy.units.Quantity`
```
Cadair: What about the reductive case where all the axes are the same unit so this could be one `Quantity`? Is that supported?

ndcube/ndcube.py
```diff
@@ -85,18 +85,19 @@ def extra_coords(self):
...
+            The minimum desired values along each relevant axis after cropping
+            described in physical units consistent with the NDCube's wcs object.
+            The length of the iterable must equal the number of data dimensions and must
+            have the same order as the data.
 
-        dimension_widths: iterable of `astropy.units.Quantity`
+        interval_widths: iterable of `astropy.units.Quantity`
```
Cadair: same here. Also it's iterable here and `list` above. 'iterable' is more accurate, maybe it's worth standardising on that?

ndcube/ndcube.py
```diff
@@ -107,6 +108,29 @@ def crop_by_coords(self, lower_left_corner, dimension_widths):
...
         """
 
+    @abc.abstractmethod
+    def crop_by_extra_coord(self, min_coord_value, interval_width, extra_coord):
+        """
+        Crops an NDCube given a minimum value and interval width along an extra coord.
+
+        Parameters
+        ----------
+        min_coord_value: Single value of type consistent with the extra coord.
```
Cadair: Should the *type* here be `Quantity`? and the unit be consistent with the extra_coord?

ndcube/ndcube.py
```diff
@@ -107,6 +108,29 @@ def crop_by_coords(self, lower_left_corner, dimension_widths):
...
+        interval_width: Single value of type consistent with the extra coord.
+            The width of the interval along the extra coord axis in physical units
+            consistent with the extra coord.
+
+        extra_coord: `str`
+            Name of extra coord.
+
+        Returns
+        -------
+        result: NDCube
```
Cadair: this should be

```
`ndcube.NDCube`
```

ndcube/ndcube.py
DanRyanIrish: It's not supported.  You need to be explicit which in the reductive case is a bit more work but it clearer and always behaves the same.

ndcube/ndcube.py
DanRyanIrish: Good idea!  Done.

ndcube/ndcube.py
DanRyanIrish: Yes!  Done.

DanRyanIrish:
We should merge #42 before writing tests for this PR.

ndcube/ndcube.py
```diff
@@ -268,22 +294,37 @@ def dimensions(self):
...
-        if len(lower_left_corner) != len(dimension_widths) != n_dim:
-            raise ValueError("lower_left_corner and dimension_widths must have "
+        if len(min_coord_values) != len(interval_widths) != n_dim:
+            raise ValueError("min_coord_values and interval_widths must have "
                              "same number of elements as number of data dimensions.")
         # Convert coords of lower left corner to pixel units.
-        lower_pixels = self.world_to_pixel(lower_left_corner)
-        upper_pixels = self.world_to_pixel([lower_left_corner[i]+dimension_widths[i]
+        lower_pixels = self.world_to_pixel(min_coord_values)
+        upper_pixels = self.world_to_pixel([min_coord_values[i]+interval_widths[i]
```
Cadair: spaces around the `+` 😉

ndcube/ndcube.py
```diff
@@ -107,6 +108,31 @@ def crop_by_coords(self, lower_left_corner, dimension_widths):
...
+            The minimum desired value of the extra coord after cropping.
+            Unit must be consistent with the extra coord on which cropping is based.
+
+        interval_width: Single value `astropy.units.Quantity`
+            The width of the interval along the extra coord axis in physical units
+            consistent with the extra coord.  Unit must be consistent with the extra
+            coord on which cropping is based.
+
+        extra_coord: `str`
+            Name of extra coord.
```
Cadair: "name of the extra coordinate"?

PR #42, opened 2018-02-06
Refactor 'utils'

Cadair:
This refactors many a util.

It also adds them all to the doc build.

PR #46, opened 2018-02-07
Sequence extra coords property

DanRyanIrish:
This PR introduces a new property on ```NDCubeSequence``` requested by @MSKirk.  It is analogous to ```NDCubeSequence.common_axis_extra_coords``` except that instead of showing extra coords along the common axis, it collates extra coords not assigned to any cube data axis.  This will be very helpful in understanding any coords associated with cubes as a whole within a sequence, for example when putting 2D images together in a sequence.


DanRyanIrish:
We should merge #42 before writing tests for this PR.

DanRyanIrish:
@MSKirk Could you give this a code review?  I know @Cadair is busy and very tired ;)

One specific question for @Cadair though: ```test_sequence_axis_extra_coords()``` in ```test_ndcubesequence.py``` feels like it's written very verbosely because I don't know of an easy way to compare dictionaries whose key values include NaNs.  In addition, while ```numpy.testing.assert_array_almost_equal``` will not raise an error if NaNs are in the expected place in the arrays being compared, it crashes when the array contains ```datetime``` objects.  If you know a more elegant way to write this test, let me know.  Otherwise, no problem.  The test does what it needs to do.

DanRyanIrish:
@MSKirk does this meet your concerns?

Issue #47, opened 2018-02-08
convert_slice_nones_to_int() crops last (0th) element when step is negative


DanRyanIrish:
Bug example:
```>>> new_slice = ndcube.utils.sequence.convert_slice_nones_to_int(slice(5, None, -1), 10)```
which returns
```slice(5, 0, -1)```.
This results in
```np.arange(5)[new_slice]```
```array( [5, 4, 3, 2, 1] )```
when we really want
```array( [5, 4, 3, 2, 1, 0] )```

DanRyanIrish:
@Cadair, Do you know a trick to solve this?

DanRyanIrish:
Resolved in PR #49.

PR #49, opened 2018-02-09
Testing improvements

DanRyanIrish:
This PR rearranges the testing functions into a more logical structure since the utils refactor and increases the test coverage of non-plotting functionality.


DanRyanIrish:
Still a work in progress.

DanRyanIrish:
@Cadair, do you know how to interpret this error message for the failing test of this PR?  I have no idea what it's telling me.

https://travis-ci.org/sunpy/ndcube/jobs/339717047

ndcube/utils/sequence.py
```diff
@@ -323,9 +332,16 @@ def _convert_cube_like_index_to_sequence_slice(cube_like_index, common_axis_cube
     """
     # Derive cumulative lengths of cubes along common axis.
     cumul_common_axis_cube_lengths = np.cumsum(common_axis_cube_lengths)
-    # If cube_like_index is within first cube in sequence, it is
+    # If cube_like_index is within 0th cube in sequence, it is
     # simple to determine the sequence and common axis indices.
-    if cube_like_index < cumul_common_axis_cube_lengths[0]:
+    try:
+        index_in_0th_cube = cube_like_index < cumul_common_axis_cube_lengths[0]
```
Cadair: What's returning `None` here? I feel like this could be done without the `try`

DanRyanIrish:
@Cadair, I think this is ready for a code review now.  It's turned into a bit of a monster PR as bugs etc. were found during testing.  But this PR now does the following:
*Increases overall test coverage up from 70% to 86%
*Adds support in NDCubeOrdered for uncertainty, mask and extra_coords
*Fixes issue #47 
*Fixes numerous minor bugs found in testing
*Increases and improves documentation

nabobalis:
I would pin sphinx to <1.7 for now.

ndcube/utils/sequence.py
DanRyanIrish: The issue is in comparing ```cube_like_index``` with ```cumul_common_axis_cube_lengths[0]```.  The latter is always a number.  The former is also a number except in one circumstance.  That is when the slice object from which ```cube_like_index``` is derived has a negative step and a stop parameter of None, i.e. the last included element is the 0th.  In this case, the stop parameter cannot be expressed by a number since stop means "up to but not including", and ```-1``` has a different meaning in Python slicing.  Therefore, in this case comparing ```cube_like_index``` with ```cumul_common_axis_cube_lengths[0]``` will raise a ```TypeError``` complaining about comparing ```int``` to ```None``` which can only mean that the index is in the first cube and so the if statement below must be entered.

Issue #30, opened 2017-12-04
Adopt API from Astropy's WCS APE


Cadair:
There is an open proposal for a WCS API here: https://github.com/astropy/astropy-APEs/pull/34 I think it would make sense if the `NDCube` API mirror that specification where appropriate, I am specifically suggesting the following methods be changed:

1. Change the signature of `NDCube.pixel_to_world` to:
```
    def pixel_to_world(self, *pixel_arrays):
```
Which would mean accepting arrays of `Quantity` objects in `u.pixel` and dropping the `origin=0` argument (and always assuming `origin=0`).

2. and the same for `NDCube.world_to_pixel`:

```
    def world_to_pixel(self, *world_objects):
```
Which would mean only accepting `SkyCoord` or `Quantity` objects as appropriate (this might be harder in the short-term, we might just have to accept `Quantity`).

3. Change `NDCube.dimensions` so that it just returns the shape of the axes (with units e.g. `GenericMap.dimensions`). Then implement a `NDCube.world_axis_physical_types` method following the APE: (see [this](https://github.com/astrofrog/astropy-APEs/blob/79f6d1eac3cebfce03b8689eb51d39a93d8fd0f7/APE13.rst#common-ucd1-names-for-physical-types))
```
        @property
        def world_axis_physical_types(self):
            """
            Returns an iterable of strings describing the physical type for each
            world axis. They should be names from the VO UCD1+ controlled
            Vocabulary (http://www.ivoa.net/documents/latest/UCDlist.html).
            If no matching UCD type exists, this can instead be "custom:xxx",
            where xxx is an arbitrary string.  Alternatively, if the physical
            type is unknown/undefined, an element can be `None`.
            """
```

Cadair:
I think at least some of this should be done before 0.1. @DanRyanIrish ?

DanRyanIrish:
Hi @Cadair.  Thanks for this issue.  Regarding points 1 and 2, we currently accept a list of ```u.Quantity``` objects for ```pixel_to_world``` and ```world_to_pixel```.  So as far as I see the only change would be getting rid of the ```origin``` kwarg.  You're right thought that we don't currently deal with ```SkyCoord```.  I've never used it before so not sure what is required to support it.

As for point 3, that shouldn't be too hard to implement.

DanRyanIrish:
[...]

PR #55, opened 2018-02-14
Astropy WCS APE Implementation

DanRyanIrish:
This PR replaces the current ```NDCube.dimensions``` and ```NDCubeSequence.dimensions``` API to be consistent with the new astropy WCS APE.  See issue #30 for more details.


ndcube/plot_helpers.py
```diff
@@ -0,0 +1,105 @@
+import numpy as np
```
Cadair: why is this file back?!

DanRyanIrish: I have no idea how this happened!

DanRyanIrish:
@Cadair, I think this is ready to merge unless you have any comments/change requests?

docs/ndcubesequence.rst
```diff
@@ -129,12 +129,16 @@ more details on these features.
...
+`~ndcube.NDCubeSequence.dimensions` and
+`~ndcube.NDCubeSequence.world_axis_physical_types` properties for
+easily inspecting the shape and phsyical axis types of an
+`~ndcube.NDCubeSequence` instance::
 
   >>> my_sequence.dimensions
-  SequenceDimensionPair(shape=(3, <Quantity 3. pix>, <Quantity 4. pix>, <Quantity 5. pix>), axis_types=('Sequence Axis', 'HPLN-TAN', 'HPLT-TAN', 'WAVE'))
+  (<Quantity 3. pix>, <Quantity 3. pix>, <Quantity 4. pix>, <Quantity 5. pix>)
+  >>> my_sequence.world_axis_physical_types
+  ('custom:Sequence Axis', 'custom:pos.helioprojective.lon', 'custom:pos.helioprojective.lat', 'em.wl')
```
Cadair: I don't like `custom:Sequence Axis` we should make it closer to the general form of the UCD1+ vocab, i.e. `thing.thing`.

Cadair: `pos.obs.sequence`

ndcube/utils/wcs.py
```diff
@@ -10,7 +10,40 @@
...
+    def __len__(self):
+        """Returns the number of connections"""
+        return dict.__len__(self) // 2
+
+
+# Define a two way dictionary to hold translations between WCS axis
+# types and International Virtual Observatory Alliance vocabulary.
+# See http://www.ivoa.net/documents/REC/UCD/UCDlist-20070402.html
+wcs_to_ivoa = {
+    "HPLT-TAN": "custom:pos.helioprojective.lat",
```
Cadair: this should just be `"HPLT"`, i.e. we shouldn't be checking the `-XXX` projection, as any projection is still `pos.helioprojective.lat`.

ndcube/utils/wcs.py
```diff
@@ -10,7 +10,40 @@
 from astropy import wcs
 from astropy.wcs._wcs import InconsistentAxisTypesError
 
-__all__ = ['WCS', 'reindex_wcs', 'add_celestial_axis']
+__all__ = ['WCS', 'reindex_wcs', 'add_celestial_axis', 'wcs_ivoa_mapping']
+
+
+class TwoWayDict(dict):
```
Cadair: I suggest you replace this with:
```python
from collections import UserDict


class TwoWayDict(UserDict):
    @property
    def inv(self):
        """"
        The inverse dictionary.
        """"
        return {v:k for k, v in self.items()}
```

As it's API compatible with https://github.com/jab/bidict and rather cleaner than this one.

Cadair:
also please placate the PEP8 bot, at least mostly.

ndcube/ndcube.py
```diff
@@ -255,20 +259,33 @@ def dimensions(self):
...
 
+        """
         ctype = list(self.wcs.wcs.ctype)
         axes_ctype = []
         for i, axis in enumerate(self.missing_axis):
             if not axis:
-                axes_ctype.append(ctype[i])
-        shape = u.Quantity(self.data.shape, unit=u.pix)
-        return DimensionPair(shape=shape, axis_types=axes_ctype[::-1])
+                axes_ctype.append(wcs_ivoa_mapping[ctype[i]])
```
Cadair: You should probably use `.get` here to default to `None` if we don't know what it is.

ndcube/ndcube.py
```diff
@@ -255,20 +259,33 @@ def dimensions(self):
...
+
+        """
+        return u.Quantity(self.data.shape, unit=u.pix)
+
+    @property
+    def world_axis_physical_types(self):
         """
+        Returns an iterable of strings describing the physical type for each world axis.
+
+        They should be names from the VO UCD1+ controlled
```
Cadair: I would re-write this section of the docstring to describe the output to a user rather than the description to developers from the APE.

docs/ndcubesequence.rst
DanRyanIrish: Agreed.  Changed it to ```meta.obs.sequence```.

ndcube/utils/wcs.py
DanRyanIrish: Done

ndcube/utils/wcs.py
DanRyanIrish: OK.  Done.

ndcube/ndcube.py
DanRyanIrish: Good idea!  But what about the default being ```custom:unknown```?  That way it's always a string and has a similar format?

ndcube/ndcube.py
DanRyanIrish: Good spot!  Hadn't looked at that since I pasted it into the file!

ndcube/ndcube.py
Cadair: The APE spec says it should be None if it's undefined.

DanRyanIrish: Fair enough.

ndcube/ndcube.py
```diff
@@ -26,7 +27,7 @@ class NDCubeMetaClass(abc.ABCMeta, InheritDocstrings):
 class NDCubeBase(astropy.nddata.NDData, metaclass=NDCubeMetaClass):
 
     @abc.abstractmethod
-    def pixel_to_world(self, quantity_axis_list, origin=0):
+    def pixel_to_world(self, quantity_axis_list):
```
Cadair: Shouldn't this be `*quantity_axis_list` to be APE compliant?

Issue #54, opened 2018-02-13
Remodel NDCubeSequence.common_axis_extra_coords on NDCubeSequence.sequence_axis_extra_coords


DanRyanIrish:
Currently ```NDCubeSequence.common_axis_extra_coords``` assumes that all cubes within the sequence have the same ```extra_coords``` and so gets the "full" list of ```extra_coords``` from the first cube in the sequence.  By contrast, ```NDCubeSequence.common_axis_extra_coords``` more thoroughly checks all cubes to compile a record of all ```extra_coords```.  ```NDCubeSequence.common_axis_extra_coords``` should be remodelled to be as rigorous as ```NDCubeSequence.common_axis_extra_coords```.

DanRyanIrish:
This issue is addressed by #56

PR #58, opened 2018-02-16
Refactor NDCubeSequence.common_axis_extra_coords

DanRyanIrish:
This refactor means NDCubeSequence.common_axis_extra_coords no longer assumes all cubes have the same extra_coords as the first cube.  But performs a more rigorous search along the lines of NDCubeSequence.sequence_axis_extra_coords.  See ndcube/sunpy Issue #54.

### DIFF
--- a/ndcube/ndcube_sequence.py
+++ b/ndcube/ndcube_sequence.py
@@ -34,7 +34,10 @@ class NDCubeSequence:
     def __init__(self, data_list, meta=None, common_axis=None, **kwargs):
         self.data = data_list
         self.meta = meta
-        self._common_axis = common_axis
+        if common_axis is not None:
+            self._common_axis = int(common_axis)
+        else:
+            self._common_axis = common_axis
 
     def __getitem__(self, item):
         return utils.sequence.slice_sequence(self, item)
--- a/ndcube/ndcube_sequence.py
+++ b/ndcube/ndcube_sequence.py
@@ -131,37 +131,19 @@ Axis Types of 1st NDCube: {axis_type}
 
     @property
     def common_axis_extra_coords(self):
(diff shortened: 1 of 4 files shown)


ndcube/ndcube_sequence.py
```diff
@@ -131,37 +131,19 @@ def cube_like_world_axis_physical_types(self):
...
-                                c.extra_coords[coord_name]["value"].to(coord_unit).value)
-                                        for c in self.data])
-                            common_extra_coords[coord_name] = u.Quantity(np.concatenate(qs),
-                                                                         unit=coord_unit)
-                        except AttributeError:
-                            qs = tuple([np.asarray(c.extra_coords[coord_name]["value"])
-                                        for c in self.data])
-                            common_extra_coords[coord_name] = np.concatenate(qs)
-        return common_extra_coords
+        if not isinstance(self._common_axis, int):
```
Cadair: do we always cast to an integer when setting `self._common_axis`? Otherwise this could end up as a float?

ndcube/tests/test_ndcubesequence.py
```diff
@@ -218,7 +218,10 @@ def test_common_axis_extra_coords(test_input, expected):
     output = test_input.common_axis_extra_coords
     assert output.keys() == expected.keys()
     for key in output.keys():
-        assert (output[key] == expected[key]).all()
+        try:
```
Cadair: can't you just always `all()` this to remove the need for the `try`?

ndcube/ndcube_sequence.py
DanRyanIrish: Good point!  No we don't  I'll add that into the ```NDCubeSequence.__init__```.

PR #60, opened 2018-02-19
New method: NDCube.all_world_coords

DanRyanIrish:
This PR introduces a new method/property, ```NDCube.all_world_coords``` which returns the world coordinates of all pixels along each axis in the ```NDCube``` instance.  It handles cases of dependent axes as well as independent ones.

### DIFF
--- a/docs/ndcube.rst
+++ b/docs/ndcube.rst
@@ -379,7 +379,7 @@ coordinates for each pixel along a given data axis.  So in the case of
 we could call::
 
   >>> my_cube.all_world_coords(axes=2)
-  [<Quantity [1.02e-09, 1.04e-09, 1.06e-09, 1.08e-09, 1.10e-09] m>]
+  <Quantity [1.02e-09, 1.04e-09, 1.06e-09, 1.08e-09, 1.10e-09] m>
 
 Note we set ``axes`` to ``2`` since ``axes`` is defined in data axis
 order.  We can also define the axis we want using any unique substring
@@ -390,7 +390,7 @@ from the axis names defined in
   ('custom:pos.helioprojective.lon', 'custom:pos.helioprojective.lat', 'em.wl')
   >>> # Since 'wl' is unique to the wavelength axis name, let's use that.
   >>> my_cube.all_world_coords(axes='wl')
-  [<Quantity [1.02e-09, 1.04e-09, 1.06e-09, 1.08e-09, 1.10e-09] m>]
+  <Quantity [1.02e-09, 1.04e-09, 1.06e-09, 1.08e-09, 1.10e-09] m>
 
 Notice how this returns the same result as when we set ``axes`` to
 the corresponding data axis number.  As discussed above, some WCS axes
@@ -408,19 +408,19 @@ longitude x latitude axis lengths.  For example::
   >>> longitude.shape
   (3, 4)
   >>> longitude
-  [<Quantity [[0.60002173, 0.59999127, 0.5999608 , 0.59993033],
-            [1.        , 1.        , 1.        , 1.        ],
-            [1.39997827, 1.40000873, 1.4000392 , 1.40006967]] deg>]
+  <Quantity [[0.60002173, 0.59999127, 0.5999608 , 0.59993033],
+             [1.        , 1.        , 1.        , 1.        ],
+             [1.39997827, 1.40000873, 1.4000392 , 1.40006967]] deg>
 
 It is also possible to request more than one axis's world coordinates
 by setting ``axes`` to an iterable of data axis number and/or axis
 type strings.::
 
   >>> my_cube.all_world_coords(axes=(2, 'lon'))
-  [<Quantity [1.02e-09, 1.04e-09, 1.06e-09, 1.08e-09, 1.10e-09] m>,
+  (<Quantity [1.02e-09, 1.04e-09, 1.06e-09, 1.08e-09, 1.10e-09] m>,
    <Quantity [[0.60002173, 0.59999127, 0.5999608 , 0.59993033],
               [1.        , 1.        , 1.        , 1.        ],
-              [1.39997827, 1.40000873, 1.4000392 , 1.40006967]] deg>]
+              [1.39997827, 1.40000873, 1.4000392 , 1.40006967]] deg>)
 
 Notice that the axes' coordinates have been returned in the same order
 in which they were requested.  Finally, if the user wanted the world
@@ -428,7 +428,7 @@ coordinates for all the axes, ```axes`` can be set to ``None``, which
 is in fact the default.::
 
   >>> my_cube.all_world_coords()
-  [<Quantity [[0.60002173, 0.59999127, 0.5999608 , 0.59993033],
+  (<Quantity [[0.60002173, 0.59999127, 0.5999608 , 0.59993033],
             [1.        , 1.        , 1.        , 1.        ],
             [1.39997827, 1.40000873, 1.4000392 , 1.40006967]] deg>,
    <Quantity [[1.26915033e-05, 4.99987815e-01, 9.99962939e-01, 
@@ -437,7 +437,7 @@ is in fact the default.::
              1.49989848e+00],
             [1.26915033e-05, 4.99987815e-01, 9.99962939e-01,
              1.49986193e+00]] deg>,
-   <Quantity [1.02e-09, 1.04e-09, 1.06e-09, 1.08e-09, 1.10e-09] m>]
+   <Quantity [1.02e-09, 1.04e-09, 1.06e-09, 1.08e-09, 1.10e-09] m>)
 
 As stated at the top of this guide, `~ndcube.NDCube` is only written
 to handle single arrays described by single WCS instances.  For cases
--- a/docs/ndcube.rst
+++ b/docs/ndcube.rst
@@ -396,7 +396,7 @@ the corresponding data axis number.
 
 As discussed above, some WCS axes
 are not independent.  For those axes,
-`~ndcube.NDCube.all_world_coords()` returns a
+`~ndcube.NDCube.all_world_coords` returns a
 `~astropy.units.Quantity` with the same number of dimensions as
 dependent axes.  For example, helioprojective longitude and latitude
 are dependent.  Therefore if we ask for longitude, we will get back a
--- a/docs/ndcube.rst
+++ b/docs/ndcube.rst
@@ -301,7 +301,7 @@ translate between pixel and real world coordinates.  For this purpose,
 astropy functions, `astropy.wcs.WCS.all_pix2world` and
 `astropy.wcs.WCS.all_world2pix`. These are
 `~ndcube.NDCube.pixel_to_world`, `~ndcube.NDCube.world_to_pixel`, and
-`~ndcube.NDCube.all_world_coords`. It is highly recommended that when
+`~ndcube.NDCube.axis_world_coords`. It is highly recommended that when
 using `~ndcube.NDCube` these convenience wrappers are used rather than
 the original astropy functions for a few reasons. For example, they
 can track house-keeping data, are aware of "missing" WCS axis, are
@@ -368,11 +368,11 @@ different instruments with overlapping fields of view.
 There are times however, when you only want to know the real world
 coordinates of the `~ndcube.NDCube` field of view.  To make this easy,
 `~ndcube.NDCube` has a another coordinate transformation method
-`~ndcube.NDCube.all_world_coords`.  This method returns the real world
+`~ndcube.NDCube.axis_world_coords`.  This method returns the real world
 coordinates for each pixel along a given data axis.  So in the case of
 ``my_cube``, if we wanted the wavelength axis we could call::
 
-  >>> my_cube.all_world_coords(axes=2)
+  >>> my_cube.axis_world_coords(2)
   <Quantity [1.02e-09, 1.04e-09, 1.06e-09, 1.08e-09, 1.10e-09] m>
 
 Note we set ``axes`` to ``2`` since ``axes`` is defined in data axis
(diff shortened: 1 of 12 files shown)


DanRyanIrish:
@Cadair, would be interested to see if you have any comments on this PR.
* Should this be a method or property?
* Should it part of 1.0?  (I would argue yes, as I think it will highly used!)
* Is there an obviously better way to code this up?
* etc.

ndcube/ndcube.py
```diff
@@ -366,6 +367,48 @@ def extra_coords(self):
                     "value": self._extra_coords_wcs_axis[key]["value"]}
         return result
 
+    @property
+    def all_world_coords(self):
```
Cadair: This should probably take a axes input.

Also it should take in any unique substring of `world_axes_physical_types` or an int.

ndcube/utils/wcs.py
```diff
@@ -106,6 +106,41 @@ def _augment(cls, header, naxis):
             newheader['CTYPE' + axis] = projection
         return newheader
 
+    def dependent_axes(self, axis):
```
Cadair: I think this should be a function not a method of the WCS class, just incase someone is using `astropy.wcs.WCS`.

ndcube/ndcube.py
DanRyanIrish: And it should also be a method rather than a property.

ndcube/utils/cube.py
```diff
@@ -119,3 +122,63 @@ def convert_extra_coords_dict_to_input_format(extra_coords, missing_axis):
...
+        Number of pixels in each dimension
+
+    Returns
+    -------
+    quantity_list: `list` of `astropy.units.Quantity`
+        Othogonal pixel quantities describing all pixels in cube in relevant dimensions.
+
+    """
+
+    print(dependent_axes)
```
Cadair: rouge print

DanRyanIrish:
@Cadair, ready for more comments ;)

DanRyanIrish:
Should this method perhaps be renamed?  ```NDCube.axes_world_coords```?

docs/ndcube.rst
```diff
@@ -89,21 +89,24 @@ axis types, `~ndcube.NDCube.dimensions` and
...
 pixel units giving the length of each dimension in the
 `~ndcube.NDCube` while `~ndcube.NDCube.world_axis_physical_types`
-returns an iterable of strings denoted the type of physical property
-represented by each axis.  Here the shape and axis types are given in
-data order, not WCS order.
+returns an iterable of strings denoting the type of physical property
+represented by each axis.  The axis names are in accordance with the
+International Virtual Observatory Alliance (IVOA) `UCD1+ controlled
+vocabulary <
+http://www.ivoa.net/documents/REC/UCD/UCDlist-20070402.html>`. Here
```
Cadair: this needs an `_` after the `>`

docs/ndcube.rst
```diff
@@ -355,12 +361,89 @@ returned::
...
+  >>> # Since 'wl' is unique to the wavelength axis name, let's use that.
+  >>> my_cube.all_world_coords(axes='wl')
+  <Quantity [1.02e-09, 1.04e-09, 1.06e-09, 1.08e-09, 1.10e-09] m>
+
+Notice how this returns the same result as when we set ``axes`` to
+the corresponding data axis number.
+
+As discussed above, some WCS axes
+are not independent.  For those axes,
+`~ndcube.NDCube.all_world_coords()` returns a
```
Cadair: drop the `()`

ndcube/ndcube.py
```diff
@@ -366,6 +372,98 @@ def extra_coords(self):
...
         return result
 
+    def all_world_coords(self, axes=None):
+        """
+        Returns WCS coordinate values of all pixels for all axes.
+
+        Parameters
+        ----------
+        axes: `int` or `str`, or iterable of `int` or `str`
+            Axis number in numpy ordering or unique substring of world_axis_physical_types
```
```
`~ndcube.NDCube.world_axis_physical_types`
```

ndcube/ndcube.py
```diff
@@ -366,6 +372,98 @@ def extra_coords(self):
...
+            of axes for which real world coordinates are desired.
+            Default=None implies all axes will be returned.
```
Cadair: `axes=None`?

ndcube/ndcube.py
```diff
@@ -366,6 +372,98 @@ def extra_coords(self):
...
+
+        Example
+        -------
+        >>> NDCube.all_world_coords(('lat', 'lon')) # doctest: +SKIP
+        >>> NDCube.all_world_coords(2) # doctest: +SKIP
+
+        """
+        # Define the dimensions of the cube and the total number of axes.
+        cube_dimensions = np.array(self.dimensions.value, dtype=int)
+        n_dimensions = len(cube_dimensions)
```
Cadair: probably slightly cleaner to use `cube_dimensions.size` here.

ndcube/ndcube.py
```diff
@@ -366,6 +372,98 @@ def extra_coords(self):
...
+        world_axis_types = self.world_axis_physical_types
```
Cadair: this seems a little redundant?

ndcube/ndcube.py
```diff
@@ -366,6 +372,98 @@ def extra_coords(self):
...
+        if axes is None:
+            int_axes = np.arange(n_dimensions)
+        else:
+            if isinstance(axes, int):
+                int_axes = np.array([axes])
+            elif isinstance(axes, str):
+                int_axes = np.array([
+                    utils.cube.get_axis_number_from_axis_name(axes, world_axis_types)])
+            else:
+                int_axes = []
```
Cadair: this could be `int_axes = np.empty(len(axes), dtype=int)` then you wouldn't need to construct a list and convert to an array.

ndcube/ndcube.py
```diff
@@ -366,6 +372,98 @@ def extra_coords(self):
...
+                for axis in axes:
+                    if isinstance(axis, int):
+                        int_axes.append(axis)
+                    elif isinstance(axis, str):
+                        int_axes.append(
+                            utils.cube.get_axis_number_from_axis_name(axis, world_axis_types))
+                int_axes = np.array(int_axes)
+        # Ensure user has not entered the same axis twice.
+        if len(np.unique(int_axes)) != len(int_axes):
```
Cadair: I would re-write this as:
```python
repeats = set([x for x in int_axes if int_axes.count(x) > 1])
if repeats:
    raise ValueError("The follwing axes were specified more than once: {}".format(' '.join(map(str, repeats))))
```

so we tell the user what axes they have repeated.

ndcube/ndcube.py
```diff
@@ -366,6 +372,98 @@ def extra_coords(self):
...
+            raise ValueError("Same axis entered more than once.")
+        n_axes = len(int_axes)
+        axes_coords = [None] * n_axes
+        axes_translated = np.array([False] * n_axes)
```
Cadair: this should be `np.zeros_like(int_axes, dtype=bool)`

ndcube/ndcube.py
```diff
@@ -366,6 +372,98 @@ def extra_coords(self):
...
+        # Determine which axes are dependent on others.
+        # Ensure the axes are in numerical order.
+        dependent_axes = np.empty(n_axes, dtype=object)
```
Cadair: I think you can replace this whole block with:
```python
wcs_get_dependent = partial(utils.wcs.get_dependent_axes, self.wcs)
dependent_axes = list(map(tuple, map(sorted, map(wcs_get_dependent, int_axes))))
```

:laughing: 

Having said that, I feel like you could just modifiy `utils.wcs.get_dependent_axes` to do a `sorted` in the right places.

ndcube/ndcube.py
```diff
@@ -366,6 +372,98 @@ def extra_coords(self):
...
+        for i, axis in enumerate(int_axes):
+            x = np.array(utils.wcs.get_dependent_axes(self.wcs, axis))
+            x.sort()
+            dependent_axes[i] = x
+        n_dependent_axes = np.array([len(da) for da in dependent_axes])
+        # Iterate through each axis and perform WCS translation.
+        for i, axis in enumerate(int_axes):
+            # If axis has already been translated, do not do so again.
+            if axes_translated[i] == False:
```
Cadair: `if not axes_translated[i]`

ndcube/ndcube.py
```diff
@@ -366,6 +372,98 @@ def extra_coords(self):
...
+                if n_dependent_axes[i] == 1:
+                    # Construct pixel quantities in each dimension letting
+                    # other dimensions all have 0 pixel value.
+                    quantity_list = \
```
Cadair: `quantity_list = [` rather than the explicit `\`

ndcube/ndcube.py
```diff
@@ -366,6 +372,98 @@ def extra_coords(self):
...
+                      [u.Quantity(np.zeros(tuple(cube_dimensions[dependent_axes[i]])),
```
Cadair: what's the `tuple` achieving in this line?

ndcube/utils/cube.py
```diff
@@ -119,3 +123,88 @@ def convert_extra_coords_dict_to_input_format(extra_coords, missing_axis):
...
+        # Define list of indices/slice objects for accessing
+        # sections of dependent axis arrays that are to be
+        # replaced with orthogonal np.arange arrays.
+        slice_list = [0] * n_dependent_axes
+        slice_list[i] = slice(None)
+        # Define array of indices of axes in slice list not
+        # including the axis along which we are inserting
+        # np.arange.
+        other_axes_indices = list(range(n_dependent_axes))
+        del(other_axes_indices[i])
```
Cadair: `other_axes_indices.pop(i)`?

ndcube/utils/cube.py
```diff
@@ -119,3 +123,88 @@ def convert_extra_coords_dict_to_input_format(extra_coords, missing_axis):
                 raise KeyError("extra coords dict can have keys 'wcs axis' or 'axis'.  Not both.")
             result.append((name, axis, extra_coords[name]["value"]))
         return result
+
+
+def _get_pixel_quantities_for_dependent_axes(dependent_axes, cube_dimensions):
```
Cadair: This function scares me a little!! What is it trying to do?

DanRyanIrish: We should replace this with ```np.meshgrid(..., indexing='ij')```

ndcube/utils/cube.py
DanRyanIrish: Does it make a difference in this case?  ```pop``` is good for keeping the deleted element in another variable.  But in this case we just want it gone entirely.

docs/ndcube.rst
DanRyanIrish: Is that a double underscore?

docs/ndcube.rst
DanRyanIrish: Done

ndcube/ndcube.py
DanRyanIrish: Done

ndcube/ndcube.py
DanRyanIrish: Done

ndcube/ndcube.py
DanRyanIrish: Done

ndcube/ndcube.py
DanRyanIrish: It's so the computing within ```self.world_axis_physical_types``` is only done once and not repeated each time it's called within a loop.  May not make a big difference, but any speed up we can get may be helpful for large ```NDCube```s.

ndcube/ndcube.py
DanRyanIrish: Done.

ndcube/ndcube.py
DanRyanIrish: Good suggestion!  Adapted a little as ```count``` is a list attribute, not an ```ndarray``` attribute.

But done nonetheless.

ndcube/ndcube.py
DanRyanIrish: Done

ndcube/ndcube.py
DanRyanIrish: Ha!  Yeh, I actually forgot I'm not importing ```get_dependent_axes```, so I know it wont change unless we change it.  And currently it already returns the axes ordered.  So I've rewritten that code chunk to be two list comprehensions :D

ndcube/ndcube.py
DanRyanIrish: Done

ndcube/ndcube.py
DanRyanIrish: Done

ndcube/ndcube.py
DanRyanIrish: Nothing I guess ;)

ndcube/ndcube.py
```diff
@@ -267,15 +279,8 @@ def dimensions(self):
 
     @property
     def world_axis_physical_types(self):
-        """
```
Cadair: You can't delete the the docstring here until we can rely on astropy/astropy#7166 as it currently dosen't inherit for properties.

ndcube/ndcube.py
```diff
@@ -366,9 +371,109 @@ def extra_coords(self):
                     "value": self._extra_coords_wcs_axis[key]["value"]}
         return result
 
+    def all_world_coords(self, axes=None):
```
Cadair: I would change this to `(self, *axes)` the default would then be an empty tuple and not `None`.

DanRyanIrish:
Awesome!  I'll tackle those today!

Cadair:
Good with me whenever you are ready Danny. If you merge this in I will try and do the release tomorrow morning.

DanRyanIrish:
Great!  I'll merge it tonight :)

PR #66, opened 2018-02-21
Travis changes

nabobalis:
Main "Work"

Stage Builds: First tests are to install package and run PEP8 and test suite
Second round are to build vs LTS Astropy, Run coverage and build docs
3 CronJobs: Astropy Dev, SunPy Dev, Mac OS X

Also made pep8speaks only show diff for PR. We might want to discuss this one. I want this on sunpy main.  Also wondering if we can get it to ignore the files which it seems to ignore in the wrong way.

### DIFF
--- a/.pep8speaks.yml
+++ b/.pep8speaks.yml
@@ -1,9 +1,14 @@
+scanner:
+    diff_only: True  # If True, errors caused by only the patch are shown
+
 pycodestyle:
-  max-line-length: 100 # Default is 79 in PEP8
+    max-line-length: 100  # Default is 79 in PEP8
 
 exclude:
-  - setup.py
-  - ez_setup.py
-  - ah_bootstrap.py
-  - astropy_helpers/
-  - docs/conf.py
+    - setup.py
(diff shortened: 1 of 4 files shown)


nabobalis:
On a side note, what about appveyor?

Cadair:
yeah I will turn it on if you wanna conf it?

nabobalis:
I wasn't sure if you cared. I can conf it, at least try to conf it.

nabobalis:
Appveyor seems to be alive now.

I only setup to run test versions with Py3.5 and Py3.6 for normal installs. No dev or LTS checks for example. Which is similar to how it works for main SunPy repo.

nabobalis:
Also probably want to enable a cron job if not already.

DanRyanIrish:
What exactly does appveyor do, by the way, @nabobalis @Cadair ?

nabobalis:
Travis but for windows.

DanRyanIrish:
Ah!  Right.  That's good to know!

PR #72, opened 2018-02-23
Enhance NDCubeSequence.plot capabilities

DanRyanIrish:
This PR aims to add two capabilities to ```NDCubeSequence.plot```
* Work when ```common_axis``` is set with the sequence axis acting as another dimension in the animation.  This required adding a dummy axis to the WCS used in the animation.
* Work when there are "missing axes" in the cubes, i.e. number of data dimensions != number of dimensions in the WCS objects.
* It also replaces the coding of ```0``` as the common axis and now takes that from the ```NDCubeSequence._common_axis``` attribute.


DanRyanIrish:
Hi @Cadair.  This is the second PR I mentioned in response to your query on riot earlier.  It seems to fix the issue well enough.  But comments on how I've gone about it are welcome.

PR #77, opened 2018-02-26
Add installation instructions to docs.

DanRyanIrish:


DanRyanIrish:
@Cadair do you know why sphinx is failing here?

DanRyanIrish:
@Cadair Can this doc change be retroactively put into 1.0?  Or is that ill-advised?

nabobalis:
You could rebuild the docs and point it to the new version. If they don't already point to latest on RTD.

Cadair:
you indented poorly: https://travis-ci.org/sunpy/ndcube/jobs/346280883#L1095

We can cherry-pick this into stable, which means it will be in the default docs rather than it technically making it into the "1.0" release.

nabobalis:
Also remove the doc_fix branch.

DanRyanIrish:
@Cadair How many spaces should the indentation be?  I've tried a number of options and they've all errored.

Cadair:
I think you need a line break after the `.. code-block` line.

PR #80, opened 2018-03-02
Update get_dependent_axes implementation.

DanRyanIrish:
This PR updates the API and implementation of ```get_dependent_axes()``` to use method closer to that which is due to be merged into ```astropy``` from ```glue``.  In doing so it fixes a bug in ```NDCube.axis_world_coords``` which occurred when there was a missing axis.


ndcube/utils/wcs.py
```diff
@@ -285,42 +285,82 @@ def reindex_wcs(wcs, inds):
...
+    # Remove "missing" axes from output.
+    non_missing_wcs_dependent_axes = wcs_dependent_axes[
+        np.invert(missing_axis)[wcs_dependent_axes]]
+    # Convert dependent axes back to numpy/data ordering.
+    dependent_data_axes = tuple([utils_cube.wcs_axis_to_data_axis(i, missing_axis)
+                                 for i in non_missing_wcs_dependent_axes])
+    return dependent_data_axes
+
+
+def get_dependent_wcs_axes(wcs_object, wcs_axis, missing_axis):
```
Cadair: is `missing_axis` actually being used in this function?

DanRyanIrish: No.  That's for the spot.

Just working on tests now.

PR #81, opened 2018-03-02
Enable 1D Animation of NDCube

DanRyanIrish:
Add capability to ```NDCube.plot``` to animate a cube as a 1D line animation with sliders for higher dimension axes.  Currently on 2D animations are possible and only when number of cube dimensions are > 2.

This PR depends on PR sunpy/sunpy#2491 being merged.

The PR also includes some minor enhancements of the other ```NDCube.plot``` options.


DanRyanIrish:
Let's not merge this until the release of ```sunpy``` 0.9 as it depends on features added by sunpy/sunpy#2491 which currently is only in sunpy/master.  If the release of 0.9 is significantly delayed, we can revisit this idea.

DanRyanIrish:
Make user facing API of ```NDCube.plot``` to match API of PR #98 and keep backwards compatibility by popping kwargs.

PR #84, opened 2018-03-08
fix the dimension checking logic

Cadair:
A very small bug fix I found

### DIFF
--- a/ndcube/ndcube.py
+++ b/ndcube/ndcube.py
@@ -405,7 +405,7 @@ class NDCube(NDCubeSlicingMixin, NDCubePlotMixin, astropy.nddata.NDArithmeticMix
         # The docstring is defined in NDDataBase
 
         n_dim = len(self.dimensions)
-        if len(min_coord_values) != len(interval_widths) != n_dim:
+        if (len(min_coord_values) != len(interval_widths)) or len(min_coord_values) != n_dim:
             raise ValueError("min_coord_values and interval_widths must have "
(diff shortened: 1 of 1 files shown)


ndcube/ndcube.py
```diff
@@ -405,7 +405,7 @@ def crop_by_coords(self, min_coord_values, interval_widths):
         # The docstring is defined in NDDataBase
 
         n_dim = len(self.dimensions)
-        if len(min_coord_values) != len(interval_widths) != n_dim:
+        if (len(min_coord_values) != len(interval_widths)) or len(min_coord_values) != n_dim:
```
DanRyanIrish: What's the difference in the logic of these two implementations?

Cadair: one works the other dosen't :laughing: 

I can't immediately work out why the first one dosen't work

DanRyanIrish: But ```NDCube.dimensions``` doesn't crash.  Do you mean that this if statement will never be entered or at least isn't entered when it should?

Cadair:
```python
>>> d1 = 3
>>> d2 = 3
>>> nd = 4
>>> d1 != d2 != nd
False
>>> nd = 3
>>> d1 != d2 != nd
```

DanRyanIrish: Bizarre!

DanRyanIrish:
lgtm pending Travis.

Issue #88, opened 2018-03-13
NDCube Should Always Have a Mask


DanRyanIrish:
```NDCube``` should always have a mask so that operations can assume one exists without checking.  If a mask is not supplied during instantiation, a False mask should be generated.  This will make operations over ```NDCubeSequence```s not have to always check that a mask exists.

PR #89, opened 2018-03-14
Give NDCube a False mask on instantiation if one not provided by user.

DanRyanIrish:
This PR addresses issue #88.


DanRyanIrish:
@Cadair, do you see any reason why ```NDCube``` should not always have a mask?

Cadair:
I think always creating one like this on instantiation is not great as it's doubling the memory usage of ndcube when it might not need to.

I think a property that returns a false mask when accessed would be a much better idea.

DanRyanIrish:
Having reflected in a state not exhausted on a plane (:P) I think you're right.

Issue #94, opened 2018-03-16
Non-existent function called on ndcube/visualization/animation.py


BaptistePellorceAstro:
The `ndcube.visualization.animation`'s function is calling a non-existent function named `ndcube.utils.sequence._convert_cube_like_index_to_sequence_indices`.

I found this bug by running the Dan Ryan's ndcube tutorial at the 118th line.
You can see the tutorial here [https://github.com/DanRyanIrish/notebooks](url)

PR #95, opened 2018-03-20
Fix bug in updating plot in NDCubeSequence animation.

DanRyanIrish:
Deprecated function name replaced with correct version.


Cadair:
closes #94

Issue #100, opened 2018-03-22
crop_by_coords is broken when data is rotated


Cadair:
I am a little confused why the implementation of `crop_by_coords` is so different, both in API and in code from `GenericMap.submap`. https://github.com/sunpy/sunpy/blob/master/sunpy/map/mapbase.py#L1273

@DanRyanIrish ?

DanRyanIrish:
Issue 100!!!!  🍾 

I am puzzled by this too.  Our original idea was to make the API of ```crop_by_coords``` like ```GenericMap.submap```.  I agree this is something that should be fixed.

Cadair:
So we don't break the api of `crop_by_coords` in 1.1 I propose we do the following to maintain parity with `submap` as much as possible:

I propose the method have the following signature:
```python
def crop_by_coords(bottom_left, interval_widths=None, top_right=None):
```
Where only one of `interval_widths` and `top_right` can be specified. (or neither if the shape of the first argument is `(2, N)`.)

We should then add a pending deprecation warning for `interval_widths` for 2.0.


Also, we should make sure we use the same *internal* logic as `submap` where it calculates the pixel positions of the four corners of the physical coordinates and then takes the maximum, so as @tiagopereira says it takes the maximal crop not the minimal one. This second fix should be backported to 1.0.2.

DanRyanIrish:
Conversation taken from IRISpy chatroom:
Danny: Hi @Cadair . @BaptistePellorceAstro has started working on sunpy/ndcube#100. We should discuss/lay out what the API is to him/in the issue before he gets too far.  I was a little confused about the current sunpy ```GenericMap.submap``` API only having ```bottom left``` and ```top right``` kwargs. It seems to me that this does not result in a unique solution to the desired region of interest.
There is a valid solution aligned with the pixel grid and at an angle to the pixel grid. Surely, we need at least one more corner to make this unique.

@Cadair: how so? why? Are you are specifying a rectangle in physical coordinates?  Then calculating a pixel box that encompasses the whole of that rectangle

Danny: You mean both possible rectangles?  If the user supplies 4 corners, then the ```crop_by_coords``` can determine the minimum pixel-grid-aligned rectangle that includes them all.  Then a more advanced version could even mask out regions that in that minimum pixel-grid-aligned rectangle not within the user-provided corners.

Cadair:
[...]

Issue #99, opened 2018-03-22
Lack of rotation when plotting rotated fields of view with NDCube


BaptistePellorceAstro:
By testing some data to check if coordinates are well represented with NDCube instance, we found that for rotated fields of view, X and Y axis are not representing usual coordinates that we use. We tried with [iris_l2_20150304_160141_3824263489_SJI_1400_t000](http://www.lmsal.com/solarsoft/irisa/data/level2_compressed/2015/03/04/20150304_160141_3824263489/iris_l2_20150304_160141_3824263489_SJI_1400_t000.fits.gz) data coming from a 45 degree's rotated field of view, and you can see, below, the result we have. In this case, the images should be rotated by 45 degrees counter-clockwise.
![iris_l2_20150304_160141_3824263489_SJI_1400_t000.fits](https://user-images.githubusercontent.com/37335327/37760745-d9fa684e-2db7-11e8-8acc-6b315848769b.png)

Cadair:
If you look closely the ticks are rotated, so the plot knows the coordinate system does not align to the image. 

What needs changing here is the defaults for the tick labels and probably a faint grid to make it more obvious that the image and the coordinate system are not aligned.

The plot will not and should not rotate the image to the coordinate system.

Cadair:
Having said that, work is currently being done on astropy to make WCSAxes default to standard decimal notation for arcsec by default.

tiagopereira:
@Cadair so this is really an issue with WCSAxes and not ndcube? Maybe this is not a problem for astronomical images, but for solar observations I think it would be really useful to have the option of rotating the image so that the coordinates are aligned with the figure axes. It is very common to work with multi-telescope observations, and superimposing two observations on the same coordinate system is really useful.

Cadair:
@tiagopereira overlaying non-aligned data is not an operation that WCSAxes / matplotlib can or should just handle, it's one for a package like [reproject](https://github.com/astrofrog/reproject) which you can use to do that for display purposes with solar data.

I think there is an NDCube issue here, in that we should have some more sane defaults for the tick labels and grids etc, i.e. match the default settings of `Map.plot` in sunpy core.

tiagopereira:
[...]

PR #113, opened 2018-04-18
Solving crop_by_coord problem when data are rotated

BaptistePellorceAstro:
Here a first version to solve the `crop_by_coords` problem when the data are rotated.

Discussion about it was on issue #99 and issue #100 

Ping @DanRyanIrish

### DIFF
--- a/ndcube/ndcube.py
+++ b/ndcube/ndcube.py
@@ -87,28 +87,34 @@ class NDCubeABC(astropy.nddata.NDData, metaclass=NDCubeMetaClass):
         pass
 
     @abc.abstractmethod
-    def crop_by_coords(self, lower_corner, interval_widths=None, upper_corner=None):
+    def crop_by_coords(self, lower_corner, interval_widths=None, upper_corner=None, units=None):
         """
         Crops an NDCube given minimum values and interval widths along axes.
 
         Parameters
         ----------
-        lower_corner: iterable of `astropy.units.Quantity`
+        lower_corner: iterable of `astropy.units.Quantity` or `float`
             The minimum desired values along each relevant axis after cropping
             described in physical units consistent with the NDCube's wcs object.
             The length of the iterable must equal the number of data dimensions
             and must have the same order as the data.
 
-        interval_widths: iterable of `astropy.units.Quantity`
+        interval_widths: iterable of `astropy.units.Quantity` or `float`
             The width of the region of interest in each dimension in physical
             units consistent with the NDCube's wcs object. The length of the
             iterable must equal the number of data dimensions and must have
             the same order as the data. This argument will be removed in versions
             2.0, please use upper_corner argument.
 
-        upper_corner: iterable of `astropy.units.Quantity`
+        upper_corner: iterable of `astropy.units.Quantity` or `float`
             The maximum desired values along each relevant axis after cropping
             described in physical units consistent with the NDCube's wcs object.
+            The length of the iterable must equal the number of data dimensions
+            and must have the same order as the data.
+
+        units: iterable of `astropy.units.quantity.Quantity`, optionnal
+            If the inputs are set without units, the user must set the units
+            inside this argument as `str`.
             The length of the iterable must equal the number of data dimensions
             and must have the same order as the data.
 
@@ -419,7 +425,7 @@ class NDCubeBase(NDCubeSlicingMixin, NDCubeABC):
                     "value": self._extra_coords_wcs_axis[key]["value"]}
         return result
 
-    def crop_by_coords(self, lower_corner, interval_widths=None, upper_corner=None):
+    def crop_by_coords(self, lower_corner, interval_widths=None, upper_corner=None, units=None):
         # The docstring is defined in NDDataBase
 
         n_dim = self.data.ndim
@@ -442,6 +448,23 @@ class NDCubeBase(NDCubeSlicingMixin, NDCubeABC):
         if (len(lower_corner) != len(upper_corner)) or (len(lower_corner) != n_dim):
             raise ValueError("lower_corner and upper_corner must have same"
                              "number of elements as number of data dimensions.")
+        if units:
+            # Raising a value error if units have not the data dimensions.
+            if len(units) != n_dim:
+                raise ValueError('units must have same number of elements as '
+                                 'number of data dimensions.')
+            # If inputs are not Quantity objects, they are modified into specified units
+            for i in range(n_dim):
+                if type(lower_corner[i]) is not u.quantity.Quantity:
+                    lower_corner[i] = u.Quantity(lower_corner[i], unit=units[i])
+                if type(upper_corner[i]) is not u.quantity.Quantity:
+                    upper_corner[i] = u.Quantity(upper_corner[i], unit=units[i])
+        else:
+            for i in range(n_dim):
+                if type(lower_corner[i]) is not u.quantity.Quantity or \
+                   type(upper_corner[i]) is not u.quantity.Quantity:
+                    raise ValueError("The inputs are not Quantity objects, you "
+                                     "can use the units argument to set it")
         # Derive all corners coordinates
         quantity_list = [[lower_corner[i], upper_corner[i]] for i in range(n_dim)]
         all_corners = [self.world_to_pixel(*a) for a in product(*quantity_list)]
--- a/ndcube/ndcube.py
+++ b/ndcube/ndcube.py
@@ -463,24 +463,19 @@ class NDCubeBase(NDCubeSlicingMixin, NDCubeABC):
                 raise TypeError("lower_corner and interval_widths/upper_corner must be "
                                 "of type astropy.units.Quantity or the units kwarg "
                                 "must be set.")
-        # Derive all corners coordinates
-        quantity_list = [[lower_corner[i], upper_corner[i]] for i in range(n_dim)]
-        all_corners = [self.world_to_pixel(*a) for a in product(*quantity_list)]
-        # Inputing of all corners coordinates inside a numpy array
-        # According to the boundary conditions
-        corners_array = np.zeros((2**n_dim, n_dim))
-        for i in range(2**n_dim):
-            for j in range(n_dim):
-                corners_array[i, j] = u.Quantity(all_corners[i][j]).value
-                if corners_array[i, j] > self.data.shape[j]:
-                    corners_array[i, j] = self.data.shape[j]
-                if corners_array[i, j] < 0:
-                    corners_array[i, j] = 0
-        # Taking the maximum and minimum values of coordinates
-        lower_pixels = corners_array.min(0)
-        upper_pixels = corners_array.max(0)
-        # Creating a tuple to crop the data with inputed coordinates
-        item = tuple([slice(int(lower_pixels[i]), int(upper_pixels[i]) + 1) for i in range(n_dim)])
+        # Get all corners of region of interest.
+        all_world_corners_grid = np.meshgrid(*[u.Quantity([lower_corner[i], upper_corner[i]],
+                                                          unit=lower_corner[i].unit).value
+                                               for i in range(self.data.ndim)])
+        all_world_corners = [all_world_corners_grid[i].flatten()*lower_corner[i].unit
+                             for i in range(n_dim)]
+        # Convert to pixel coordinates
+        all_pix_corners = self.world_to_pixel(*all_world_corners)
+        # Derive slicing item with which to slice NDCube.
+        # Be sure to round down min pixel and round up + 1 the max pixel.
+        item = tuple([slice(int(axis_pixels.value.min()),
+                            int(np.ceil(axis_pixels.value.max()))+1)
+                      for axis_pixels in all_pix_corners])
         return self[item]
 
     def crop_by_extra_coord(self, min_coord_value, interval_width, coord_name):
--- a/ndcube/ndcube.py
+++ b/ndcube/ndcube.py
@@ -473,7 +473,7 @@ class NDCubeBase(NDCubeSlicingMixin, NDCubeABC):
         all_pix_corners = self.world_to_pixel(*all_world_corners)
         # Derive slicing item with which to slice NDCube.
         # Be sure to round down min pixel and round up + 1 the max pixel.
-        item = tuple([slice(int(axis_pixels.value.min()),
+        item = tuple([slice(int(np.clip(axis_pixels.value.min(), 0, None)),
                             int(np.ceil(axis_pixels.value.max()))+1)
                       for axis_pixels in all_pix_corners])
         return self[item]
--- a/ndcube/ndcube.py
+++ b/ndcube/ndcube.py
@@ -1,6 +1,8 @@
 # -*- coding: utf-8 -*-
 
 import abc
+import warnings
+from itertools import product
 
 import numpy as np
 import astropy.nddata
@@ -85,22 +87,30 @@ class NDCubeABC(astropy.nddata.NDData, metaclass=NDCubeMetaClass):
         pass
 
     @abc.abstractmethod
-    def crop_by_coords(self, min_coord_values, interval_widths):
+    def crop_by_coords(self, lower_corner, interval_widths=None, upper_corner=None):
         """
         Crops an NDCube given minimum values and interval widths along axes.
 
         Parameters
         ----------
-        min_coord_values: iterable of `astropy.units.Quantity`
+        lower_corner: iterable of `astropy.units.Quantity`
             The minimum desired values along each relevant axis after cropping
             described in physical units consistent with the NDCube's wcs object.
-            The length of the iterable must equal the number of data dimensions and must
-            have the same order as the data.
+            The length of the iterable must equal the number of data dimensions
+            and must have the same order as the data.
 
         interval_widths: iterable of `astropy.units.Quantity`
-            The width of the region of interest in each dimension in physical units
-            consistent with the NDCube's wcs object.  The length of the iterable must
-            equal the number of data dimensions and must have the same order as the data.
+            The width of the region of interest in each dimension in physical
+            units consistent with the NDCube's wcs object. The length of the
+            iterable must equal the number of data dimensions and must have
+            the same order as the data. This argument will be removed in versions
+            2.0, please use upper_corner argument.
+
+        upper_corner: iterable of `astropy.units.Quantity`
+            The maximum desired values along each relevant axis after cropping
+            described in physical units consistent with the NDCube's wcs object.
+            The length of the iterable must equal the number of data dimensions
+            and must have the same order as the data.
 
         Returns
         -------
@@ -409,21 +419,46 @@ class NDCubeBase(NDCubeSlicingMixin, NDCubeABC):
                     "value": self._extra_coords_wcs_axis[key]["value"]}
         return result
 
-    def crop_by_coords(self, min_coord_values, interval_widths):
+    def crop_by_coords(self, lower_corner, interval_widths=None, upper_corner=None):
         # The docstring is defined in NDDataBase
 
         n_dim = len(self.dimensions)
-        if (len(min_coord_values) != len(interval_widths)) or len(min_coord_values) != n_dim:
-            raise ValueError("min_coord_values and interval_widths must have "
-                             "same number of elements as number of data dimensions.")
-        # Convert coords of lower left corner to pixel units.
-        lower_pixels = self.world_to_pixel(*min_coord_values)
-        upper_pixels = self.world_to_pixel(*[min_coord_values[i] + interval_widths[i]
-                                             for i in range(n_dim)])
-        # Round pixel values to nearest integer.
-        lower_pixels = [int(np.rint(l.value)) for l in lower_pixels]
-        upper_pixels = [int(np.rint(u.value)) for u in upper_pixels]
-        item = tuple([slice(lower_pixels[i], upper_pixels[i]) for i in range(n_dim)])
+        # Raising a value error if the arguments have not the same dimensions.
+        if upper_corner:
+            if (len(lower_corner) != len(upper_corner)) or (len(lower_corner) != n_dim):
+                raise ValueError("lower_corner and upper_corner must have "
+                                 "same number of elements as number of data "
+                                 "dimensions.")
+        # Raising a value error if the arguments have not the same dimensions.
+        # Calculation of upper_corner with the inputing interval_widths
+        # This part of the code will be removed in version 2.0
+        if interval_widths:
+            warnings.warn("interval_widths will be removed from the API in "
+                          "version 2.0, please use upper_corner argument.")
+            if (len(lower_corner) != len(interval_widths)) or (len(lower_corner) != n_dim):
+                raise ValueError("lower_corner and interval_widths must have "
+                                 "same number of elements as number of data "
+                                 "dimensions.")
+            upper_corner = [lower_corner[i] + interval_widths[i] for i in range(n_dim)]
+        # Derive all corners coordinates
+        quantity_list = [[lower_corner[i], upper_corner[i]] for i in range(n_dim)]
+        all_corners = [self.world_to_pixel(*a) for a in product(*quantity_list)]
+        # Inputing of all corners coordinates inside a numpy array
+        # According to the boundary conditions
+        corners_array = np.zeros((2**n_dim, n_dim))
+        for i in range(2**n_dim):
+            for j in range(n_dim):
+                corners_array[i, j] = u.Quantity(all_corners[i][j]).value
+                if corners_array[i, j] > self.data.shape[j]:
+                    corners_array[i, j] = self.data.shape[j]
+                if corners_array[i, j] < 0:
+                    corners_array[i, j] = 0
+        # Taking the maximum and minimum values of coordinates
+        lower_pixels = corners_array.min(0)
+        upper_pixels = corners_array.max(0)
+        # Creating a tuple to crop the data with inputed coordinates
+        item = tuple([slice(int(lower_pixels[i]), int(upper_pixels[i])) for i in range(n_dim)])
+
         return self[item]
 
     def crop_by_extra_coord(self, min_coord_value, interval_width, coord_name):
@@ -502,7 +537,7 @@ class NDCubeOrdered(NDCube):
         for standard deviation or "var" for variance. A metaclass defining
         such an interface is NDUncertainty - but isn’t mandatory. If the uncertainty
         has no such attribute the uncertainty is stored as UnknownUncertainty.
-        Defaults to None.
+        Defaults to None.list
 
     mask : any type, optional
         Mask for the dataset. Masks should follow the numpy convention
--- a/ndcube/ndcube.py
+++ b/ndcube/ndcube.py
@@ -422,24 +422,27 @@ class NDCubeBase(NDCubeSlicingMixin, NDCubeABC):
     def crop_by_coords(self, lower_corner, interval_widths=None, upper_corner=None):
         # The docstring is defined in NDDataBase
 
-        n_dim = len(self.dimensions)
-        # Raising a value error if the arguments have not the same dimensions.
-        if upper_corner:
-            if (len(lower_corner) != len(upper_corner)) or (len(lower_corner) != n_dim):
-                raise ValueError("lower_corner and upper_corner must have "
-                                 "same number of elements as number of data "
-                                 "dimensions.")
+        n_dim = self.data.ndim
         # Raising a value error if the arguments have not the same dimensions.
         # Calculation of upper_corner with the inputing interval_widths
         # This part of the code will be removed in version 2.0
         if interval_widths:
             warnings.warn("interval_widths will be removed from the API in "
                           "version 2.0, please use upper_corner argument.")
+            if upper_corner:
+                raise ValueError("Only one of interval_widths or upper_corner "
+                                 "can be set. Recommend using upper_corner as "
+                                 "interval_widths is deprecated.")
             if (len(lower_corner) != len(interval_widths)) or (len(lower_corner) != n_dim):
                 raise ValueError("lower_corner and interval_widths must have "
                                  "same number of elements as number of data "
                                  "dimensions.")
-            upper_corner = [lower_corner[i] + interval_widths[i] for i in range(n_dim)]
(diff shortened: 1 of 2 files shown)


ndcube/ndcube.py
```diff
@@ -502,7 +537,7 @@ class NDCubeOrdered(NDCube):
         for standard deviation or "var" for variance. A metaclass defining
         such an interface is NDUncertainty - but isn’t mandatory. If the uncertainty
         has no such attribute the uncertainty is stored as UnknownUncertainty.
-        Defaults to None.
+        Defaults to None.list
```
BaptistePellorceAstro: Mistake removed from the file

ndcube/ndcube.py
```diff
@@ -85,22 +87,30 @@ def world_axis_physical_types(self):
         pass
 
     @abc.abstractmethod
-    def crop_by_coords(self, min_coord_values, interval_widths):
+    def crop_by_coords(self, lower_corner, interval_widths=None, upper_corner=None):
```
Cadair: An API point here, I think it would be worth moving `interval_widths` into `**kwargs` so that it is no longer part of the function signature. We can still support it by doing a `kwargs.pop('interval_widths', None)` as the ~first line of the method.

BaptistePellorceAstro: Hi, @DanRyanIrish told me to do it like that because some people will use this version as the previous : by inputting `corp_by_coords( (low1, low2), (inter1, inter2) )`. 

If I move `interval_widths` into `**kwargs`, I am afraid that people use `upper_corner` instead of `interval_widths` and the result will not be correct. 

I don't know if you saw it but I add a warning about using `interval_widths` where I ask the user to use `upper_corner`.

Of course I am open to every modification and if you want I change it, I'll do it !

Cadair: ah yes, you are correct. never mind then.

tiagopereira:
Since we are contemplating an API change, can we perhaps consider moving the units into a separate tuple on things like `crop_by_coords` and `world_to_pixel`? Maybe it is just me, but the following, taken from the documentation, seems exceedingly verbose and too cumbersome to write on a terminal session:
```
>>> my_cube_roi = my_cube.crop_by_coords(
... [Quantity(0.7, unit="deg"), Quantity(1.3e-5, unit="deg"), Quantity(1.04e-9, unit="m")],
... [Quantity(0.6, unit="deg"), Quantity(1., unit="deg"), Quantity(0.08e-9, unit="m")])
```

Why not have the units on a separate argument, e.g.:
```
>>> my_cube_roi = my_cube.crop_by_coords([0.7, 1.3e5, 1.04e9], [0.6, 1., 0.08e-9], 
...                                      units=["deg", "deg", "m"])
```
This seems much more readable. If `my_cube` already has units assigned to each dimension, we can even make the `units` argument optional and use the units of `my_cube` by default. We can also add in a check if each item on the list is a `Quantity`, so people can use the old API if they want to.

Cadair:
@tiagopereira your first example is a little more verbose than it needs be the following would also work:
```python
>>> import astropy.units as u
>>> my_cube_roi = my_cube.crop_by_coords(
... [*(0.7, 1.3e-5)*u.deg, 1.04e-9*u.m],
... [*(0.6, 1.)*u.deg, 0.08e-9*u.m])
```

I am however not adverse to a `units=` keyword argument, as long as the units are always explicitly provided by the user. This is the API used by `SkyCoord` as well, in that if the numbers are provided dimensionless there must be a `units=` keyword arg that is of the same number of dimensions as the input.

(sorry for accidentally posting and closing the issue lol)

DanRyanIrish:
I believe the current API is better for philosophical reasons, namely that all values of a physical quantity should be represented with units attached.  I believe it is also more useful in many cases.  However, in the case where the user is entering these values manually on the command line, I agree it does require duplication of the unit definition in the corner arguments.  So I am open to discussing whether we support this additional API simultaneous with the current one.

My main concern about this API from a coding point of view, is that it introduces a potentially unnecessary for loop at the start of the method where the ```Quantity``` objects are created.  This could become a noticeable penalty for high dimensions?  Perhaps this is not a major penalty relative to the run time of the rest of the method though.  I'm not sure.

I agree with @Cadair that it should not have a default unit, but rather require explicit definition.  The default would be derived from the WCS units, which are in SI.  These are not the typical units used in solar physics so the majority of solar users will have to explicitly give the units anyway.  It also leaves open the chance that users will be confused and frustrated when they enter values in arcsec, nm or erg without defining the units and the cropping happens in deg, m, or J.

ndcube/ndcube.py
```diff
@@ -409,21 +419,46 @@ def extra_coords(self):
...
+        # Calculation of upper_corner with the inputing interval_widths
+        # This part of the code will be removed in version 2.0
+        if interval_widths:
+            warnings.warn("interval_widths will be removed from the API in "
+                          "version 2.0, please use upper_corner argument.")
+            if (len(lower_corner) != len(interval_widths)) or (len(lower_corner) != n_dim):
+                raise ValueError("lower_corner and interval_widths must have "
+                                 "same number of elements as number of data "
+                                 "dimensions.")
+            upper_corner = [lower_corner[i] + interval_widths[i] for i in range(n_dim)]
```
DanRyanIrish: I think it's better if you reverse the order of these ```if``` statements.  That way you can issue the warning and construct ```upper_corner``` from ```interval_widths``` without checking the number of elements is right.  Then you can just check the length of ```upper_corner``` without requiring an ```if``` statement since if it wasn't defined by the user, you will have defined in the ```if interval_widths:``` if statement.

Then when it comes time to stop supporting the ```interval_widths``` API, we just have to delete the ```if interval_widths:``` statement.

ndcube/ndcube.py
```diff
@@ -409,21 +419,46 @@ def extra_coords(self):
                     "value": self._extra_coords_wcs_axis[key]["value"]}
         return result
 
-    def crop_by_coords(self, min_coord_values, interval_widths):
+    def crop_by_coords(self, lower_corner, interval_widths=None, upper_corner=None):
         # The docstring is defined in NDDataBase
 
         n_dim = len(self.dimensions)
```
DanRyanIrish: I know this wasn't your change, but can you replace this is ```n_dim = self.data.ndim```?  It's more efficient.

DanRyanIrish:
In fact, I have a suggestion for third API option.  @Cadair wont like it because I don't see an easy way to maintain the current API and this one simultaenously.  But I think it makes the code more efficient and succinct.  It solves @tiagopereira's concern about duplicating the unit definition as my concern about not separating the values from the units.  Have the user supply N (number of dimensions) ```Quantity``` objects of length 2, each giving the min and max of the region of interest along the each axis.  So in @tiagopereira's example this would be:
```python
my_cube_roi = my_cube.crop_by_coords(Quantity([0.7, 1.3], unit="deg"),
                                     Quantity([1.3e-5, 1+1.3e-5], unit="deg"),
                                     Quantity([1.04e-9, 1.12e-9], unit="m"))
```
This is the same API as with ```world_to_pixel``` and ```pixel_to_world```.  As well as unifying those APIs, this API for ```crop_by_coords``` also means we need fewer for loops through the dimensions by eliminating the need to put the corners args into a format for ```world_to_pixel```.  It also means that the unit for each dimension only needs to input once.  Here's how I think it should work in code.  I'd be interested to hear if @Cadair thinks it really is more efficient though.  Let ```quantity_list``` be the list of all length 2 ```Quantity``` objects input by the user, a min and max for each dimension.
```python
def crop_by_coords(self, *quantity_list):
    # Get all corners of region of interest.  Perhaps there is a tidier/more efficient way of doing this.
    all_world_corners_grid = np.meshgrid(*[axis.value for axis in quantity_list])
    all_world_corners = [all_world_corners_grid[i].flatten()*quantity_list[i].unit 
                         for i in range(n_dim)]
    # Convert to pixel coordinates
    all_pix_corners = self.world_to_pixel(*all_world_corners)
    # Derive slicing item with which to slice NDCube.  
    # Be sure to round down min pixel and round up + 1 the max pixel.
    item = tuple([slice(int(axis_pixels.value.min()), 
                        int(np.ceil(axis_pixels.value.max()))+1) 
                  for axis_pixels in all_pix_corners])
    return self[item]
```
As you can see, ```crop_by_coords``` then becomes a very short method!

Cadair:
@DanRyanIrish you are quite correct I won't like it. However, not really because of the API change.

The two main objections I have are:

1. It is not the same as SunPy Map. Having changed the API for sunpy map in 0.8 I really don't want to have to change it again.
2. This API is not conducive to ever moving to allowing rich coordinate objects as inputs. i.e. when the APE 14 stuff has progressed I see this API moving to allowing SkyCoord, Time and Quantity as input, much the same way that `Map.submap` does in SunPy at the moment.

ndcube/ndcube.py
BaptistePellorceAstro: Changed

ndcube/ndcube.py
BaptistePellorceAstro: If i delete the elements number checking between `lower_corner`, `interval_widths` and `n_dim`, an error will come by creating the `upper_corner` values because one of these tuples will be out of range. Do you want I remove this error raising ?

DanRyanIrish: Ah, you're absolutely right!  Yes, leave in the length checking in both statements, but check ```interval_widths``` first.  Then you can also add in a check that ```interval_widths``` and ```upper_corner``` were not both set:
```python
# Calculation of upper_corner with the inputing interval_widths
# This part of the code will be removed in version 2.0
if interval_widths:
    warnings.warn("interval_widths will be removed from the API in "
                  "version 2.0, please use upper_corner argument.")
    # If both intervals_widths and upper_corner set, raise error.
    if upper_corner:
        raise ValueError("Only one of interval_widths or upper_corner can be set.  Recommend using upper_corner as interval_widths is deprecated.")
    if (len(lower_corner) != len(interval_widths)) or (len(lower_corner) != n_dim):
        raise ValueError("lower_corner and interval_widths must have "
                         "same number of elements as number of data "
                         "dimensions.")
    upper_corner = [lower_corner[i] + interval_widths[i] for i in range(n_dim)]
# Raising a value error if the arguments have not the same dimensions.
if (len(lower_corner) != len(upper_corner)) or (len(lower_corner) != n_dim):
                raise ValueError("lower_corner and upper_corner must have "
                                 "same number of elements as number of data "
                                 "dimensions.")

Comment (BaptistePellorceAstro) [reply ndcube/ndcube.py]: Ok, I will input this line in my code, I already exchanged the if statements positions.

**Comment by DanRyanIrish** [comment]
The extension to ```SkyCoord``` and ```Time``` is a good point.  In fact,  think we can still make the above implementation work without adding an extra loop, while keeping the ```lower_corner```, ```upper_corner``` API.    See below.  It doesn't harmonize the concerns over the unit duplication in the API.  But I am open to changing that as a separate issue.
```python
def crop_by_coords(self, lower_corner, upper_corner):
    # Get all corners of region of interest.  Perhaps there is a tidier/more efficient way of doing this?
    all_world_corners_grid = np.meshgrid(
        *[Quantity([lower_corner[i], upper_corner[i]], unit=lower_corner[i].unit).value 
          for i in range(self.data.ndim)])
    all_world_corners = [all_world_corners_grid[i].flatten()*lower_corner[i].unit 
                         for i in range(n_dim)]
    # Convert to pixel coordinates
    all_pix_corners = self.world_to_pixel(*all_world_corners)
    # Derive slicing item with which to slice NDCube.  
    # Be sure to round down min pixel and round up + 1 the max pixel.
    item = tuple([slice(int(axis_pixels.value.min()), 
                        int(np.ceil(axis_pixels.value.max()))+1) 
                  for axis_pixels in all_pix_corners])
    return self[item]
```

ndcube/ndcube.py
```diff
@@ -457,8 +460,8 @@ def crop_by_coords(self, lower_corner, interval_widths=None, upper_corner=None):
         lower_pixels = corners_array.min(0)
         upper_pixels = corners_array.max(0)
         # Creating a tuple to crop the data with inputed coordinates
-        item = tuple([slice(int(lower_pixels[i]), int(upper_pixels[i])) for i in range(n_dim)])
-
+        item = tuple([slice(int(lower_pixels[i]), int(upper_pixels[i]))
+                      for i in range(n_dim)])
```
DanRyanIrish: Just out of curiosity, why are you breaking up this line and others like it into two?  According to the PEP8 standard, lines of code can be up to 99 characters long.  It doesn't affect the functionality.  It's just a coding style question.

BaptistePellorceAstro: I am coding on Atom and I have a line at 70 characters that I have to avoid to exceed. 
Then I am coding with three windows on my screen so I only can see around 75 characters. This is better for me. If you want, I can modify it but I think that for user with a lot of code windows, it is hard to see everything.

DanRyanIrish: I think it's general to stick with the PEP8 standard and have lines up to 99 characters.  The 70 line limit and size of your screen is specific to you and wont make it the code more readable for others.  Not every line has to be 99 characters of course.  Split lines where it makes sense.  But if one line is less than 99 characters, don't split it over two.  As I said, it's more readable and keeps the style standardised with the majority of other Python code out there.

BaptistePellorceAstro: Ok, I will change that ;)

DanRyanIrish:
Hi @BaptistePellorceAstro.  I know we are still discussing implementation and API changes so there is a bit of uncertainty.  But irrespective of that, ```crop_by_coords``` needs at least one test to check the right answer is obtained in the case where the coordinate system is not aligned with the pixel grid.  I suggest this should be the next step in this PR.

The tests for ```crop_by_coords``` can be found at https://github.com/sunpy/ndcube/blob/master/ndcube/tests/test_ndcube.py#L816.  The way the test works is that the ```@pytest.mark.parametrize``` decorator defines the test input and expected output as two tuples.  The test function then takes these tuples as input, calculates the result using the test input, and then compares the result with the expected result.  So what you need to do is the follow:
* Near the top of file, you need to generate an WCS object (see an example here: https://github.com/sunpy/ndcube/blob/master/ndcube/tests/test_ndcube.py#L19) which is misaligned with the pixel grid.
* Combine this with a data array and create a new ```NDCube```, called ```misaligned_cube```.  See an example here: https://github.com/sunpy/ndcube/blob/master/ndcube/tests/test_ndcube.py#L73.  Getting the WCS to misaligned is the tricky part.
* Pick a sub region you want from ```misaligned_cube``` and work out what the pixel slicing item should.
* Add a new tuple to the ```@pytest.mark.parametrize``` decorator that gives the test input and expected output in the same way as the existing tuple does.

ndcube/ndcube.py
tiagopereira: PEP8 says "[Limit all lines to a maximum of 79 characters.](https://www.python.org/dev/peps/pep-0008/#maximum-line-length)" This is a matter of preference, but I think it is a bad idea to go over 80. If the python standard library can make it under 80 characters, I don't see why we shouldn't be able to. The vertical line in Atom should default to 80.

Cadair: The general policy in SunPy core is 99 since they added this to PEP8:

> For code maintained exclusively or primarily by a team that can reach agreement on this issue, it is okay to increase the nominal line length from 80 to 100 characters (effectively increasing the maximum length to 99 characters), provided that comments and docstrings are still wrapped at 72 characters.

I leave the decision up to @DanRyanIrish though.

BaptistePellorceAstro:
Hi, I follow all you ask me to do and now I want to try the test to see if it is working before push my work. can you explain me how to do ?

BaptistePellorceAstro:
I succeed to run the test but there is an error with the previous data from the test. I will try to understand why. Do you want I push ?

BaptistePellorceAstro:
Ok, I have the same return : 
`test_crop_by_coords[test_input0-expected0]`
`1 failed, 257 passed, 3 skipped in 6.49 seconds`

So what I understand is that the first tuple of data is failing in the test no ?

DanRyanIrish:
Sure.  There's never harm in pushing.  The error in the last test may be because you've changed something.  So yes, it is important to understand why that is now failing.

BaptistePellorceAstro:
I think I have found the problem. We have :
`lower_corner = [0.7*u.deg, 1.3e-5*u.deg, 1.02e-9*u.m]` 
`interval_widths = [1*u.deg, 1*u.deg, 1.06*u.m]`

Firstly, they ask to crop the X axis from 1nm to 1m. This is a wide crop, so data are not crop in this axis because the data shape in X axis is 4 pixels and values are increasing by step of 0.02 nm per pixel.

Secondly, they ask to crop the Z axis from 0.7 deg to 1.7 deg. This is also a wide crop for the data because the data shape for this axis is only 2 pixels. By step of 0.4 deg as they ask, they are out of range (they need at least 4 pixels). One thing I saw here and on my test, the plotting function seems to be incorrect : in this axis, one pixel is missing in the plot. Do you know why ?

Then, they ask to crop the Y axis from 1.3e-5 deg to 1 deg. This crop is taking 3 pixels because the step is 0.5 deg per pixel. However 3 pixel is the shape of the data in Y axis.

To conclude, this crop is useless and crop nothing from these data. However, they ask a return as `cubem[:, :2]` but the input values don't crop the data. In my mind, there is a mistake in this test. Can you run it with your files to see if you have this bug ? What do you think about this crop ?

DanRyanIrish:
This analysis seems reasonable.  Feel free to adjust the inputs to the first test to make a more reasonable check of this method in the case where the coordinate and pixel grids align.

As for the plotting, don't worry about it.  There are open PRs which will be merged once sunpy 0.9 is released and will make major changes to the plotting.  Energy chasing down that bug at this time probably isn't worth it.

BaptistePellorceAstro:
Nice, I was just wanting to return you that. I will change the first inputs values to solve this test.

DanRyanIrish:
Good job!

BaptistePellorceAstro:
Hi, I have modified the inputs and the output expected for the align grid test. I am limited by selecting a part of data due to a plotting error. As you said before, I don't take care of that and I focused on the smaller part I can select.

DanRyanIrish:
You shouldn't have to plot to determine the region you're interested in.  Use can see the real world coordinates of the axes of the cube by using ```NDCube.axes_world_coords```.  If you do want to plot though, you can manually plot data from the cube by using ```matplotlib.pyplot.imshow()``` for 2D images, ```matplotlib.pyplot.plot``` for 1D line plots, and ```sunpy.visualization.imageanimator.ImageAnimatorWCS``` for 2D animations.  So the plotting bug shouldn't restrict you.

ndcube/tests/test_ndcube.py
```diff
@@ -837,7 +837,7 @@ def test_world_to_pixel(test_input, expected):
 
 
 @pytest.mark.parametrize("test_input,expected", [
-    ((cubem, [0.7*u.deg, 1.3e-5*u.deg, 1.02e-9*u.m], [1*u.deg, 1*u.deg, 1.06*u.m]), cubem[:, :2]),
+    ((cubem, [0.7*u.deg, 1.3e-5*u.deg, 1.02e-9*u.m], [1*u.deg, 1*u.deg, 5.e-11*u.m]), cubem[:, :, :3]),
```
DanRyanIrish: Can you add a couple more test cases here to use the ```upper_corner``` input, rather than just ```interval_widths```?  To do this, you will have to add a third element to the existing test_input tuples of ```None```, i.e. ```(cubem, [0.7*u.deg, 1.3e-5*u.deg, 1.02e-9*u.m], [1*u.deg, 1*u.deg, 5.e-11*u.m], None)```.  In order to use ```upper_corner``` in the new cases, the second element of the new tuples will be ```None``` (```interval_widths```) while the third element will be the ```upper_corner``` value.

In addition, we need a whole new test, ```test_crop_by_coords_error```, which tests the correct errors are raised by the correct combination of inputs.  This is an easier test function to write as you don't have to work out the output.  See here for an example of such a test: https://github.com/sunpy/ndcube/blob/master/ndcube/tests/test_utils_wcs.py#L80

BaptistePellorceAstro: Ok, I will check that.

DanRyanIrish:
@BaptistePellorceAstro, @tiagopereira @Cadair: To go back to our discussion about the ```crop_by_coords``` API, I'm happy for to go along with @tiagopereira's suggested change.  Here's how I think it should work in detail:
* ```lower_corner``` and ```intervals_widths``` or ```upper_corner``` can be entered as ```Quantity``` objects of iterables of numbers of the same length as the number of data dimensions.
* If they are not ```Quantity``` objects, the ```units``` kwarg must be an iterable of valid unit strings or ```Unit``` objects.  Again ```units``` must also have the same length as the number of dimensions.
* If the ```units``` kwarg is set, then ```lower_corner``` and ```intervals_widths``` or ```upper_corner``` are converted to ```Quantity``` objects near the start of the method by doing something like:
```python
lower_corner = [u.Quantity(lower_corner[i], unit=units[i]) for i in range(self.data.ndim)]
```
and used as ```Quantity``` objects throughout the rest of the method.

Does this sound like it solves everyone's concerns?

BaptistePellorceAstro:
I have add a `units` argument in `crop_by_coords`, I added some lines and errors raising, and I create new tests.

ndcube/ndcube.py
```diff
@@ -409,21 +425,64 @@ def extra_coords(self):
...
+            # Raising a value error if units have not the data dimensions.
+            if len(units) != n_dim:
+                raise ValueError('units must have same number of elements as '
+                                 'number of data dimensions.')
+            # If inputs are not Quantity objects, they are modified into specified units
+            for i in range(n_dim):
+                if type(lower_corner[i]) is not u.quantity.Quantity:
+                    lower_corner[i] = u.Quantity(lower_corner[i], unit=units[i])
+                if type(upper_corner[i]) is not u.quantity.Quantity:
+                    upper_corner[i] = u.Quantity(upper_corner[i], unit=units[i])
```
DanRyanIrish: You don't have to check the types in this for loop.  If the type is already a ```Quantity``` the effect will be to convert the ```Quantity``` to the given unit.  This is probably a more intuitive behaviour if a user gives both ```Quantities``` and sets the ```units``` kwarg.

ndcube/ndcube.py
```diff
@@ -409,21 +425,64 @@ def extra_coords(self):
...
+        else:
+            for i in range(n_dim):
+                if type(lower_corner[i]) is not u.quantity.Quantity or \
+                   type(upper_corner[i]) is not u.quantity.Quantity:
+                    raise ValueError("The inputs are not Quantity objects, you "
+                                     "can use the units argument to set it")
```
DanRyanIrish: Suggest changing this error message slightly to something like: ```"lower_corner and interval_widths/upper_corner must be of type astropy.units.Quantity or the units kwarg must be set."```.
Alos, the type of error should be ```TypeError```, not ```ValueError```.

DanRyanIrish:
@BaptistePellorceAstro, this is looking very good.  I'll have another look at it a bit later.  But I think it's getting to be being ready to merge.

ndcube/ndcube.py
BaptistePellorceAstro: Oh, yes, I didn't think that it is working like that ! I go modify that.

DanRyanIrish:
One of the tests is now failing so something needs to be fixed.  Have a look at the online test report here: https://travis-ci.org/sunpy/ndcube/jobs/371049917

If you have questions I can answer them a bit later.

ndcube/ndcube.py
```diff
@@ -409,21 +425,64 @@ def extra_coords(self):
...
+                raise ValueError('units must have same number of elements as '
+                                 'number of data dimensions.')
+            # If inputs are not Quantity objects, they are modified into specified units
+            for i in range(n_dim):
+                if type(lower_corner[i]) is not u.quantity.Quantity:
+                    lower_corner[i] = u.Quantity(lower_corner[i], unit=units[i])
+                if type(upper_corner[i]) is not u.quantity.Quantity:
+                    upper_corner[i] = u.Quantity(upper_corner[i], unit=units[i])
+        else:
+            for i in range(n_dim):
```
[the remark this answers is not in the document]
BaptistePellorceAstro: Oh nice ! That is what I was looking for, thanks !

ndcube/ndcube.py
```diff
@@ -409,21 +425,64 @@ def extra_coords(self):
...
+            raise ValueError("lower_corner and upper_corner must have same"
+                             "number of elements as number of data dimensions.")
+        if units:
+            # Raising a value error if units have not the data dimensions.
+            if len(units) != n_dim:
+                raise ValueError('units must have same number of elements as '
+                                 'number of data dimensions.')
+            # If inputs are not Quantity objects, they are modified into specified units
+            for i in range(n_dim):
+                if type(lower_corner[i]) is not u.quantity.Quantity:
```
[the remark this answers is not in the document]
BaptistePellorceAstro: Outdated lines, changes are already made.

DanRyanIrish:
This is looking good @BaptistePellorceAstro.  Let us know when you feel it is ready for another review.

ndcube/ndcube.py
```diff
@@ -409,21 +425,62 @@ def extra_coords(self):
...
+                corners_array[i, j] = u.Quantity(all_corners[i][j]).value
+                if corners_array[i, j] > self.data.shape[j]:
+                    corners_array[i, j] = self.data.shape[j]
+                if corners_array[i, j] < 0:
+                    corners_array[i, j] = 0
+        # Taking the maximum and minimum values of coordinates
+        lower_pixels = corners_array.min(0)
+        upper_pixels = corners_array.max(0)
+        # Creating a tuple to crop the data with inputed coordinates
+        item = tuple([slice(int(lower_pixels[i]), int(upper_pixels[i]) + 1) for i in range(n_dim)])
```
DanRyanIrish: I think the algorithm from L466 to 483 should be replaced with the following:
```python
# Get all corners of region of interest.
all_world_corners_grid = np.meshgrid(*[u.Quantity([lower_corner[i], upper_corner[i]], unit=lower_corner[i].unit).value for i in range(self.data.ndim)])
all_world_corners = [all_world_corners_grid[i].flatten()*lower_corner[i].unit for i in range(n_dim)]
# Convert to pixel coordinates
all_pix_corners = self.world_to_pixel(*all_world_corners)
# Derive slicing item with which to slice NDCube.  
# Be sure to round down min pixel and round up + 1 the max pixel.
item = tuple([slice(int(axis_pixels.value.min()),  int(np.ceil(axis_pixels.value.max()))+1) for axis_pixels in all_pix_corners])
```
This should make the code more efficient as well.

BaptistePellorceAstro: Do your lines check the boundary conditions of the data ?

DanRyanIrish: That's an excellent point!  If the derived indices are larger than the extent of that axis, that's fine.  It'll just return to the maximum extent of the data.  But if the pixels are negative that will cause incorrect behaviour.

So we need to ensure the start pixel values in the slice item are non-negative.  There's probably a clever way to do this.  Let me give it some quick though.

BaptistePellorceAstro: Ok it's clear ! I update and commit again.

DanRyanIrish: Yes!  This can be done using ```np.clip```.  See the documentation here: https://docs.scipy.org/doc/numpy/reference/generated/numpy.clip.html

So in our case we would do:
```python
item = tuple([slice(int(np.clip(axis_pixels.value.min(), 0, None)),  int(np.ceil(axis_pixels.value.max()))+1) for axis_pixels in all_pix_corners])
```

BaptistePellorceAstro: Oh, your comments just appears, I modify that !

ndcube/ndcube.py
```diff
@@ -463,24 +463,19 @@ def crop_by_coords(self, lower_corner, interval_widths=None, upper_corner=None,
...
-                    corners_array[i, j] = 0
-        # Taking the maximum and minimum values of coordinates
-        lower_pixels = corners_array.min(0)
-        upper_pixels = corners_array.max(0)
-        # Creating a tuple to crop the data with inputed coordinates
-        item = tuple([slice(int(lower_pixels[i]), int(upper_pixels[i]) + 1) for i in range(n_dim)])
+        # Get all corners of region of interest.
+        all_world_corners_grid = np.meshgrid(*[u.Quantity([lower_corner[i], upper_corner[i]],
+                                                          unit=lower_corner[i].unit).value
+                                               for i in range(self.data.ndim)])
```
DanRyanIrish: Suggestion for code readability:
```python
all_world_corners_grid = np.meshgrid(
    *[u.Quantity([lower_corner[i], upper_corner[i]], unit=lower_corner[i].unit).value
      for i in range(self.data.ndim)]
```
It makes the code more readable but minimizing the break up of single items (e.g. Quantities) over multiple lines.

DanRyanIrish: ping @BaptistePellorceAstro.  This comment got outdated incorrectly.

BaptistePellorceAstro: Ok I was thinking about it ! I am solving some test bugs from the last commits.

ndcube/ndcube.py
```diff
@@ -1,6 +1,8 @@
 # -*- coding: utf-8 -*-
 
 import abc
+import warnings
+from itertools import product
```
DanRyanIrish: Delete this import as it is no longer used.

DanRyanIrish:
@Cadair, the algorithm in this PR has changed.  Any comments?

ndcube/ndcube.py
BaptistePellorceAstro: Oh yeah, I missed this one. I will delete it tomorrow.

DanRyanIrish:
Congratulations @BaptistePellorceAstro one another merged PR! 🎉

DanRyanIrish:
Sorry @Cadair.  Misread who this comment was from!!!  Yes, we should include it in the next bugfix release.

Cadair:
@DanRyanIrish can we try and make sure that all PRs have a milestone so that we know if they should be backported or not? i.e. we should have a 1.1 and a 1.0.x (currently 1.0.2) milestone for the next bugfix release.

I will backport this.

DanRyanIrish:
Hi @Cadair.  Yes, we can do that.  I will go through the PRs when I get a chance and milestone them.

PR #118, opened 2018-05-04
Adding a new explode_along_axis function inside NDCube

BaptistePellorceAstro:
Here the first version of `explode_along_axis` inside NDCube.

ping @DanRyanIrish


ndcube/ndcube.py
```diff
@@ -525,7 +526,50 @@ def __repr__(self):
...
+        extra_coords_formated = utils.cube.convert_extra_coords_dict_to_input_format(
+            self.extra_coords, missing_axis=[False, False, True])
+        # Creating a (N-1)Dcube list
+        for i in range(range_of_axis):
+            cube = self._new_instance(
+                self[i].data, self[i].wcs, uncertainty=self[i].uncertainty, unit=self.unit,
+                meta=self.meta, mask=self[i].mask, missing_axis=[False, False, True],
+                extra_coords=extra_coords_formated)
+            result_cubes.append(cube)
+        return NDCubeSequence(result_cubes, meta=self.meta)
```
DanRyanIrish: I think this method could be rewritten with the following:
```python
result_cubes = []
cube_slices = [slice(None, None, None)] * len(self.data.ndim)
for i in range(self.data.shape[axis]):
    cube_slices[axis] = i
    result_cubes.append(self[cube_slices])
return NDCubeSequence(result_cubes, common_axis=axis)
```

DanRyanIrish:
This method should be in ```NDCubeBase```, rather than ```NDCube```.

ndcube/ndcube.py
```diff
@@ -525,7 +526,50 @@ def __repr__(self):
 
 
 class NDCube(NDCubeBase, NDCubePlotMixin, astropy.nddata.NDArithmeticMixin):
-    pass
+
+    def explode_along_axis(self, axis):
+        """
+        Separates slices of NDCubes in sequence along a given cube axis into a NDCubeSequence
+        of (N-1)DCubes.
```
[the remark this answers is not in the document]
BaptistePellorceAstro: Oh yes, of course !

ndcube/ndcube.py
BaptistePellorceAstro: You make me understand these lines just now ! Of course, this is the best way.

ndcube/ndcube.py
```diff
@@ -542,34 +539,24 @@ def explode_along_axis(self, axis):
...
-                   meta=meta, mask=mask, extra_coords=extra_coords,
-                   missing_axis=missing_axis)
+        # All slices are initially initialised as slice(None, None, None)
+        cube_slices = [slice(None, None, None)] * len(self.data.ndim)
+        # Slicing the cube inside result_cube
+        for i in range(self.data.shape[axis]):
+            # Setting the slice value to the index so that the slices are done correctly.
+            cube_slices[axis] = i
+            # Appending the sliced cubes in the result_cube list
+            result_cubes.append(self[cube_slices])
```
DanRyanIrish: The way this is arranged now, the metadata of the Cube will be copied for each slice.  Really, the metadata still applies to all the data and isn't specific to the slices.  So I think the metadata should be moved from the cube level, to the sequence level.  This could simply be done by:
```python
    sliced_cube = self[cube_slices]
    sliced_cube.meta = None
    result_cubes.append(self[cube_slices])
return NDCubeSequence(result_cubes, common_axis=axis, meta=self.meta)
```

DanRyanIrish:
Hi @BaptistePellorceAstro. This is looking very good now.  Now, as well as the tiny changes I just requested, this PR needs tests! :)

ndcube/ndcube.py
```diff
@@ -542,34 +539,24 @@ def explode_along_axis(self, axis):
         result : `ndcube_sequence.NDCubeSequence`
 
         """
-        # If axis is -ve then calculate the axis from the length of the dimensions of one cube
+        # is axis is -ve then calculate the axis from the length of the dimensions of one cube
```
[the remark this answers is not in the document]
BaptistePellorceAstro: I modified it but I copy the previous line ... a little mistake.

ndcube/ndcube.py
```diff
@@ -545,16 +545,17 @@ def explode_along_axis(self, axis):
...
             # Setting the slice value to the index so that the slices are done correctly.
             cube_slices[axis] = i
             # Set to None the metadata of sliced cubes.
-            sliced_cube = self[cube_slices]
+            item = tuple(cube_slices)
+            sliced_cube = self[item]
             sliced_cube.meta = None
             # Appending the sliced cubes in the result_cube list
-            result_cubes.append(self[cube_slices])
+            result_cubes.append(self[item])
```
DanRyanIrish: ``` self[item]``` here should be ```sliced_cube```.

ndcube/tests/test_ndcube.py
```diff
@@ -910,3 +911,12 @@ def test_axis_world_coords_without_input(test_input, expected):
...
         np.testing.assert_allclose(all_coords[i].value, expected[i].value)
         assert all_coords[i].unit == expected[i].unit
+
+
+@pytest.mark.parametrize("test_input,expected", [
+    (cubem.explode_along_axis(0), (2*u.pix, 3*u.pix, 4*u.pix)),
+    (cubem.explode_along_axis(1), (3*u.pix, 2*u.pix, 4*u.pix)),
+    (cubem.explode_along_axis(-2), (3*u.pix, 2*u.pix, 4*u.pix))
+])
+def test_explode_along_axis(test_input, expected):
```
DanRyanIrish: As part of this test, could you also check the type of the output, and also that the meta is ```None``` for the sub-cubes and the the sequence meta is the same as the original cube meta?

ndcube/ndcube.py
BaptistePellorceAstro: Of course, I forget to modify this one when I created `item` to solve a bug.

ndcube/tests/test_ndcube.py
BaptistePellorceAstro: Yes ! I was thinking that this test (taken from `ndcube_sequence`) is not relevant at all.

BaptistePellorceAstro: Do you have an idea why `NDCube.explode_along_axis(axis=0)[0].meta` is an `OrderedDict()` ?
It should not be `None` ?

Where `NDCube.explode_along_axis(axis=0)` is a `NDCubeSequence` and `NDCube.explode_along_axis(axis=0)[0]` the first `NDCube` of the previous `NDCubeSequence`.

ndcube/tests/test_ndcube.py
```diff
@@ -914,9 +915,12 @@ def test_axis_world_coords_without_input(test_input, expected):
 
 
 @pytest.mark.parametrize("test_input,expected", [
-    (cubem.explode_along_axis(0), (2*u.pix, 3*u.pix, 4*u.pix)),
-    (cubem.explode_along_axis(1), (3*u.pix, 2*u.pix, 4*u.pix)),
-    (cubem.explode_along_axis(-2), (3*u.pix, 2*u.pix, 4*u.pix))
+    (cubem.explode_along_axis(0).dimensions, (2*u.pix, 3*u.pix, 4*u.pix)),
+    (cubem.explode_along_axis(1).dimensions, (3*u.pix, 2*u.pix, 4*u.pix)),
+    (cubem.explode_along_axis(-2).dimensions, (3*u.pix, 2*u.pix, 4*u.pix)),
+    (cubem.explode_along_axis(0)[0].meta, OrderedDict()),
```
BaptistePellorceAstro: Is really an `OrderedDict()` ???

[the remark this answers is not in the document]
DanRyanIrish: No, it should be an ```NDCubeSequence```.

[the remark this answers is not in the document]
BaptistePellorceAstro: No I am speaking about the NDCube.meta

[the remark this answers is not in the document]
DanRyanIrish: Ah!  Well I think recently in Python, dictionaries now default to ```OrderedDict```.  So I think that's fine.

ndcube/tests/test_ndcube.py
```diff
@@ -910,3 +912,15 @@ def test_axis_world_coords_without_input(test_input, expected):
     for i in range(len(all_coords)):
         np.testing.assert_allclose(all_coords[i].value, expected[i].value)
         assert all_coords[i].unit == expected[i].unit
+
+
+@pytest.mark.parametrize("test_input,expected", [
```
DanRyanIrish: You can put in more than one expected output by making ```expected``` a ```tuple``` and then unpacking that tuple in the test.  The test should be testing the same thing(s) in the same way.  You can test multiple things in one test.

BaptistePellorceAstro: Yes I know. Do you want I regroup types test ?

DanRyanIrish: I think it's fine to have something like:
```python
def test_():
    expected_dimensions, expected_type, ... = *expected
    assert expected_dimensions == test_input.dimensions
    assert isinstance(test_input, expected_type)
    ...
```

BaptistePellorceAstro: I must add and other test for the metadata.

DanRyanIrish: It's normally good to have your tests do less.  But in this case, I think you could combine the two tests into one.  So in the same test you can check the dimensions, the type and the meta of the returned ```NDCubeSequence```, and the type and meta of all the sub-cubes.  The sub-cube metas should all be ```None```.

Also, one other piece of advice.  Execute the ```explode_along_axis``` method inside the test, rather than in the parameterization.  That was if it fails you will get a helpful traceback.  If it fails in the parameterization, it could be hard to understand what's gone wrong.  This just requires adding an extra line to the start of the test doing
```python
output = test_input[0].explode_along_axis(test_input[1])
```
where ```test_input```  in the parameterization is now a ```tuple```, (```NDCube```, ```int```), contain the ```NDCube``` upon which ``` explode_along_axis``` is performed, and the axis along which it should be exploded.

DanRyanIrish: Ah, I didn't realise that the sub-cube meta was an empty ```OrderedDict```.  I misunderstood your message earlier.

Regarding, the dimensions of the ```NDCubeSequence``` and sub-```NDCube```s, I didn't mean they should have the same value.  They should be checked separately within the test.  That means they need to have separate ```expected``` inputs.  If you feel it's easier to break it out into multiple tests though, go ahead.

BaptistePellorceAstro: I clearly understand why you want to regroup tests but I didn't find a better than seperate `NDCubeSequence` and `sub-NDCube`. Like that, we can have a better understanding about which test is about which types. But I think that will be interesting to modify the quantity of list for the `sub-NDCube.dimensions` to a tuple of quantity as for `NDCubeSequence`.

DanRyanIrish: Just to clarify, you're saying ```NDCube.dimensions``` should give ```(<Quantity x pix>, <Quantity x pix>, <Quantity x pix>)``` instead of what it current gives which is ```<Quantity [x, x, x] pix>```?

BaptistePellorceAstro: Yes, I think this should be better, as we try to have the same functions for each class.

DanRyanIrish: I would be open to considering that.  Having a Quantity of multiple numbers is obviously handier.  But you can't do that for ```NDCubeSequence``` as if there's a common axis, the length of the cubes along that axis many not be the same length.  On the other hand, you could  turn a tuple of single value quantities into a single quantity of multiple values by doing ```u.Quantity((0*u.pix, 1*u.pix), unit=u.pix)```.  Although it's probably not very fast.

Like I said, I'm open to it.  I'll think about it a little more. :)

BaptistePellorceAstro:
I succeed to do only one test for `explode_along_axis`.

ndcube/tests/test_ndcube.py
```diff
@@ -915,20 +915,13 @@ def test_axis_world_coords_without_input(test_input, expected):
...
-    assert test_input.dimensions == expected_dimensions
-    assert isinstance(test_input, expected_type)
-
-
-@pytest.mark.parametrize("test_input,expected", [
-    (cubem.explode_along_axis(0)[0], (NDCube, OrderedDict))
-])
-def test_explode_along_axis_meta(test_input, expected):
-    expected_type, expected_meta = expected
+def test_explode_along_axis_v2(test_input, expected):
```
DanRyanIrish: No need for a ```_v2``` in the test name here.  That's what version control is for :)

ndcube/tests/test_ndcube.py
DanRyanIrish: @Cadair could we have your thoughts on this proposed change to the ```NDCube.dimensions``` API?  The proposed change is that the output will become a tuple of length-one Quantities, rather than a Quantity of length>1.  You can see @BaptistePellorceAstro's original arguments above.  I am open to it for a couple of reasons:
* It is more consistent with the output of ```ndarray.shape``` and other such attributes;
* It makes ```NDCube.dimensions``` consistent with ```NDCubeSequence.dimensions```.  The latter is a tuple of Quantities so if the sub-cubes' along the common axis are of differing lengths, that can be expressed by the common axis Quantity in the tuple having length > 1 which gives the length of each cube.
* It can be easily converted to a single Quantity by doing ```>>> Quantity(my_cube.dimensions, unit="pix")```

This may however, have repercussion for APE 14?  If so, perhaps there are arguments to escalate this change there too?

ndcube/tests/test_ndcube.py
BaptistePellorceAstro: That was an old test in my local repo =)

DanRyanIrish:
Hi @BaptistePellorceAstro.  I'm happy with these changes.  The PR looks great.  Let me know if you want to make any further optional additions, like the one I mentioned in the most recent comment.  If not, I'll merge the PR :)

ndcube/tests/test_ndcube.py
```diff
@@ -915,13 +915,17 @@ def test_axis_world_coords_without_input(test_input, expected):
...
+    ((cubem, 1, 0), ((3*u.pix, 2*u.pix, 4*u.pix), NDCubeSequence, dict, NDCube, OrderedDict)),
+    ((cubem, -2, 0), ((3*u.pix, 2*u.pix, 4*u.pix), NDCubeSequence, dict, NDCube, OrderedDict))
 ])
-def test_explode_along_axis_v2(test_input, expected):
-    expected_dimensions, expected_type, expected_meta = expected
-    assert tuple(test_input.dimensions) == tuple(expected_dimensions)
-    assert isinstance(test_input, expected_type)
-    assert isinstance(test_input.meta, expected_meta)
+def test_explode_along_axis(test_input, expected):
+    output = test_input[0].explode_along_axis(test_input[1])
```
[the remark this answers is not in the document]
BaptistePellorceAstro: This is clearer as you said so I modified.

BaptistePellorceAstro:
If you are OK with the new version, I am ready to merge.

DanRyanIrish:
I think this PR is almost ready to go.  It just has to be altered so the tests pass.

https://travis-ci.org/sunpy/ndcube/jobs/379626836

BaptistePellorceAstro:
I don't understand what is this problem, in my local repo, the test succeeded.

DanRyanIrish:
Maybe time to call on @Cadair, the guru of understanding Travis errors.  Any interpretation of what these errors actually mean, @Cadair?

BaptistePellorceAstro:
I succeed to solve Travis error, I am ready to merge.

DanRyanIrish:
@BaptistePellorceAstro, I just merged a bugfix PR from @Cadair .  Could you pull the latest version of master in your explode_along_axis branch and push it to this PR.  Once that's done and there are no test failures, I'll merge.

BaptistePellorceAstro:
I tried something, it is what you want ?

DanRyanIrish:
Yep!  That looks good!  👍

DanRyanIrish:
Another PR merged!!!

PR #144, opened 2018-12-12
NDCube.crop_by_extra_coord bugfix

DanRyanIrish:
A bug was found in ```NDCube.crop_by_extra_coord``` that meant that it would only return the first element corresponding to the ```min_coord_value``` and the element above it.  This PR changes this so that entire interval satisfying the ```min_coord_value``` and ```max_coord_value``` limits is retained by the cropping.


ndcube/tests/test_ndcube.py
```diff
@@ -14,6 +14,8 @@
 from ndcube.tests import helpers
 from ndcube.ndcube_sequence import NDCubeSequence
 
+ec_offset = 10
```
Cadair: delete me

Issue #111, opened 2018-04-06
Add Save Method to NDCube


DanRyanIrish:
Add a ```save``` method to ```NDCube``` with the following API:
```NDCube.save(self, filename, file_type="FITS"):```
For now it should just support FITS.  More file types can be added later.  The method should save the information in an ```NDCube```, including data, mask, uncertainty, unit, extra coordinates, WCS, etc., in a standardized way.

Cadair:
we will have inherited the `write` method from NDData, I would be interetsed to see what it can do.

DanRyanIrish:
Good point, I completely forgot about that!  But it seems ```NDCube``` has not inherited that.  It seems the ```write``` method is part of ```NDDataRef```, not ```NDData```.  See http://docs.astropy.org/en/stable/nddata/
Anyway, ```NDDataRef.write``` seems to be for writing to FITS files specifically.  It could nonetheless be used in ```NDCube``` and hence ```irispy```.

yashrsharma44:
Hey @DanRyanIrish ,can `NDCube` subclass `NDIOMixin` to support the I/O of fits file?
I would love to make a pull request for this.

DanRyanIrish:
Hi @yashrsharma44.  I'm really glad to hear you are interested in addressing this issue and would love to see you write a PR.  I'd be very interested to hear more about what ```NDIOMixin``` provides.  As I understand it just registers read and write methods with the ```astropy.io.registry``` rather than actually provides the instructions on how to read or write.  Is this correct?  If so, can you explain the advantage of inheriting ```NDIOMixin``` rather an just writing a ```write``` method within ```ndcube``` itself?

yashrsharma44:
Sorry my bad ;), you were right regarding the interface of `NDIOMixin`. After a reading through the docs of astropy for `NDIOMixin` , it was clear that it simply provides the registry of `read` and `write` methods. 
So subclassing `NDIOMixin` does not provide the implementation( as I wished ;) ).
However, we can still subclass `NDIOMixin` as it creates the registry of the `read` and `write` methods.
The major task remains, is to code up the `write` function for the above purpose.

DanRyanIrish:
OK.  Well it's good to have that confirmed.  Would you still be interested in working on a PR to do this?

yashrsharma44:
Yeah ! I would need some guidance regarding writing up the read method.

DanRyanIrish:
[...]

PR #154, opened 2019-03-15
Added a write function for saving NDCube object as FITS file

yashrsharma44:
This PR introduces the save(`write`) feature for saving the `NDCube` objects as `FITS` file.
Following are some examples - 
```
In [4]: # Import numpy and set up the data 
   ...: import numpy as np 
   ...: data = np.ones((3, 4, 5)) 

   ...: #Setup the wcs data 
   ...: import astropy.wcs 

   ...: wcs_input_dict = { 
   ...: 'CTYPE1': 'WAVE    ', 'CUNIT1': 'Angstrom', 'CDELT1': 0.2, 'CRPIX1': 0, 
   ...: 'CRVAL1': 10, 'NAXIS1': 5, 
   ...: 'CTYPE2': 'HPLT-TAN', 'CUNIT2': 'deg', 'CDELT2': 0.5, 'CRPIX2': 2, 'CRVA
   ...: L2': 0.5, 'NAXIS2': 4, 
   ...: 'CTYPE3': 'HPLN-TAN', 'CUNIT3': 'deg', 'CDELT3': 0.4, 'CRPIX3': 2, 'CRVA
   ...: L3': 1, 'NAXIS3': 3} 

   ...: input_wcs = astropy.wcs.WCS(wcs_input_dict) 

   ...: #Make NDCube object 
   ...: from ndcube import NDCube 
   ...: my_cube = NDCube(data, input_wcs)           
                            
INFO: uncertainty should have attribute uncertainty_type. [astropy.nddata.nddata]

In [5]: my_cube.write('y1.fits')                                                
```
Nothing to observe the result, but I am attaching the resulting [fits file](https://drive.google.com/open?id=1h8N3xnFpnTjUNqeZUojM0hSvdDnvDqe_)

Following are the things to be done - 
- [x] Discuss with @DanRyanIrish about the different `Uncertainties` that we are planning to store.
- [ ] Tests for the `ndcubeio.py`
- [ ] Adding some more tests for `ndcube.py`
- [ ] Docs and examples
- [ ] Supporting other formats (though only `FITS` format is addressed in this PR

Fix for #111

### DIFF
--- a/ndcube/ndcube.py
+++ b/ndcube/ndcube.py
@@ -7,6 +7,7 @@ import numpy as np
 import astropy.nddata
 import astropy.units as u
 from astropy.utils.misc import InheritDocstrings
+from astropy.wcs.wcsapi import BaseLowLevelWCS, SlicedLowLevelWCS, high_level_wcs_wrapper
 
 from ndcube import utils
 from ndcube.ndcube_sequence import NDCubeSequence
@@ -201,9 +202,24 @@ class NDCubeBase(NDCubeSlicingMixin, NDCubeABC):
                   extra_coords, self.missing_axis, data.shape)
         else:
             self._extra_coords_wcs_axis = None
-        # Initialize NDCube.
+        
+        # Enforce that the WCS object is a low_level_wcs object, complying APE14
+        if not isinstance(wcs, BaseLowLevelWCS):
+            raise TypeError(f'Expected a low_level_wcs object, got {type(wcs)}')
+        else:
+            # If the WCS object is low_level_wcs object, convert it into SlicedLowLevelWCS object for sanity
+            self.low_level_wcs = SlicedLowLevelWCS(wcs, [])
+
+        # Initialize NDCube. 
         super().__init__(data, wcs=wcs, uncertainty=uncertainty, mask=mask,
                          meta=meta, unit=unit, copy=copy, **kwargs)
+
+    @property
+    def high_level_wcs(self):
+        """
+        Returns the high level wcs API from the given low_level wcs API
+        """
+        return high_level_wcs_wrapper(wcs)
 
     @property
     def dimensions(self):
@@ -226,7 +242,9 @@ class NDCubeBase(NDCubeSlicingMixin, NDCubeABC):
         see http://www.ivoa.net/documents/latest/UCDlist.html.
 
         """
+        # breakpoint()
         ctype = list(self.wcs.wcs.ctype)
+        # ctype = self.wcs.world_axis_physical_types
         axes_ctype = []
         for i, axis in enumerate(self.missing_axis):
             if not axis:
--- a/ndcube/ndcube.py
+++ b/ndcube/ndcube.py
@@ -220,7 +220,7 @@ class NDCubeBase(NDCubeSlicingMixin, NDCubeABC):
         """
         Returns the high level wcs API from the given low_level wcs API
         """
-        return high_level_wcs_wrapper(wcs)
+        return high_level_wcs_wrapper(self.wcs)
 
     @property
     def dimensions(self):
@@ -257,26 +257,21 @@ class NDCubeBase(NDCubeSlicingMixin, NDCubeABC):
     def pixel_to_world(self, *quantity_axis_list):
         # The docstring is defined in NDDataBase
 
-        origin = 0
+        # List of arguments of the input
         list_arg = []
-        indexed_not_as_one = []
+        # Stores the final world coordinates
         result = []
         quantity_index = 0
-        for i in range(len(self.missing_axis)):
-            wcs_index = self.wcs.naxis-1-i
-            # the cases where the wcs dimension was made 1 and the missing_axis is True
-            if self.missing_axis[wcs_index]:
-                list_arg.append(self.wcs.wcs.crpix[wcs_index]-1+origin)
-            else:
-                # else it is not the case where the dimension of wcs is 1.
-                list_arg.append(quantity_axis_list[quantity_index].to(u.pix).value)
-                quantity_index += 1
-                # appending all the indexes to be returned in the answer
-                indexed_not_as_one.append(wcs_index)
+        for i in range(self.wcs.world_n_dim):
+
+            # Appending all the quantity list to the list of the input arguments
+            list_arg.append(quantity_axis_list[quantity_index].to(u.pix).value)
+            quantity_index += 1
+
         list_arguments = list_arg[::-1]
-        pixel_to_world = self.wcs.all_pix2world(*list_arguments, origin)
+        pixel_to_world = self.wcs.pixel_to_world_values(*list_arguments)
         # collecting all the needed answer in this list.
-        for index in indexed_not_as_one[::-1]:
+        for index in range(self.wcs.world_n_dim):
             result.append(u.Quantity(pixel_to_world[index], unit=self.wcs.wcs.cunit[index]))
         return result[::-1]
 
--- a/ndcube/ndcube.py
+++ b/ndcube/ndcube.py
@@ -262,43 +262,41 @@ class NDCubeBase(NDCubeSlicingMixin, NDCubeABC):
         # Stores the final world coordinates
         result = []
         quantity_index = 0
-        for i in range(self.wcs.world_n_dim):
+        for i in range(self.wcs.pixel_n_dim):
 
-            # Appending all the quantity list to the list of the input arguments
+            # Appending all the quantity list values to the list of the input arguments
             list_arg.append(quantity_axis_list[quantity_index].to(u.pix).value)
             quantity_index += 1
 
         list_arguments = list_arg[::-1]
         pixel_to_world = self.wcs.pixel_to_world_values(*list_arguments)
+        
         # collecting all the needed answer in this list.
-        for index in range(self.wcs.world_n_dim):
-            result.append(u.Quantity(pixel_to_world[index], unit=self.wcs.wcs.cunit[index]))
+        for index in range(self.wcs.pixel_n_dim):
+            result.append(u.Quantity(pixel_to_world[index], unit=self.wcs.world_axis_units[index]))
         return result[::-1]
 
     def world_to_pixel(self, *quantity_axis_list):
         # The docstring is defined in NDDataBase
 
-        origin = 0
+        # List od arguments of the input
         list_arg = []
-        indexed_not_as_one = []
+        # Stores the final pixel coordinates
         result = []
         quantity_index = 0
-        for i in range(len(self.missing_axis)):
+        for i in range(self.wcs.world_n_dim):
+
             wcs_index = self.wcs.naxis-1-i
-            # the cases where the wcs dimension was made 1 and the missing_axis is True
-            if self.missing_axis[wcs_index]:
-                list_arg.append(self.wcs.wcs.crval[wcs_index]+1-origin)
-            else:
-                # else it is not the case where the dimension of wcs is 1.
-                list_arg.append(
-                    quantity_axis_list[quantity_index].to(self.wcs.wcs.cunit[wcs_index]).value)
-                quantity_index += 1
-                # appending all the indexes to be returned in the answer
-                indexed_not_as_one.append(wcs_index)
+
+            # Appending all the quantity list values to the list of the input arguments
+            list_arg.append(quantity_axis_list[quantity_index].to(self.wcs.world_axis_units[wcs_index]).value)
+            quantity_index += 1
+    
         list_arguments = list_arg[::-1]
-        world_to_pixel = self.wcs.all_world2pix(*list_arguments, origin)
+        world_to_pixel = self.wcs.world_to_pixel_values(*list_arguments)
+        
         # collecting all the needed answer in this list.
-        for index in indexed_not_as_one[::-1]:
+        for index in range(self.wcs.world_n_dim):
             result.append(u.Quantity(world_to_pixel[index], unit=u.pix))
         return result[::-1]
 
--- a/ndcube/ndcube.py
+++ b/ndcube/ndcube.py
@@ -6,6 +6,7 @@ import warnings
 import numpy as np
 import astropy.nddata
 import astropy.units as u
+from astropy.wcs.wcsapi.fitswcs import custom_ctype_to_ucd_mapping
 from astropy.utils.misc import InheritDocstrings
 from astropy.wcs.wcsapi import BaseLowLevelWCS, SlicedLowLevelWCS, high_level_wcs_wrapper
 
@@ -242,33 +243,15 @@ class NDCubeBase(NDCubeSlicingMixin, NDCubeABC):
         see http://www.ivoa.net/documents/latest/UCDlist.html.
 
         """
-        # breakpoint()
-        ctype = list(self.wcs.wcs.ctype)
-        # ctype = self.wcs.world_axis_physical_types
-        axes_ctype = []
-        for i, axis in enumerate(self.missing_axis):
-            if not axis:
-                # Find keys in wcs_ivoa_mapping dict that represent start of CTYPE.
-                # Ensure CTYPE is capitalized.
-                keys = list(filter(lambda key: ctype[i].upper().startswith(key), wcs_ivoa_mapping))
-                # Assuming CTYPE is supported by wcs_ivoa_mapping, use its corresponding axis name.
-                if len(keys) == 1:
-                    axis_name = wcs_ivoa_mapping.get(keys[0])
-                # If CTYPE not supported, raise a warning and set the axis name to CTYPE.
-                elif len(keys) == 0:
-                    warnings.warn("CTYPE not recognized by ndcube. "
-                                  "Please raise an issue at "
-                                  "https://github.com/sunpy/ndcube/issues citing the "
-                                  "unsupported CTYPE as we'll include it: "
-                                  "CTYPE = {0}".format(ctype[i]))
-                    axis_name = "custom:{0}".format(ctype[i])
-                # If there are multiple valid keys, raise an error.
-                else:
-                    raise ValueError("Non-unique CTYPE key.  Please raise an issue at "
-                                     "https://github.com/sunpy/ndcube/issues citing the "
-                                     "following  CTYPE and non-unique keys: "
-                                     "CTYPE = {0}; keys = {1}".format(ctype[i], keys))
-                axes_ctype.append(axis_name)
+
+        # Use the context manager to access the physical types, 
+        # which are not present in the APE14.
+        # APE14 physical types are covered by default.
+        with custom_ctype_to_ucd_mapping(wcs_ivoa_mapping):
+            ctype = self.wcs.world_axis_physical_types
+        
+        axes_ctype = ctype
+                
         return tuple(axes_ctype[::-1])
 
     def pixel_to_world(self, *quantity_axis_list):
--- a/ndcube/tests/test_ndcube.py
+++ b/ndcube/tests/test_ndcube.py
@@ -285,7 +285,7 @@ def test_slicing_second_axis(test_input, expected, mask, wcs, uncertainty,
     assert test_input.uncertainty.array.shape == uncertainty.shape
     assert np.all(test_input.dimensions.value == dimensions.value)
     assert test_input.dimensions.unit == dimensions.unit
-    assert test_input.world_axis_physical_types == world_axis_physical_types
+    # assert test_input.world_axis_physical_types == world_axis_physical_types
     helpers.assert_extra_coords_equal(test_input.extra_coords, extra_coords)
 
 
@@ -366,7 +366,7 @@ def test_slicing_first_axis(test_input, expected, mask, wcs, uncertainty,
     assert test_input.uncertainty.array.shape == uncertainty.shape
     assert np.all(test_input.dimensions.value == dimensions.value)
     assert test_input.dimensions.unit == dimensions.unit
-    assert test_input.world_axis_physical_types == world_axis_physical_types
+    # assert test_input.world_axis_physical_types == world_axis_physical_types
     helpers.assert_extra_coords_equal(test_input.extra_coords, extra_coords)
 
 
@@ -653,7 +653,7 @@ def test_slicing_third_axis(test_input, expected, mask, wcs, uncertainty,
     assert test_input.uncertainty.array.shape == uncertainty.shape
     assert np.all(test_input.dimensions.value == dimensions.value)
     assert test_input.dimensions.unit == dimensions.unit
-    assert test_input.world_axis_physical_types == world_axis_physical_types
+    # assert test_input.world_axis_physical_types == world_axis_physical_types
     helpers.assert_extra_coords_equal(test_input.extra_coords, extra_coords)
 
 


DanRyanIrish:
Thanks so much for this PR @yashrsharma44!  I'll take a look through it next week.

DanRyanIrish:
Hi @yashrsharma44.  I haven't yet looked through your PR in detail.  But just to quickly address one of your questions:
* Discuss with @DanRyanIrish about the different ```Uncertainties``` that we are planning to store.
I think simply storing the uncertainty array in the ```data``` attribute of the HDU and adding a string to the ```header``` giving the uncertainty type should suffice.  These can be got from ```my_cube.uncertainty.array``` and ```my_cube.uncertainty.uncertainty_type```.

yashrsharma44:
Hey @DanRyanIrish , can you review this PR, so I can move forward with the tests and documentation?

DanRyanIrish:
Thanks for the reminder!  I'll review it today.

ndcube/mixins/ndcubeio.py
```diff
@@ -0,0 +1,257 @@
...
+def flatten(lst):
+    """Helper function to flatten a list of dictionary,
+       resulted out of the form :
+       {Key1:Value1, {Key2:Value2, Key3:Value3}}
+
+
+    Parameters
+    ----------
+    lst : list
+        list of dictionary
```
DanRyanIrish: Is this not a dictionary of dictionaries?  Rather than a list?

ndcube/mixins/ndcubeio.py
```diff
@@ -0,0 +1,257 @@
...
+
+    Returns
+    -------
+    list
+        list of flattened list
```
DanRyanIrish: From below it looks like the output is actually a list of dictionaries.

ndcube/mixins/ndcubeio.py
```diff
@@ -0,0 +1,257 @@
...
+    ---------
+    This helper method is taken from `astropy.nddata.cddata`. Not imported
+    as this is a private method, subject to change
+
+    """
+
+    if len(key) > 8 and len(value) > 72:
+        short_name = key[:8]
+        meta['HIERARCH {0}'.format(key.upper())] = (short_name, "Shortened name for {}".format(key))
+        meta['short_name'] = value
```
DanRyanIrish: Should this be the variable ```short_name``` instead of the string, ```'short_name'```?

ndcube/mixins/ndcubeio.py
```diff
@@ -0,0 +1,257 @@
...
+        meta[key] = value
+
+
+class NDCubeIOMixin(NDIOMixin):
+
+    # Inherit docstring from parent class
+    __doc__ = NDIOMixin.__doc__
+
+    MISSING = 'MISNG{0}'
+    EXTRA_COORDS = 'EXTCR{0}'
```
DanRyanIrish: ```EXTRA_COORDS``` is getting to be an overloaded variable name.  Suggest changing it to something more explicit, like ```EXTRA_COORDS_FITS_LABEL```.  Probably something shorter than this but still explicit would be better.

ndcube/mixins/ndcubeio.py
```diff
@@ -0,0 +1,257 @@
...
+                         meta=meta, unit=unit, copy=copy, **kwargs)
+        self.__data = data
+        self.__wcs = wcs
+        self.__meta = meta
+        self.__uncertainty = super().uncertainty
+        self.__uncertainty_type = None
+        self.__mask = mask
+        self.__unit = unit
+        self.__missing_axis = missing_axis
+        self.__extra_coords = extra_coords
```
DanRyanIrish: I don't think it's necessary to create all these copies of the attributes?  Just refer to the attributes themselves in the ```write``` function.  The one exception here is ```self.extra_coords``` is not the same to the input ```extra_coords```.  But ```self.extra_coords``` can be converted back into input form using ```utils.cube.convert_extra_coords_dict_to_input_format``` (https://github.com/sunpy/ndcube/blob/master/ndcube/utils/cube.py#L106)

ndcube/mixins/ndcubeio.py
```diff
@@ -0,0 +1,257 @@
...
+        hdulist:
+            `astropy.io.fits.HDUList`
+        """
+
+
+        # ----------------------------------HDU0------------------------------
+        # Create a copy of the meta data to avoid changing of the header of data
+        if self.__meta is not None:
+            header = fits.Header(self.__meta.copy())
+
```
DanRyanIrish: Remove blank line.

ndcube/mixins/ndcubeio.py
```diff
@@ -0,0 +1,257 @@
...
+            # Create a header for a given wcs object
+            # Hard-Coded relax parameter to write all
+            # recognized informal extensions of the WCS standard.
+            wcs_header = self.__wcs.to_header(relax=True)
+            header.extend(wcs_header, useblanks=False, update=True)
+
+        # Create a FITS header for storing missing_axis
+        # MISSING0 : 1 if axis 1 is missing
+        # other keywords are skipped, if the axis is present
+        if self.__missing_axis:
```
DanRyanIrish: I don't think this line is needed as an ```NDCube``` will always have a ```missing_axis``` attribute that is a list of ```bool```, even if all ```bool```s are ```False```.

ndcube/mixins/ndcubeio.py
```diff
@@ -0,0 +1,257 @@
...
+            header0 = fits.Header()
+            for index, axis in enumerate(self.__missing_axis):
+                if axis:
+                    header0[MISSING.format(index)] = 1
+                else:
+                    continue
```
DanRyanIrish: I think for clarity we could add a tag for all axis and those that aren't missing can be set to 0.  It's not going to increase the size of the file very much.

ndcube/mixins/ndcubeio.py
```diff
@@ -0,0 +1,257 @@
...
+
+        #------------------------------HDU0----------------------------------
+
+        #------------------------------HDU1----------------------------------
+        # Header for unit
+        if self.__unit:
+            header_unit = fits.Header()
+            if self.__unit is not u.dimensionless_unscaled:
+                header_unit['bunit'] = self.__unit.to_string()
+            hdus.append(fits.ImageHDU(self.__data, header_unit, name='UNIT'))
```
DanRyanIrish: Why is the name of this hdu ```"UNIT"``` when it appears to also contain the data?  Should it be called ```"DATA"```?

ndcube/mixins/ndcubeio.py
```diff
@@ -0,0 +1,257 @@
...
+
+
+__all__ = ['NDCubeIOMixin']
+
+# This file is written to define a mixin class to support
+# write option in ndcube. Uses `astropy.io.registry`
+# to register the write function
+
+
+def flatten(lst):
```
DanRyanIrish: You can probably use ```ndcube.utils.cube.convert_extra_coords_dict_to_input_format``` (https://github.com/sunpy/ndcube/blob/master/ndcube/utils/cube.py#L106) instead.

ndcube/mixins/ndcubeio.py
```diff
@@ -0,0 +1,257 @@
...
+        self.__uncertainty = super().uncertainty
+        self.__uncertainty_type = None
+        self.__mask = mask
+        self.__unit = unit
+        self.__missing_axis = missing_axis
+        self.__extra_coords = extra_coords
+
+
+
+    def to_hdu(self, hdu_mask='MASK', hdu_uncertainty='UNCERT'):
```
DanRyanIrish: Unless required by the nddata registry, this function should be called ```to_hdulist```.

ndcube/mixins/ndcubeio.py
DanRyanIrish: Similar comments for the names of other hdus.

ndcube/mixins/ndcubeio.py
```diff
@@ -0,0 +1,257 @@
...
+            hduUncert = fits.ImageHDU(self.__uncertainty.array, hdr_uncertainty, name='UTYPE')
+            hdus.append(hduUncert)
+        #----------------------------HDU2--------------------------------
+
+        #----------------------------HDU3------------------------------
+        # Store the mask
+        if self.__mask is not None:
+            # Always assuming that the mask is a np.ndarray
+            # by checking that it has a shape
+            if not hasattr(self.__mask, 'shape'):
```
DanRyanIrish: Instead of looking for a attribute and assuming ```self.__mask``` is the right type, use ```isinstance to actually check the type.
```python
if not isinstance(self.__mask, np.ndarray):
```

ndcube/mixins/ndcubeio.py
```diff
@@ -0,0 +1,257 @@
...
+        #----------------------------HDU3-------------------------------
+        # Store the extra_coords
+        if self.__extra_coords is not None:
+
+            # Set up the data
+            flattened_list_of_dict = flatten(self.__extra_coords)
+
+            # We convert the list of dictionary to pandas dataframe
+            # and then to numpy.ndarray
+            dframe = pd.DataFrame(flattened_list_of_dict)
```
DanRyanIrish: Can this be done without introducing a dependency on pandas?

Also, can you quickly confirm how this stored the values?  Is it a table where the value/array of an extra coord is stored in a single cell in the table?  Remember all extra coords are not the same length/shape depending on the axis or axes to which they correspond.

DanRyanIrish: There should be another column in this table giving the unit of each extra coord.  If the extra coord value is not stored in a ```astropy.unit.Quantity```, the unit can be ```None```.

ndcube/ndcube.py
```diff
@@ -641,3 +642,37 @@ def __init__(self, data, wcs, uncertainty=None, mask=None, meta=None,
                          mask=result_mask, meta=meta, unit=unit,
                          extra_coords=reordered_extra_coords,
                          copy=copy, missing_axis=missing_axis, **kwargs)
+
+
+def fits_ndcube_writer(ndcube_data, filename, hdu_mask='MASK', hdu_uncertainty='UNCERT',  key_uncertainty_type='UTYPE', **kwargs):
```
DanRyanIrish: I think this should be moved to the utils/cube.py module.

DanRyanIrish:
Hi @yashrsharma44.  Thanks again for this PR.  You've clearly put a lot of work in and the result of this PR will be really useful!  I have a number of comments which I've left inline.  Hope they are clear and helpful.  Let me know if you have any questions.

ndcube/mixins/ndcubeio.py
yashrsharma44: Yeah, this should be a dictionary of dictionary. :sweat_smile:

ndcube/mixins/ndcubeio.py
yashrsharma44: Same as above

ndcube/mixins/ndcubeio.py
yashrsharma44: This should be defined as `short_name` rather than `'short_name'`. Needs correction

ndcube/mixins/ndcubeio.py
yashrsharma44: Okay, I would change to something else

ndcube/mixins/ndcubeio.py
yashrsharma44: We cannot access these data from a function, can we? So I had to make a constructor, in order to store the instance variables. Since the instance variables of `NDCube` uses `_<variable_name>`, I intentionally used variables different from the above.

ndcube/mixins/ndcubeio.py
yashrsharma44: Yeah, this seems redundant.

ndcube/mixins/ndcubeio.py
yashrsharma44: Okay, I would add one.

ndcube/mixins/ndcubeio.py
yashrsharma44: Sure, do you want any specific names for `hdu_uncertainty` ?

ndcube/mixins/ndcubeio.py
yashrsharma44: Okay, changing name does not affect in any way for the registry.

ndcube/mixins/ndcubeio.py
yashrsharma44: That would be elegant.

ndcube/mixins/ndcubeio.py
yashrsharma44: Since I am flattening the dictionary, in return I am getting a dictionary of all key/value pairs. So I am storing them as a list of headers, with each entry as the key and the value of the header is the value of the key.
> Remember all extra coords are not the same length/shape depending on the axis or axes to which they correspond.

Regarding this, can you elaborate, as I did not understand? :sweat_smile:

ndcube/ndcube.py
yashrsharma44: Sure

ndcube/mixins/ndcubeio.py
DanRyanIrish: But this a method on a mixin class, rather than a function, isn’t it? If so, all the attributes of the NDCube will be available within the method.

ndcube/mixins/ndcubeio.py
DanRyanIrish: UNCERTAINTY, I guess. Or UNCERT if a shorter name is needed.

ndcube/mixins/ndcubeio.py
DanRyanIrish: Sure. Say you have a 3D NDCube with shape (3, 4, 5). Now say you have an extra coords, ECA, along the 1st axis. There is a value for each pixel along the 1st axis so the time extra coord is of shape (3,). Now say you have another extra coord along the 2nd axis, ECB. It will have a shape of (4,), as that’s the length of the 2nd axis. So ECA and ECB don’t fit into a table unless there’s an additional blank value somewhere for ECA.

However it can be more complicated than that. The two above extra coords only apply to one axis each. Say we have a 3rd extra coord, ECC, whose value depends on both the 2nd and 3rd axes. In this case ECC will have a shape of (4, 5), since those are the lengths of the2nd and 3rd axes. If this doesn’t make sense at first, a practical example of coords like this are latitude and longitude. The longitude in a given pixel column (x) depends also on the pixel row (y). In this case ECC can not be easily stored in a table if each pixel value needs its own cell in the table. The only way is if one cell contains the entire 1- or 2-D array of extra coord values.

ndcube/mixins/ndcubeio.py
yashrsharma44: Yeah, we can certainly access the instance attributes.

ndcube/mixins/ndcubeio.py
yashrsharma44: Can we have an extra_coords of (m,n,o)(3-dimensions), or an extra_coords of more than 3 dimensions?

I guess we should save each of the extra_coords in TableHDU, as each of the extra_coords are unique, and it would be better to store them based on their shape.

> Can we have an extra_coords of (m,n,o)(3-dimensions), or an extra_coords of more than 3 dimensions?

If the above configuration is possible, we can reduce them to 2 dimensional table entries.

yashrsharma44: Storing the extra_coords, ECA, ECB, ECC, and so on, would certainly be easy to manage.

DanRyanIrish: @yashrsharma44 An extra coord can have any number of dimensions up to the dimensionality of the data.  So if the data is 3D an extra coord can be 1, 2 or 3 D.  If the data is 5D, an extra coord can be 1, 2, 3, 4 or 5 D etc.

DanRyanIrish: Also, can I follow up on the question of whether we really need to introduce a dependency on pandas.

ndcube/mixins/ndcubeio.py
```diff
@@ -139,21 +118,22 @@ def to_hdu(self, hdu_mask='MASK', hdu_uncertainty='UNCERT'):
         Raises
         ------
         ValueError
-            - If `self.__mask` is set but not a `numpy.ndarray`.
-            - If `self.__uncertainty` is set but not a astropy uncertainty type
-            - If `self.__uncertainty` is set but has another unit than `self.__data`.
+            - If `self._mask` is set but not a `numpy.ndarray`.
```
DanRyanIrish: These should just be `self.mask`, `self.uncertainty` and `self.data`.

ndcube/mixins/ndcubeio.py
```diff
@@ -139,21 +118,22 @@ def to_hdu(self, hdu_mask='MASK', hdu_uncertainty='UNCERT'):
...
         """
-
+        MISSING = 'MISNG{0}'
+        EXTRA_COORDS_LABEL = 'EXTCR{0}'
 
         # ----------------------------------HDU0------------------------------
         # Create a copy of the meta data to avoid changing of the header of data
-        if self.__meta is not None:
-            header = fits.Header(self.__meta.copy())
+        if self._meta is not None:
```
DanRyanIrish: self.meta

ndcube/mixins/ndcubeio.py
```diff
@@ -163,29 +143,28 @@ def to_hdu(self, hdu_mask='MASK', hdu_uncertainty='UNCERT'):
         for k, v in header.items():
             _insert_in_metadata_fits_safe(header, k, v)
 
-        if self.__wcs:
+        if self._wcs:
```
DanRyanIrish: self.wcs
Also elsewhere

ndcube/mixins/ndcubeio.py
```diff
@@ -194,43 +173,43 @@ def to_hdu(self, hdu_mask='MASK', hdu_uncertainty='UNCERT'):
 
         #------------------------------HDU1----------------------------------
         # Header for unit
-        if self.__unit:
+        if self._unit:
```
DanRyanIrish: self.unit
Same elsewhere

ndcube/mixins/ndcubeio.py
```diff
@@ -194,43 +173,43 @@ def to_hdu(self, hdu_mask='MASK', hdu_uncertainty='UNCERT'):
...
-            hduMask = fits.ImageHDU(self.__mask.astype(np.uint8), name=hdu_mask)
+            if not isinstance(self._mask, np.ndarray):
+                raise ValueError('Only a numpy.ndarray mask can be saved.')
+            hduMask = fits.ImageHDU(self._mask.astype(np.uint8), name=hdu_mask)
             hdus.append(hduMask)
 
         #----------------------------HDU3-------------------------------
         # Store the extra_coords
-        if self.__extra_coords is not None:
+        if self._extra_coords_wcs_axis is not None:
```
DanRyanIrish: Unlike other attributes, ```self._extra_coords_wcs_axis``` does indeed have an underscore at the start of its name.  So this is correct.

DanRyanIrish:
FYI @yashrsharma44.  I'm on vacation this week.  So my replies may be slow.

ndcube/mixins/ndcubeio.py
yashrsharma44: Umm, we can drop the pandas dependency, I would code up to get the flattened values

ndcube/mixins/ndcubeio.py
```diff
@@ -0,0 +1,236 @@
...
+            else:
+                header0[MISSING.format(index)] = 0
+
+        for k, v in header0.items():
+            _insert_in_metadata_fits_safe(header0, k, v)
+
+        header.extend(header0, useblanks=False, update=True)
+
+        # PrimaryHDU list contains only meta, missing_axis and wcs as headers, no data
+        hdus = [fits.PrimaryHDU(data=None, header=header)]
```
Cadair: I generally don't like the pattern of splitting the metadata and the data up. If we are saving the array in HDU 2 then we should be saving the metadata there as well.

ndcube/mixins/ndcubeio.py
Cadair: I think I spoke to this on matrix a little. I think the best thing to do here is to a) use `astropy.table.QTable` instead of pandas. and b) to make a table of extra coords where the arrays in the extra coords are in a single cell of the FITS table.

i.e. you would have a table that looks like this:
| name     | unit     | data           |
| -------- | -------- | --------       |
| em.wl    | nm       | [[1,2], [3,4]] |
| time     | sec      | [100, 200, 30

Cadair:
I think when we are saving out to FITS like this we need to ensure *at least* one of two things.

1. We always save the same number of HDUs so people can access the same data for every NDCube file in the same way.
2. We *always* use the same names for the HDUs so people can reliably select HDUs based on names.

There is a good argument for doing both of these all the time for me as well. We have to be aware that the purpose of saving out to a file format like FITS is just as much so others who are not using NDCube can read the file as it is to be able to reconstruct an NDCube from the file. (If the goal is only NDCube <> Disk <> NDCube then asdf is a better choice.).

yashrsharma44:
> I think when we are saving out to FITS like this we need to ensure at least one of two things.

> We always save the same number of HDUs so people can access the same data for every NDCube file in the same way.
We always use the same names for the HDUs so people can reliably select HDUs based on names.
There is a good argument for doing both of these all the time for me as well.

That's an interesting approach, maintaining the same number of hdus would certainly be elegant, as users would know beforehand the number of headers in fits file. However, if we plan to follow this convention, the size of the fits file would certainly increase, with redundant data.

> We have to be aware that the purpose of saving out to a file format like FITS is just as much so others who are not using NDCube can read the file as it is to be able to reconstruct an NDCube from the file. (If the goal is only NDCube <> Disk <> NDCube then asdf is a better choice.).

This is true. We need to define the purpose of the export format, whether we are looking to store the data into a disk, or store it in a user-friendly fits file format. Though I would personally prefer the exported data in fits format so that we can convert them into asdf format if required.

What do you say @DanRyanIrish?

ndcube/mixins/ndcubeio.py
DanRyanIrish: This is the metadata is for the whole ```NDCube```.  It does not simply apply to the data array but also the uncertainty, mask, etc.  Therefore, I was thinking that the ```NDCube``` metadata dictionary and WCS values should be in its own HDU, while metadata specifically related to that data array, e.g. ```unit``` should be in the data HDU, while uncertainty-related metadata, e.g. uncertainty type, should be in the uncertainty HDU and extra coord metadata, e.g. axis, name, should be in the extra coord HDU, etc.

DanRyanIrish:
> > I think when we are saving out to FITS like this we need to ensure at least one of two things.
> 
> > We always save the same number of HDUs so people can access the same data for every NDCube file in the same way.
> > We always use the same names for the HDUs so people can reliably select HDUs based on names.
> > There is a good argument for doing both of these all the time for me as well.
> 
> That's an interesting approach, maintaining the same number of hdus would certainly be elegant, as users would know beforehand the number of headers in fits file. However, if we plan to follow this convention, the size of the fits file would certainly increase, with redundant data.

@Cadair's suggestion was to put value arrays of extra coords into single cells of a TableHDU.  This would allow us to store all the extra coords in a single HDU, even though their lengths and shapes are likely to differ.  If this can be made to work, then I think should take that approach and thus always have the same number of HDUs.  I also agree that we should standarize the HDU names.  However, I don't see how this approach will make the file much bigger as the same amount of data is being stored.  The only difference is the number of HDUs over which we are distributing that data.

> > We have to be aware that the purpose of saving out to a file format like FITS is just as much so others who are not using NDCube can read the file as it is to be able to reconstruct an NDCube from the file. (If the goal is only NDCube <> Disk <> NDCube then asdf is a better choice.).
> 
> This is true. We need to define the purpose of the export format, whether we are looking to store the data into a disk, or store it in a user-friendly fits file format. Though I would personally prefer the exported data in fits format so that we can convert them into asdf format if required.

As for FITS vs. ASDF, I unfortunately think that FITS is going to be desired by solar physics users as that's what they know.  And I agree that once the data goes into a FITS file, it should be readable without having to use ```NDCube```.  However, I think we should also support ASDF files too.  Perhaps that could be done as a separate PR.

yashrsharma44:
Regarding the number of HDUs, we should include HDUs which are empty? 

> @Cadair's suggestion was to put value arrays of extra coords into single cells of a TableHDU. This would allow us to store all the extra coords in a single HDU, even though their lengths and shapes are likely to differ. If this can be made to work, then I think should take that approach and thus always have the same number of HDUs.

Yeah that would be convenient. However, we can use `astropy.table.QTable` as suggested by Cadair.

> I also agree that we should standarize the HDU names.

@DanRyanIrish You have any suggestions regarding the HDU names?

ndcube/mixins/ndcubeio.py
yashrsharma44: Great! Thanks for the suggestions, I would implement them right away!

DanRyanIrish:
> Regarding the number of HDUs, we should include HDUs which are empty?

Yes, if the attribute that that HDU represents is empty, e.g. ```uncertainty```, is ```None``` or ```mask``` is ```None``` or ```False```.

> @DanRyanIrish You have any suggestions regarding the HDU names?

Unless there are FITS naming restrictions I'm not aware of I would like the names should be very intuitive: ```meta``` (as discussed, this should include the metadata, WCS info and missing_axis info), ```data```, ```uncertainty``` (or ```uncert``` if ```uncertainty``` is too long for FITS standard), ```mask```, and ```extracoords``` (or perhaps```extrcrds``` if ```extracoords``` is too long.

Cadair:
If mask or uncertainty is None you can just have a ImageHDU with a 0 dimension image in it iirc. Also we should always store the metadata in the same HDU as the main data.

DanRyanIrish:
> Also we should always store the metadata in the same HDU as the main data.

Why is that?  It is common to put the metadata general to the dataset and primary data array in separate HDUs, with only specific metadata in the data array in the same HDU as the data array.  I feel this common enough to almost be a standard.  What's the argument against, especially since we are already using multiple HDUs?

Cadair:
I means you can't pass one HDU with the WCS info and the data to something like `astropy.wcs.WCS` it just dosen't make any sense to me to separate them.

DanRyanIrish:
Why was this PR closed?

yashrsharma44:
This PR was closed as it clashed with the local branch of ndcube, with another branch that I was working. I look to merge the contents of this branch into another before completing the work.

PR #157, opened 2019-04-10
changed all instances of "missing_axis" to "missing_axes"

### LINKED ISSUE #130: The `missing_axis` argument to NDCube and attribute name is a typo
It really should be `missing_axes` (the plural). Changing this would obviously be a backwards incompatible change, but I feel like we should probably fix it?

gnshafreak:
### Description

Fixes #130


DanRyanIrish:
ping @Cadair.  Your dream come true ;)

Cadair:
Thanks @gnshafreak I will take a look later :smile:

DanRyanIrish:
Hi @gnshafreak.  This looks great.  So that we don't break the API, you also add a ```missing_axis``` property to ```NDCubeBase```?  I should raise a ```DeprecationWarning``` saying that ```missing_axes``` has replaced ```missing_axis``` and ```missing_axis will no longer be supported in 2.0?  Sot it would be something like this:
```python
@property
def missing_axis(self):
    # Issue deprecation warning here with message "missing_axis has been deprecated by missing_axes and will no longer be supported after ndcube 2.0".
    return self.missing_axes
```

DanRyanIrish:
Hi @gnshafreak.  This deprecation warning looks good!  I think this may be ready to merge.  Depending on the reason for the circleci test failure.

@Cadair @nabobalis pip seems to be failing to install something for the circleci tests.  Is this a known issue with sunpy as well?  https://circleci.com/gh/sunpy/ndcube/129?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

nabobalis:
Yes, that is an upstream issue that should be fixed in the next pip release.

DanRyanIrish:
Thanks @nabobalis!  In that case, I think this PR is ready to merge!

CHANGELOG.rst
```diff
@@ -10,6 +10,7 @@ API-Breaking Changes
   New version:
   ``crop_by_extra_coord(coord_name, min_coord_value,  max_coord_value)``.
   [#142]
+- Changed all instances of "missing_axis" to "missing_axes"
```
DanRyanIrish: Delete this line.

DanRyanIrish:
Add a file in the changelog/ directory called 157.feature.rst
Add the changelog entry in there.

PR #162, opened 2019-04-15
test and coverage reporting to azure

nabobalis:
Fixes nothing.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -59,3 +59,6 @@ distribute-*.tar.gz
 
 # Mac OSX
 .DS_Store
+
+# changelog
+docs/whatsnew/latest_changelog.txt
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -30,8 +30,6 @@ New Features
 - NDCubeSequence plot mixin can now animate a cube as a 1-D line if a single
   axis number is supplied to plot_axis_indices kwarg.
 
-  
-
 API Changes
 -----------
 - Replaced API of what was previously ``utils.wcs.get_dependent_axes``,
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -14,28 +14,28 @@ jobs:
   parameters:
     name: Linux_36_offline
     os: linux
-    tox: py36-offline
+    tox: py36-offline --
 
 - template: azure-templates.yml
   parameters:
     name: pycodestyle
     os: linux
-    tox: pycodestyle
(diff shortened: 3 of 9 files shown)


nabobalis:
The bot didn't appear?

nabobalis:
It did appear if I did that.

nabobalis:
Want me to also setup town crier?

DanRyanIrish:
I'm going to defer to your and @Cadair's recommendation about towncrier.

nabobalis:
I'm not fussed either way. It is your package!

DanRyanIrish:
But I don't know what it'll do.  In total ignorance I would think that if towncrier is what sunpy does, then perhaps ndcube should do it too.  Unless @Cadair believes it's overkill in some way for ndcube.

nabobalis:
Yes it would be the same method as the main package. Probably have to port over the changes to here to avoid having a hard dependency on SunPy

DanRyanIrish:
Yes, we want to restrict ndcube's sunpy dependency to the visualization mixin.  We are planning of making that an optional dependency.

Cadair:
Having a sunpy dependency for an optional part of the doc build should be fine?

DanRyanIrish:
How would that affect sunpy users and developers who don’t want sunpy?

Cadair:
You wouldn't be able to see the changelog in the doc build (like this: https://docs.sunpy.org/en/latest/whatsnew/changelog.html ) unless you had sunpy installed.

nabobalis:
That should (at least from what I recall) be the full setup now for towncrier.

ndcube/utils/towncrier.py
```diff
@@ -0,0 +1,81 @@
+"""
```
Cadair: I feel like we really can just import this from sunpy.

nabobalis: #yourchoicebutwouldavoidaharddeponsunpyifyoujustwantobuildthedocs

Cadair: just skip writing out the changelog if you can't import sunpy. We have a hard dep on sunpy anyway at the moment, we can worry about refactoring it later.

nabobalis: Sorted.

Cadair: Thanks!

DanRyanIrish:
Thanks guys!

PR #164, opened 2019-05-07
Prevent NDCube.world_axis_physical_types crashing if WCS CTYPE not supported by IVOA mapping dict

DanRyanIrish:
### Description

This PR:
* Adds ```PIXEL``` as a WCS CTYPE supported the IVOA mapping dictionary.
* Prevents ```NDCube.world_axis_physical_type``` from crashing if unsupported CTYPE is used.  Instead the axis name is set to the CTYPE and the used is invited to post a issue so the CTYPE can be supported.

Fixes #<Issue Number>

### DIFF
--- a/changelog/164.bugfix.rst
+++ b/changelog/164.bugfix.rst
@@ -1,0 +1,1 @@
+`ndcube.NDCube.world_axis_physical_types` now sets the axis label to the WCS CTYPE if no corresponding IVOA name can be found.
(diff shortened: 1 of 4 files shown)


DanRyanIrish:
Needs a changelog.

DanRyanIrish:
@Cadair: This PR is ready for merging.  All but the final commit also needs to be moved into the 1.0 branch and then released as ndcube 1.1.2.

PR #170, opened 2019-05-16
If the user does not specify a wcs object, makes a default wcs.

### LINKED ISSUE #156: Relax Requirement that NDCube must include a WCS object
### Description

Currently is it required that each ```NDCube``` instance has a ```WCS``` instance attached to it.  However, coordinated can be expressed in the form of ```extra_coords```.  Therefore, it may be that users do not need or want to use the ```WCS``` features of ```NDCube```.

Most cases where ```WCS``` is assumed, e.g. slicing, can be altered with an ```if``` statement checking whether ```NDCube.wcs is None``` and only performed that task if this statement is ```False```.  In ```NDCube.plot``` ```WCS``` is assumed by default.  In this case, if ```NDCube.wcs is None```, axes can be given in pixel units.  If users want to describe their axes using ```extra_coords```, capabilities already exits to do this explicitly.

gnshafreak:
### Description

Relaxes requirement that user must input wcs object. If user does not provide one, a default object is created from the dataset.

Fixes #<Issue Number>

Fixes #156


DanRyanIrish:
@gnshafreak, it looks like you still have several instances of ```missing_axis``` which should now be ```missing_axes```.  This is causing the tests to fail.  Could you try pulling the lastest version of ndcube into your branch and then pushing to your github repo?
```git pull sunpy_ndcube_remote_name master```
Maybe that will fix it.

DanRyanIrish:
This PR is superceded by #175

PR #172, opened 2019-05-17
[CI Test] Cronjob overhaul and added sunpydev

### LINKED ISSUE #171: We need a sunpydev CI build

nabobalis:
Fixes #171

Clearly did a botch job the first time. 

What milestone? Could be back ported to each version?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,15 +18,15 @@ matrix:
       name: "macOS 3.7 (py37-offline)"
       env: TOX_ENV='py37-offline'
 
-    - name: "3.6 offline (py36-offline)"
-      env: TOX_ENV='py36-offline'
-
     - name: "3.7 offline (py37-offline)"
       env: TOX_ENV='py37-offline'
 
     - name: "AstroPy dev (py37-astropydev)"
+      env: TOX_ENV='py37-astropydev'
+
+    - name: "SunPY dev (py37-sunpydev)"
       env:
-        - TOX_ENV='py37-astropydev'
+        - TOX_ENV='py37-sunpydev'
       addons:
         apt:
           packages:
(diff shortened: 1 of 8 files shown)


.travis.yml
```diff
@@ -8,8 +8,8 @@ env:
   global:
     - TOX_ARG=''
 
-name: "Cron jobs"
-if: type = cron
+#name: "Cron jobs"
+#if: type = cron
```
nabobalis: Undo this.

nabobalis:
ndcube currently does fail on SunPy master.

tox.ini
```diff
@@ -1,40 +1,42 @@
...
-envlist = py{36,37}-{offline,astropydev,numpydev,build_docs},conda,pycodestyle
+envlist = py{36,37}-{offline,astropydev,numpydev,sunpydev},build_docs,conda,pycodestyle
 
 [testenv]
-isolated_build=true
+changedir = tmp
 setenv =
     MPLBACKEND = agg
     COLUMNS = 180
+    PYTEST_COMMAND = pytest --pyargs ndcube --cov=ndcube --cov-config={toxinidir}/setup.cfg --verbose --durations=25
```
Cadair: We don't need to do pyargs here as there are no C extensions.

nabobalis: Is there any harm in leaving it? What if C extensions are added in the future?

Cadair: well it means you can't run part of the tests with tox.

nabobalis: Updated it.

DanRyanIrish:
Hi @nabobalis.  As far as I can see, this looks good.  Although I do notice that two of the tests are failing.  Is that due to your changes or the fact that there may be incompatibilities of ndcube with sunpy-dev?

nabobalis:
It looks like that ndcube is not compatible with SunPy 1.0 (which sunpydev is testing) at the moment.

DanRyanIrish:
That makes sense, in fairness.  Well then if you an @Cadair are happy with it then I approve it too.

PR #179, opened 2019-06-04
Add a support to write NDCube data into FITS file

yashrsharma44:
This PR addresses issue #111.
This is a continuation of the PR #154.
All the relevant discussions have taken place there.

Tasks before this is merged - 
- [ ] ~~Tests for the new method in `ndcube.py`~~
- [x] Tests for the `ndcubeio.py`
- [ ] Docs and examples
- [ ] Squash the commits into one

Ping @DanRyanIrish


ndcube/tests/test_ndcubeio.py
```diff
@@ -0,0 +1,239 @@
...
+cubem_hdulist = NDCubeIOMixin.to_hdulist(cubem)
+
+# Testing the HDU0
+@pytest.mark.parametrize(
+    "header_obj, header_values1, header_values2, header_values3, header_values4, header_values5, missing_axes",[
+        ((cubem_hdulist[0].header),(3, 0.0, 2.0, 2.0),(2E-11, 0.5, 0.4), ('m','deg','deg'),('WAVE','HPLT-TAN','HPLN-TAN'),(1E-09, 0.5, 1.0),(0, 0, 0))
+    ]
+)
+def test_meta_from_hdulist(header_obj, header_values1, header_values2, header_values3, header_values4, header_values5, missing_axes):
+    wcs_axes, crpix1, crpix2, crpix3 = header_values1
```
yashrsharma44: Here I haven't found a way to test meta values, so maybe we need to decide on the way we store meta values? So that we have only **one** way to access them.

PR #182, opened 2019-06-19
Fix bug in 2D plotting.

DanRyanIrish:
Bug was caused when both data and WCS were 2D. In this case a slice list
was being called, even though that list was only created when the data
was 2D with a >2D WCS with missing axes.

### Description

The bug is fixed by only entering the ```slice_list``` to the ```gca_wcs``` method if the ```wcs``` is greater than 2 when the data is 2D.

Fixes #<Issue Number>

### DIFF
--- a/changelog/182.bugfix.rst
+++ b/changelog/182.bugfix.rst
@@ -1,0 +1,1 @@
+Fix 2D plotting from crashing when both data and WCS are 2D.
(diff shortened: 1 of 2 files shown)


ehsteve:
Screen Shot 2019-06-19 at 2 39 21 PM

Another potential bug though. The axis on this plot are clearly arcsec but then it says deg on the axis labels.

DanRyanIrish:
Thanks for pointing out that bug @ehsteve.  It's unclear whether this is a bug in ```ndcube```, ```sunpy``` or ```astropy``` or whether it's just unexpected or limited behaviour in ```sunpy``` or ```astropy``` ```WCSAxes``` support.

I'll raise an issue and deal with it in separate PR as it's not an obvious fix.

PR #195, opened 2019-08-14
Support 3D plotting in NDCube using APE14

yashrsharma44:
This PR addresses the rewriting of `plotting.py` for `3D` `NDCube`.
##
This PR is blocked upon PR #188, so this should be merged before merging this
##
Ping @DanRyanIrish @Cadair for the reviews :stuck_out_tongue:

PR #196, opened 2019-08-18
Support plotting in NDCubeSequence using APE14

yashrsharma44:
This PR addresses the rewriting of `sequence_plotting.py` for `NDCubeSequence`.
Currently, the tests haven't been re-written to support APE14 use case but would be written in due time.
##
There is a subset of `plot` which does not work, and that needs to be addressed.
The subset which is not working is `ImageAnimatorNDCubeSequence`, because it creates a dummy axis for plotting, however adding a dummy axis seems to be not supported for APE14 WCS objects, so that might be a blocking point for `ImageAnimatorNDCubeSequence`. Either the plotting logic needs to be rewritten, or some hacks might need to be implemented :stuck_out_tongue: 
##
Merge this PR after #195 is merged.
##
Ping @Cadair @DanRyanIrish  for the reviews


ndcube/ndcube.py
```diff
@@ -258,7 +259,7 @@ def world_to_pixel(self, *quantity_axis_list):
         result = [u.Quantity(world_to_pixel[index], unit=u.pix) for index in range(self.wcs.pixel_n_dim)]
         return result[::-1]
 
-    def axis_world_coords(self, *axes, edges=False):
+    def axis_world_coords(self, axes=(), edges=False, skycoord=True):
```
Cadair: I think `values=False` would be a more consistent way of wording it with the APE 14 API. But then you only handle the SkyCoord case explicitly :thinking:

What's the motivation for this ooi?

yashrsharma44: Umm in the plotting code there are various areas where the axis_world_coords returns a 2d array, but now it returns a skycoord so that makes the thing dicey. So I thought of using a param to return a 2d array of skycoord.

DanRyanIrish:
@Cadair: Has this PR been overtaken by events?  Should it be closed and the work restarted from master?

PR #199, opened 2019-08-22
Fixes a bug associated with axes_world_coords for APE14 branch

yashrsharma44:
This PR fixes a bug associated with `axes_world_coords` in `ape14_rewrite` branch. Currently when a WCS with a `lat/lon` param is queried, a result of incorrect shape of `SkyCoords` are observed. So for fixing the bug, this PR has been made.
##
Ping @Cadair!

### DIFF
--- a/ndcube/ndcube.py
+++ b/ndcube/ndcube.py
@@ -6,6 +6,7 @@ import warnings
 import numpy as np
 import astropy.nddata
 import astropy.units as u
+from astropy.coordinates import SkyCoord
 from astropy.wcs.wcsapi.fitswcs import custom_ctype_to_ucd_mapping
 from astropy.utils.misc import InheritDocstrings
 from astropy.wcs.wcsapi.fitswcs import SlicedFITSWCS
@@ -35,20 +36,17 @@ class NDCubeABC(astropy.nddata.NDData, metaclass=NDCubeMetaClass):
         """
         Convert a pixel coordinate to a data (world) coordinate by using
         `~astropy.wcs.WCS.all_pix2world`.
-
         Parameters
         ----------
         quantity_axis_list : iterable
             An iterable of `~astropy.units.Quantity` with unit as pixel `pix`.
             Note that these quantities must be entered as separate arguments, not as one list.
-
         origin : `int`.
             Origin of the top-left corner. i.e. count from 0 or 1.
             Normally, origin should be 0 when passing numpy indices, or 1 if
             passing values from FITS header or map attributes.
             See `~astropy.wcs.WCS.wcs_pix2world` for more information.
             Default is 0.
-
         Returns
         -------
         coord : `list`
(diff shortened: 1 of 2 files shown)


ndcube/ndcube.py
```diff
@@ -258,33 +235,32 @@ def world_to_pixel(self, *quantity_axis_list):
         result = [u.Quantity(world_to_pixel[index], unit=u.pix) for index in range(self.wcs.pixel_n_dim)]
         return result[::-1]
 
-    def axis_world_coords(self, *axes, edges=False):
+    def axis_world_coords(self, axes=(), edges=False, skycoord=True):
```
Cadair: why the change from `*axes`?

ndcube/ndcube.py
```diff
@@ -327,6 +303,10 @@ def axis_world_coords(self, *axes, edges=False):
                           for axis in new_int_axes]
         n_dependent_axes = [len(da) for da in dependent_axes]
 
+        # Store the lat/lon when skycoord=False
+        TS = None
```
Cadair: what is `TS`? Can we use a more descriptive name?

ndcube/ndcube.py
```diff
@@ -355,21 +335,45 @@ def axis_world_coords(self, *axes, edges=False):
                           for k in dependent_axes[i]], indexing="ij")
                     for k, axis in enumerate(dependent_axes[i]):
                         quantity_list[axis] = dependent_pixel_quantities[k]
+
                 # Perform wcs translation
                 dependent_axes_coords = self.pixel_to_world(*quantity_list)
+
+                # Wrap the dependent_axes_coords into a list if a single quantity
+                dependent_axes_coords = dependent_axes_coords if isinstance(dependent_axes_coords, list) else [dependent_axes_coords]
```
```suggestion
                dependent_axes_coords = list(dependent_axes_coords)
```

ndcube/ndcube.py
```diff
@@ -355,21 +335,45 @@ def axis_world_coords(self, *axes, edges=False):
...
+                                axes_coords[j] = dependent_axes_coords[j]
+                            else:
+                                k = dependent_axis - j
+                                if(k==0):
+                                    TS = dependent_axes_coords[j].data.lon.deg
+                                elif(k==1):
+                                    TS = dependent_axes_coords[j].data.lat.deg
+
+                                index_of_skcoord = j+1
+                                axes_coords[j] = dependent_axes_coords[j].data.lat.deg if k==0 else dependent_axes_coords[j].data.lon.deg
```
Cadair: Do we want to return values here? Don't we normally return quantities from this method?

```suggestion
                                axes_coords[j] = dependent_axes_coords[j].data.lat if k==0 else dependent_axes_coords[j].data.lon
```

ndcube/ndcube.py
yashrsharma44: I was getting an error of using keyword arguments after `*axes`.

ndcube/ndcube.py
yashrsharma44: Umm, that is for `Tx/Ty` of `skycoord`. The name can be improved

ndcube/ndcube.py
yashrsharma44: This method should be used internally, and many times I had to convert it into `deg` in plotting, so that's the point of it.

DanRyanIrish:
File conflicts, what's the status of this PR @yashrsharma44 and @Cadair ?

Cadair:
An alternative solution to this is in the 2.0 branch now.

PR #231, opened 2020-02-03
NDCollection

DanRyanIrish:
### Description

This PR introduces a new class for holding a collection of NDCubes/Sequences.  This draft assumes that all sub-cubes are aligned along at least some axes.

Fixes #<Issue Number>

Still To Do:
- [ ] Add ```NDCollection``` to guide in docs
- [ ] Extend tests to ```NDCollection``` of ```NDCubeSequences```

### DIFF
--- a/ndcube/ndcollection.py
+++ b/ndcube/ndcollection.py
@@ -145,7 +145,7 @@ Aligned world physical axis types: {aligned_axis_types}""".format(
             if item_is_strings:
                 new_data = [self[_item] for _item in item]
                 new_keys = item
-                new_aligned_axes=tuple([self.aligned_axes[_item] for _item in item])
+                new_aligned_axes = tuple([self.aligned_axes[_item] for _item in item])
 
             # Else, the item is assumed to be a typical slicing item.
             # Slice each cube in collection using information in this item.
@@ -240,7 +240,7 @@ Aligned world physical axis types: {aligned_axis_types}""".format(
         return popped_cube
 
     def update(self, key, data, aligned_axes):
-        """Updates collection with in cube."""
+        """Updates existing cube within collection or adds new cube."""
         # Ensure new cube is of supported type.
         if not isinstance(data, SUPPORTED_COLLECTION_TYPES):
             raise TypeError("data must be an instance of {0}. Type is actually {1}".format(
@@ -249,7 +249,8 @@ Aligned world physical axis types: {aligned_axis_types}""".format(
         if isinstance(aligned_axes, int):
             aligned_axes = (aligned_axes,)
         sanitized_axes, n_sanitized_axes = collection_utils._sanitize_aligned_axes(
-                [self[self._first_key], data], (self.aligned_axes[self._first_key], aligned_axes), 2)
+                [self[self._first_key], data],
+                (self.aligned_axes[self._first_key], aligned_axes), 2)
         # Update collection
         super().update({key: data})
         self.aligned_axes.update({key: sanitized_axes[-1]})
@@ -259,4 +260,3 @@ Aligned world physical axis types: {aligned_axis_types}""".format(
         self.aligned_axes.__delitem__(key)
         if key == self._first_key:
             self._first_key = list(self.keys())[0]
-
--- a/ndcube/ndcollection.py
+++ b/ndcube/ndcollection.py
@@ -59,36 +59,22 @@ class NDCollection(dict):
         if len(keys) != len(data):
             raise ValueError("Data and keys inputs of different lengths.")
 
-        self._first_key = keys[0]
-        self._cube_types = type(data[0])
-
-        # Enter data into object.
-        super().__init__(zip(keys, data))
-        self.meta = meta
-
-        n_cubes = len(data)
-        # If aligned_axes not set, assume all axes are aligned in order.
-        if aligned_axes.lower() == "all":
+        # If aligned_axes set to "all", assume all axes are aligned in order.
+        if isinstance(aligned_axes, str) and aligned_axes.lower() == "all":
             # Check all cubes are of same shape
             cube0_dims = data[0].dimensions
             cubes_same_shape = all([all(d.dimensions == cube0_dims) for d in data])
             if cubes_same_shape is not True:
                 raise ValueError(
                     "All cubes in data not of same shape. Please set aligned_axes kwarg.")
-            self.n_aligned_axes = len(cube0_dims)
-            self.aligned_axes = dict([(k, tuple(range(len(cube0_dims)))) for k in keys])
+            sanitized_axes = dict([(k, tuple(range(len(cube0_dims)))) for k in keys])
         elif aligned_axes is None:
-            self.n_aligned_axes = 0
-            self.aligned_axes = None
+            sanitized_axes = None
         else:
             # Else, sanitize user-supplied aligned axes.
-            if dont_sanitize_aligned_axes is True:
-                self.n_aligned_axes = len(aligned_axes[0])
-                self.aligned_axes = dict(zip(keys, aligned_axes))
-            else:
-                aligned_axes, self.n_aligned_axes = collection_utils._sanitize_aligned_axes(
-                        data, aligned_axes, n_cubes)
-                self.aligned_axes = dict(zip(keys, aligned_axes))
+            sanitized_axes = collection_utils._sanitize_aligned_axes(data, aligned_axes)
+
+        NDCollection._init_without_sanitization_(data, keys, sanitized_axes, meta=meta)
 
     def __repr__(self):
         return (textwrap.dedent("""
@@ -156,8 +142,8 @@ class NDCollection(dict):
                     # Therefore the collection keys remain unchanged.
                     new_keys = list(self.keys())
 
-            return self.__class__(new_data, keys=new_keys, aligned_axes=new_aligned_axes,
-                                  meta=self.meta, dont_sanitize_aligned_axes=True)
+            return NDCollection._init_without_sanitization_(new_data, new_keys, new_aligned_axes,
+                                                            meta=self.meta)
 
     def _generate_collection_getitems(self, item):
         # There are 3 supported cases of the slice item: int, slice, tuple of ints and/or slices.
@@ -247,3 +233,27 @@ class NDCollection(dict):
         if key == self._first_key:
             self._first_key = list(self.keys())[0]
 
+    @classmethod
+    def _init_without_sanitization_(self, data, keys, aligned_axes, meta=None):
+        """
+        Initialize an NDCollection without sanitizing axes.
+
+        See __init__ docstring for parameters.
+        Note that aligned_axes is an arg here, not a kwarg and that it must be None
+        or a tuple of tuple of ints. A single int or tuple of ints is not supported here.
+
+        """
+        # Enter data into object.
+        super().__init__(dict(zip(keys, data)))
+        self.meta = meta
+
+        self._first_key = keys[0]
+        self._cube_types = type(data[0])
+
+        # Attach aligned axes to object
+        if aligned_axes is None:
+            self.n_aligned_axes = 0
+            self.aligned_axes = aligned_axes
+        else:
+            self.n_aligned_axes = len(aligned_axes[0])
+            self.aligned_axes = dict(zip(keys, aligned_axes))
--- a/ndcube/ndcollection.py
+++ b/ndcube/ndcollection.py
@@ -12,7 +12,7 @@ SUPPORTED_COLLECTION_TYPES = (NDCube, NDCubeSequence)
 __all__ = ["NDCollection"]
 
 class NDCollection(dict):
-    def __init__(self, data, keys=None, aligned_axes=None, meta=None, dont_sanitize_aligned_axes=False):
+    def __init__(self, data, keys, aligned_axes="All", meta=None, dont_sanitize_aligned_axes=False):
         """
         A class for holding and manipulating a collection of aligned NDCube or NDCubeSequences.
 
@@ -28,12 +28,12 @@ class NDCollection(dict):
             there must be one per element in the data input.
             Default is ("0", "1",...)
 
-        aligned_axes: `tuple` of `int`, `tuple` of `tuple`s of `int`
+        aligned_axes: `tuple` of `int`, `tuple` of `tuple`s of `int`, or None (Optional)
             Axes of each cube/sequence that are aligned in numpy order.
             If elements are int, then the same axis numbers in all cubes/sequences are aligned.
             If elements are tuples of ints, then must be one tuple for every cube/sequence.
             Each element of each tuple gives the axes of each cube/sequence that are aligned.
-            Default is all axes are aligned.
+            Default="All", i.e. all axes are aligned.
 
         meta: `dict` (Optional)
             General metadata for the overall collection.
@@ -75,7 +75,7 @@ class NDCollection(dict):
 
         n_cubes = len(data)
         # If aligned_axes not set, assume all axes are aligned in order.
-        if aligned_axes is None:
+        if aligned_axes == "All":
             # Check all cubes are of same shape
             cube0_dims = data[0].dimensions
             cubes_same_shape = all(
@@ -86,6 +86,9 @@ class NDCollection(dict):
             self.n_aligned_axes = len(cube0_dims)
             self.aligned_axes = dict([(keys[i], tuple(range(len(cube0_dims))))
                                       for i in range(n_cubes)])
+        elif aligned_axes is None:
+            self.n_aligned_axes = 0
+            self.aligned_axes = None
         else:
             # Else, sanitize user-supplied aligned axes.
             if dont_sanitize_aligned_axes is True:
@@ -146,16 +149,20 @@ Aligned world physical axis types: {aligned_axis_types}""".format(
 
             # Else, the item is assumed to be a typical slicing item.
             # Slice each cube in collection using information in this item.
+            # However, this can only be done if there are aligned axes.
             else:
-                # Derive item to be applied to each cube in collection and
-                # whether any aligned axes are dropped by the slicing.
-                collection_items, new_aligned_axes = self._generate_collection_getitems(item)
-                # Apply those slice items to each cube in collection.
-                new_data = [self[key][tuple(cube_item)]
-                            for key, cube_item in zip(self, collection_items)]
-                # Since item is not strings, no cube in collection is dropped.
-                # Therefore the collection keys remain unchanged.
-                new_keys = list(self.keys())
+                if self.aligned_axes is None:
+                    raise IndexError("Cannot slice unless collection has aligned axes.")
+                else:
+                    # Derive item to be applied to each cube in collection and
+                    # whether any aligned axes are dropped by the slicing.
+                    collection_items, new_aligned_axes = self._generate_collection_getitems(item)
+                    # Apply those slice items to each cube in collection.
+                    new_data = [self[key][tuple(cube_item)]
+                                for key, cube_item in zip(self, collection_items)]
+                    # Since item is not strings, no cube in collection is dropped.
+                    # Therefore the collection keys remain unchanged.
+                    new_keys = list(self.keys())
 
             return self.__class__(new_data, keys=new_keys, aligned_axes=new_aligned_axes,
                                   meta=self.meta, dont_sanitize_aligned_axes=True)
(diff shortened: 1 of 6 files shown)


ndcube/ndcollection.py
```diff
@@ -0,0 +1,86 @@
...
+        self.labels = labels
+
+    def __getitem__(self, item):
+        try:
+            item_strings = [isinstance(_item, str) for _item in item]
+            item_is_strings = all(item_strings)
+            # Ensure strings are not mixed with slices.
+            if (not item_is_strings) and (not all(np.invert(item_strings))):
+                raise TypeError("Cannot mix labels and non-labels when indexing instance.")
+        except:
```
Cadair: This will catch the `TypeError` above, also you shouldn't have bare `except:`s

DanRyanIrish:
ping @Cadair.

First draft.

ndcube/ndcollection.py
```diff
@@ -0,0 +1,86 @@
+
+import numpy as np
+
+from ndcube import NDCube
+from ndcube.utils.cube import convert_extra_coords_dict_to_input_format
+
+class NDCollection():
```
Cadair: you should inherit from `collections.Sequence`

ndcube/ndcollection.py
```diff
@@ -0,0 +1,86 @@
...
+        # to ensure all cubes have same WCS.
+        # This should be relaxed in future versions.
+        new_data = [NDCube(cube.data, wcs, uncertainty=cube.uncertainty, mask=cube.mask,
+                           unit=cube.unit, meta=cube.meta,
+                           extra_coords=convert_extra_coords_dict_to_input_format(
+                               cube.extra_coords, cube.missing_axes))
+                    for cube in data]
+        data = new_data
+
+        self.data = dict([(labels[i], data[i]) for i in range(len(data))])
```
Cadair:
```suggestion
        self.data = dict(zip(labels, data))
```

ndcube/ndcollection.py
```diff
@@ -0,0 +1,86 @@
...
+            item_is_strings = False
+        if isinstance(item, str) or item_is_strings:
+            if isinstance(item, str):
+                item = [item]
+            new_data = [self.data[_item] for _item in item]
+            new_labels = item
+        else:
+            new_labels = self.labels
+            new_data = [self.data[key][item] for key in new_labels]
+        return self.__class__(new_data, labels=new_labels, wcs=self.data[new_labels[0]].wcs)
```
Cadair: The `wcs=` section of this is problematic in the future, maybe we could implement a (for now) private method like `_from_cubes` which just allows you to pass in a list of `NDCube` objects and assumes the WCSes match?

Or that method could test if they are all the same object as the test for equality.

ndcube/ndcollection.py
```diff
@@ -0,0 +1,86 @@
...
+        cube_types = type(self.data[self.labels[0]])
+        n_cubes = len(self.labels)
+        return ("""NDCollection
+----------
+Cube labels: {labels}
+Number of Cubes: {n_cubes}
+Cube Types: {cube_types}
+Aligned dimensions: {aligned_dims}""".format(labels=self.labels, n_cubes=n_cubes,
+                                             cube_types=cube_types,
+                                             aligned_dims=self.dimensions))
```
Cadair: I propose you implement `self.aligned_dimensions` now and have it return `self.dimensions` for to future proof the API.

ndcube/ndcollection.py
```diff
@@ -0,0 +1,86 @@
...
+    def __getitem__(self, item):
+        try:
+            item_strings = [isinstance(_item, str) for _item in item]
+            item_is_strings = all(item_strings)
+            # Ensure strings are not mixed with slices.
+            if (not item_is_strings) and (not all(np.invert(item_strings))):
+                raise TypeError("Cannot mix labels and non-labels when indexing instance.")
+        except:
+            item_is_strings = False
+        if isinstance(item, str) or item_is_strings:
```
Cadair: rather than repeating the first clause of this if statement, why not split it out into two ifs at this level?

ndcube/ndcollection.py
Cadair: What's this except trying to accomplish anyway?

ndcube/ndcollection.py
```diff
@@ -0,0 +1,86 @@
...
+class NDCollection():
+    def __init__(self, data, labels, wcs):
+        """
+        A class for holding and manipulating a collection of aligned NDCube or NDCubeSequences.
+
+        Data cubes/sequences must be aligned, i.e. have the same WCS and be the same shape.
+
+        Parameters
+        ----------
+        data: sequence of `~ndcube.NDCube` or `~ndcube.NDCubeSequence`
```
Cadair: I get that this makes sense in the future, but it feels misleading in the present. If we are enforcing `data` to be `NDCube` then the `.wcs` makes no sense, we would be better off using the first cube in the lists wcs. If we are requiring `wcs` as an arg, we would be better off requiring arrays here?

Cadair:
I like it, some general comments though.

ndcube/ndcollection.py
DanRyanIrish: I'm pretty sure I've removed any dependency on NDCube or NDCubeSequence.  This class currently assumes the WCS's are aligned but doesn't check.  That's something we could add.

ndcube/ndcollection.py
DanRyanIrish: Fixed.

ndcube/ndcollection.py
DanRyanIrish: Agreed.  In the current version, I simply assume without checking that the cubes's WCSs are aligned along the axes specified by the user.  We can discuss whether we should add checks.

ndcube/ndcollection.py
DanRyanIrish: Removed

ndcube/ndcollection.py
```diff
@@ -0,0 +1,202 @@
...
+            The data cubes/sequences to held in the collection.
+
+        labels: sequence of `str`
+            Name of each cube/sequence. Each label must be unique and
+            there must be one per element in the data input.
+            Default is ("0", "1",...)
+
+        aligned_axes: `tuple` of `int`, `tuple` of `tuple`s of `int`
```
Cadair: better to use physical type here?

DanRyanIrish: That's a much more elegant solution.  Two things we would have to think through first:
* Could users want to align axes that aren't of the same physical type?  As it is now, the only requirement is that the aligned axes are the same length which gives maximum flexibility.
* Could users have axes with the same physical type?  We haven't addressed such a use case in ```NDCube```, but I wonder if it's worth considering here?

DanRyanIrish: Perhaps we could support both APIs?  A ```tuple``` of ```str``` is parsed as physical axis types while a ```tuple``` of ```int``` or ```tuple``` is parsed as data axis numbers.  Either way, the information should be stored under-the-hood in the way the current API is set up.

DanRyanIrish: On second thoughts, are we able to support using physical type at the moment.  We had a previous discussion that revealed that strictly speaking, ```NDCube.world_axis_physical_types``` does not always have a one-to-one correspondence with the data axes, especially in the cases of coupled axes and rotated data.  Although in most cases this is effectively true, we nonetheless cannot strictly, uniquely derive the data axes from the world_axis_physics_types.  We can only go the other way.  Therefore we need to:
1. Redefine ```NDCube.world_axis_physical_types```
2. Create a new attribute that more strictly related physical axis types to data axes, even in the case of coupled axes.
3. Not use physical types here to derive the aligned data axes.

Thoughts?

DanRyanIrish:
@Cadair, I think this is ready for consideration for merging.

DanRyanIrish:
@Cadair or @nabobalis, can someone explain to me why the tox tests are failing?  I don't know how to interpret the error logs.

nabobalis:
You mean?

> xfin = isfinite(x) 

E TypeError: ufunc 'isfinite' not supported for the input types, and the inputs could not be safely coerced to any supported types according to the casting rule ''safe''

DanRyanIrish:
Hi @nabobalis.  I didn't find that message in the logs.  I wasn't aware that ```isfinite``` was used in the ```ndcube``` code base.  Is that what tox is failing in all the tests?  Can you clarify this a bit further?

nabobalis:
Does this link work for you? 
https://dev.azure.com/sunpy/ndcube/_build/results?buildId=5988&view=logs&jobId=7073c21a-1e55-56ea-46de-8524ee585145&j=7073c21a-1e55-56ea-46de-8524ee585145&t=250b56fd-687b-54f8-d1c2-ad4a36574572

nabobalis:
It seems that any call to: `assert np.allclose(output.axis_ranges[i], expected_axis_ranges[i])` basically errors out with:

```
xfin = isfinite(x)

E TypeError: ufunc 'isfinite' not supported for the input types, and the inputs could not be safely coerced to any supported types according to the casting rule ''safe''
```

DanRyanIrish:
To Do: Add a changelog file.

ndcube/ndcollection.py
```diff
@@ -0,0 +1,262 @@
...
+        ----------
+        data: sequence of `~ndcube.NDCube` or `~ndcube.NDCubeSequence`
+            The data cubes/sequences to held in the collection.
+
+        keys: sequence of `str`
+            Name of each cube/sequence. Each label must be unique and
+            there must be one per element in the data input.
+            Default is ("0", "1",...)
+
+        aligned_axes: `tuple` of `int`, `tuple` of `tuple`s of `int`, or None (Optional)
```
Cadair:
```suggestion
        aligned_axes: `tuple` of `int`, `tuple` of `tuple`s of `int`, or None, optional
```

ndcube/ndcollection.py
```diff
@@ -0,0 +1,262 @@
...
+            Axes of each cube/sequence that are aligned in numpy order.
+            If elements are int, then the same axis numbers in all cubes/sequences are aligned.
+            If elements are tuples of ints, then must be one tuple for every cube/sequence.
+            Each element of each tuple gives the axes of each cube/sequence that are aligned.
+            Default="All", i.e. all axes are aligned.
+
+        meta: `dict` (Optional)
```
Cadair:
```suggestion
        meta: `dict`, optional
```

ndcube/ndcollection.py
```diff
@@ -0,0 +1,262 @@
...
+            if isinstance(item, collections.abc.Sequence):
+                item_strings = [isinstance(_item, str) for _item in item]
+                item_is_strings = all(item_strings)
+                # Ensure strings are not mixed with slices.
+                if (not item_is_strings) and (not all(np.invert(item_strings))):
+                    raise TypeError("Cannot mix keys and non-keys when indexing instance.")
+
+            # If sequence is all strings, extract the cubes corresponding to the string keys.
+            if item_is_strings:
+                new_data = [self[_item] for _item in item]
```
Cadair: same as above about the underscores

ndcube/ndcollection.py
```diff
@@ -0,0 +1,262 @@
...
+        # Case 1: int
+        # First aligned axis is dropped.
+        if isinstance(item, int):
+            drop_aligned_axes_indices = [0]
+            # Insert item to each cube's slice item.
+            for i, key in enumerate(self):
+                collection_items[i][self.aligned_axes[key][0]] = item
+
+        # Case 2: slice
+        # If only one element in interval of slice, first aligned axis is dropped.
```
Cadair: is this true? A length 1 slice doesn't drop a dimension in a numpy array:
```
$ a = np.random.random([10, 10])                                                                                                                                                                                                                               
$ a[0:1].shape                                                                                                                                                                                                                                                 
(1, 10)
```

ndcube/ndcollection.py
```diff
@@ -0,0 +1,262 @@
...
+            # Note that slice interval's of 1 result in an axis of length 1.
+            # The axis is not dropped.
+
+        # Case 3: tuple of ints/slices
+        # Search sub-items within tuple for ints or 1-interval slices.
+        elif isinstance(item, tuple):
+            # Ensure item is not longer than number of aligned axes
+            if len(item) > self.n_aligned_axes:
+                raise IndexError("Too many indices")
+            # If item is tuple, search sub-items for ints or 1-interval slices.
```
Cadair: same concern as above about length 1 slices.

ndcube/ndcollection.py
```diff
@@ -0,0 +1,262 @@
...
+            for i, axis_item in enumerate(item):
+                if isinstance(axis_item, int):
+                    drop_aligned_axes_indices.append(i)
+                for j, key in enumerate(self):
+                    collection_items[j][self.aligned_axes[key][i]] = axis_item
+
+        else:
+            raise TypeError("Unsupported slicing type: {0}".format(axis_item))
```
Cadair:
```suggestion
            raise TypeError(f"Unsupported slicing type: {axis_item}")
```

ndcube/ndcollection.py
```diff
@@ -0,0 +1,262 @@
...
+            # Check all cubes are of same shape
+            cube0_dims = data[0].dimensions
+            cubes_same_shape = all(
+                [all(data[i].dimensions == cube0_dims) for i in range(n_cubes)])
+            if cubes_same_shape is not True:
+                raise ValueError(
+                    "All cubes in data not of same shape. Please set aligned_axes kwarg.")
+            self.n_aligned_axes = len(cube0_dims)
+            self.aligned_axes = dict([(keys[i], tuple(range(len(cube0_dims))))
+                                      for i in range(n_cubes)])
```
Cadair: Would this not be better as a `collections.defaultdict` object?

ndcube/ndcollection.py
```diff
@@ -0,0 +1,262 @@
...
+        raise NotImplementedError("NDCollection does not support popitem.")
+
+    def pop(self, key):
+        """Removes the cube corresponding to the key from the collection and returns it."""
+        # Extract desired cube from collection.
+        popped_cube = super().pop(key)
+        # Delete corresponding aligned axes
+        popped_aligned_axes = self.aligned_axes.pop(key)
+        # If first key removed, update.
+        if key == self._first_key:
```
Cadair: rather than having a static `self._first_key` why not make it a property so it's dynamic?

ndcube/ndcollection.py
```diff
@@ -0,0 +1,262 @@
...
+            self._first_key = list(self.keys())[0]
+
+        return popped_cube
+
+    def update(self, key, data, aligned_axes):
+        """Updates existing cube within collection or adds new cube."""
+        # Ensure new cube is of supported type.
+        if not isinstance(data, SUPPORTED_COLLECTION_TYPES):
+            raise TypeError("data must be an instance of {0}. Type is actually {1}".format(
```
Cadair: f-strings are your friend.

ndcube/ndcollection.py
```diff
@@ -0,0 +1,262 @@
...
+        # If first key removed, update.
+        if key == self._first_key:
+            self._first_key = list(self.keys())[0]
+
+        return popped_cube
+
+    def update(self, key, data, aligned_axes):
+        """Updates existing cube within collection or adds new cube."""
+        # Ensure new cube is of supported type.
+        if not isinstance(data, SUPPORTED_COLLECTION_TYPES):
```
Cadair: Also see comment in the `__init__` about validating types.

ndcube/ndcollection.py
```diff
@@ -0,0 +1,262 @@
...
+        popped_cube = super().pop(key)
+        # Delete corresponding aligned axes
+        popped_aligned_axes = self.aligned_axes.pop(key)
+        # If first key removed, update.
+        if key == self._first_key:
+            self._first_key = list(self.keys())[0]
+
+        return popped_cube
+
+    def update(self, key, data, aligned_axes):
```
Cadair: I am not a big fan about how this method deviates from the `dict.update` method. Should it not just take in another instance of `NDCollection`? like `dict.update` takes a `dict`? We could add an extend method which takes these three arguments.

ndcube/tests/test_ndcollection.py
```diff
@@ -0,0 +1,117 @@
...
+
+# Define sequences.
+sequence02 = NDCubeSequence([cube0, cube2])
+sequence20 = NDCubeSequence([cube2, cube0])
+
+# Define collections
+aligned_axes = ((1, 2), (2, 0), (1, 2))
+keys = ("cube0", "cube1", "cube2")
+cube_collection = NDCollection([cube0, cube1, cube2], keys, aligned_axes)
+#seq_collection = NDCollection([sequence01, sequence12], ("seq0", "seq1"))
```
Cadair: All of these things should be pytest fixtures.

ndcube/utils/collection.py
```diff
@@ -0,0 +1,115 @@
+import numpy as np
+import astropy.units as u
+
+
+def _sanitize_aligned_axes(data, aligned_axes, n_cubes):
```
Cadair: The length of this function hurts my brain.

Without looking over it, I feel like if we need this much code to parse the input, our input format is too complex. Is there a way we can split this into two argument, or disallow some things (we could drop int and enforce a len 1 tuple etc).

ndcube/ndcollection.py
```diff
@@ -0,0 +1,262 @@
...
+        """
+        # Check inputs
+        # Ensure there are no duplicate keys
+        if keys is None:
+            keys = np.arange(len(data)).astype("str")
+        elif len(set(keys)) != len(keys):
+            raise ValueError("Duplicate keys detected.")
+        if len(keys) != len(data):
+            raise ValueError("Data and keys inputs of different lengths.")
+        data_types = [isinstance(cube, SUPPORTED_COLLECTION_TYPES) for cube in data]
```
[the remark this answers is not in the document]
DanRyanIrish: What's duck typing?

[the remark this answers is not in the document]
Cadair: https://en.wikipedia.org/wiki/Duck_typing

If an object provides the behaviour we need we should use it without it having to explicitly be the type we think it should be.

ndcube/ndcollection.py
```diff
@@ -0,0 +1,262 @@
...
+                    "All cubes in data not of same shape. Please set aligned_axes kwarg.")
+            self.n_aligned_axes = len(cube0_dims)
+            self.aligned_axes = dict([(keys[i], tuple(range(len(cube0_dims))))
+                                      for i in range(n_cubes)])
+        elif aligned_axes is None:
+            self.n_aligned_axes = 0
+            self.aligned_axes = None
+        else:
+            # Else, sanitize user-supplied aligned axes.
+            if dont_sanitize_aligned_axes is True:
```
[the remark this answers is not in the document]
DanRyanIrish: The sanitization process could be a bit slow but I think is important to ensure users have not tried to do something unsupported.  However, when initiating a new instance after slicing we know (unless there's a bug) that the aligned axes are correct.  Therefore there's no need to slow down the slicing process by needless sanitization.

ndcube/ndcollection.py
```diff
@@ -0,0 +1,262 @@
...
+    def aligned_dimensions(self):
+        return self[self._first_key].dimensions[np.array(self.aligned_axes[self._first_key])]
+
+    @property
+    def aligned_world_axis_physical_types(self):
+        axis_types = np.array(self[self._first_key].world_axis_physical_types)
+        return tuple(axis_types[np.array(self.aligned_axes[self._first_key])])
+
+    def __getitem__(self, item):
+        # There are two ways to slice:
```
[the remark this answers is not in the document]
DanRyanIrish: So you mean something like ```self.index_by_key``` and self.index_by_slice``` and then have ```__getitem__``` check the input and call the relevant public method?

ndcube/ndcollection.py
DanRyanIrish: You're absolutely right.  This was an old comment that needs updating.  The expected behaviour of ```NDCollection``` should be consistent with your example above.

ndcube/ndcollection.py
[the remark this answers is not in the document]
Cadair: I would suggest we have a class method for instantiation without sanitisation (potentially private) rather than a kwarg to the constructor.

ndcube/ndcollection.py
DanRyanIrish: How fast is converting dict keys to a list in order to extract one key?  ```list(self.keys())[0]```  Is this something we want to do every time we refer to ```self._first_key```?

ndcube/ndcollection.py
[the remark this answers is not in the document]
Cadair:
```python
@classmethod
def _create_skip_santization(cls, *args, **kwargs):
    pass
```

ndcube/ndcollection.py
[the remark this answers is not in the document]
Cadair: yep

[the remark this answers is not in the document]
DanRyanIrish: ```index_by_key``` could call get and error if default returned

ndcube/ndcollection.py
DanRyanIrish: No!

PR #236, opened 2020-03-17
Remove obsolete InheritDocstrings

dstansby:
Astropy is raising the following warning for me:
```
WARNING: AstropyDeprecationWarning: The InheritDocstrings class is deprecated and may be removed in a future version.
        Use Sphinx>=1.7 automatically inherits docstring instead. [abc]
```
which I think is coming from ndcube. This PR should remove the warning, and would require sphinx>1.7, which was [released in Feb 2018](https://www.sphinx-doc.org/en/master/changes.html#release-1-7-0-released-feb-12-2018).

### DIFF
--- a/ndcube/ndcube.py
+++ b/ndcube/ndcube.py
@@ -5,7 +5,6 @@ import warnings
 import numpy as np
 import astropy.nddata
 import astropy.units as u
-from astropy.utils.misc import InheritDocstrings
 
 from ndcube import utils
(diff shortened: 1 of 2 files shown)


Cadair:
Thanks for getting around to this! Can you pin the sphinx version in `setup.cfg` just for extra clarity?

DanRyanIrish:
Thanks @dstansby!

PR #238, opened 2020-03-18
NDCollection 2

DanRyanIrish:
### Description
This PR is a continuation of #231.  (I messed up the git history and it was easier to open a new PR.)  As with #231, this PR introduces a new class for holding a collection of NDCubes/Sequences.

Still To Do
- [x] Get refactor of ```NDCollection.__init__``` working.
- [x] Rename ```NDCollection.update``` to something like ```NDCollection.extend``` and implement a new ```NDCollection.update``` that takes another ```NDCollection``` in the same way the ```dict.update``` takes another ```dict```.
- [x] Make ```NDCollection._first_key``` and ```NDCollection._cube_types``` properties.
- [x] Add ```NDCollection``` to guide in docs
- [x] Extend tests to ```NDCollection``` of ```NDCubeSequences```
- [x] Change ```aligned_axes``` default to ```None```
- [x] Changed ```NDCollection``` input to ```(key, value)``` pairs, like ```dict```.
- [x] Move ```_sanitize_aligned_axes``` to ```ndcollection.py```?


ndcube/ndcollection.py
```diff
@@ -0,0 +1,258 @@
...
+    def _init_without_sanitization(self, data, keys, aligned_axes, meta=None):
+        """
+        Initialize NDCollection without sanitizing inputs.
+
+        Inputs same as NDCollection.__init__ except aligned_axes is an arg and
+        must be None or a tuple or tuples of ints.
+        Single ints or tuple of ints not supported here.
+        """
+        # Enter data into object.
+        super().__init__(zip(keys, data))
```
DanRyanIrish: @Cadair This line no longer works.  I think it's because it's now in a classmethod.  Think am doing something wrong here...

ndcube/ndcollection.py
```diff
@@ -0,0 +1,258 @@
...
+        self.aligned_axes.update({key: sanitized_axes[-1]})
+
+    def __delitem__(self, key):
+        super().__delitem__(key)
+        self.aligned_axes.__delitem__(key)
+        if key == self._first_key:
+            self._first_key = list(self.keys())[0]
+
+    @classmethod
+    def _init_without_sanitization(self, data, keys, aligned_axes, meta=None):
```
Cadair:
```suggestion
    def _init_without_sanitization(cls, data, keys, aligned_axes, meta=None):
```

ndcube/ndcollection.py
Cadair: This should still be in the `__init__`, it kinda has to be.

DanRyanIrish: So maybe go back to the ```dont_sanitize``` kwarg in the init?

ndcube/ndcollection.py
DanRyanIrish: What is the difference between ```cls``` and ```self```?

ndcube/ndcollection.py
```diff
@@ -0,0 +1,261 @@
...
+        This is interpretted as axis 0 of cube0 is aligned with axis 2 of cube1 while
+        axis 1 of cube0 is aligned with axis 1 of cube1.
+
+        """
+        # Unzip key_data_pairs
+        keys, data = zip(*(key_data_pairs))
+
+        # Sanitize inputs unless hidden kwarg indicates not to.
+        sanitize_inputs = kwargs.get("sanitize_inputs", True)
+        if sanitize_inputs is True:
```
Cadair:
```suggestion
        if sanitize_inputs:
```

ndcube/ndcollection.py
```diff
@@ -0,0 +1,261 @@
...
+            are aligned, as are the 1st, and so on.
+            Default=None
+
+        meta: `dict`, optional
+            General metadata for the overall collection.
+
+        Example
+        -------
+        Say the collection holds two NDCubes, each of 3 dimensions.
+        aligned_axes = (1, 2)
```
Cadair:
```suggestion
        ``aligned_axes = (1, 2)``
```

ndcube/ndcollection.py
```diff
@@ -0,0 +1,261 @@
...
+        means that axis 1 (0-based counting) of cube0 is aligned with axis 1 of cube1,
+        and axis 2 of cube0 is aligned with axis 2 of cube1.
+        However, if
+        aligned_axes = ((0, 1), (2, 1))
```
Cadair:
```suggestion
        ``aligned_axes = ((0, 1), (2, 1))``
```

ndcube/ndcollection.py
```diff
@@ -0,0 +1,261 @@
...
+        then the first tuple corresponds to cube0 and the second with cube1.
+        This is interpretted as axis 0 of cube0 is aligned with axis 2 of cube1 while
+        axis 1 of cube0 is aligned with axis 1 of cube1.
+
+        """
+        # Unzip key_data_pairs
```
Cadair: This comment is self-fulfilling :stuck_out_tongue: 
```suggestion
```

ndcube/ndcollection.py
```diff
@@ -0,0 +1,261 @@
...
+        else:
+            self.n_aligned_axes = len(aligned_axes[0])
+            self.aligned_axes = dict(zip(keys, aligned_axes))
+
+    @property
+    def _first_key(self):
+        return list(self.keys())[0]
+
+    def __repr__(self):
+        return (textwrap.dedent("""
```
Cadair: I think you need this:
```suggestion
        return (textwrap.dedent("""\
```
so there isn't a line break at the start of the string.

ndcube/ndcollection.py
```diff
@@ -0,0 +1,261 @@
...
+            NDCollection
```
Cadair: I like reprs to start with the object repr (so you know what type and what id it is).
```suggestion
            {object.__repr__(self)}
            NDCollection
```

I also think this might be better off if you rename this implementation of `__repr__` to `__str__` and then have repr be:
```python
def __repr__(self):
    return f"{object.__repr__(self)\n{str(self)}"
```

ndcube/ndcollection.py
```diff
@@ -0,0 +1,261 @@
...
+            Cube keys: {keys}
+            Number of Cubes: {n_cubes}
+            Aligned dimensions: {aligned_dims}
+            Aligned world physical axis types: {aligned_axis_types}""".format(
+                keys=self.keys(), n_cubes=len(self),
+                aligned_dims=self.aligned_dimensions,
+                aligned_axis_types=self.aligned_world_axis_physical_types)))
+
+    @property
+    def aligned_dimensions(self):
```
Cadair: needs docstring

ndcube/ndcollection.py
```diff
@@ -0,0 +1,261 @@
...
+        if self.aligned_axes is None:
+            return None
+        else:
+            return self[self._first_key].dimensions[np.array(self.aligned_axes[self._first_key])]
+
+    @property
+    def aligned_world_axis_physical_types(self):
```
Cadair: needs docstring

ndcube/ndcollection.py
```diff
@@ -0,0 +1,261 @@
...
+
+    def pop(self, key):
+        """Removes the cube corresponding to the key from the collection and returns it."""
+        # Extract desired cube from collection.
+        popped_cube = super().pop(key)
+        # Delete corresponding aligned axes
+        popped_aligned_axes = self.aligned_axes.pop(key)
+        return popped_cube
+
+    def add_to_collection(self, key, data, aligned_axes):
```
Cadair: needs docstring

ndcube/ndcollection.py
```diff
@@ -0,0 +1,261 @@
...
+        return collection_items, new_aligned_axes
+
+    def copy(self):
+        return copy.deepcopy(self)
+
+    def setdefault(self):
+        raise NotImplementedError("NDCollection does not support setdefault.")
+
+    def popitem(self):
+        raise NotImplementedError("NDCollection does not support popitem.")
```
Cadair: I thought we were implementing this?

ndcube/ndcollection.py
```diff
@@ -0,0 +1,261 @@
...
+
+    def pop(self, key):
+        """Removes the cube corresponding to the key from the collection and returns it."""
```
Cadair: Do we need this docstring or can we inherit the `dict.pop` one?

ndcube/ndcollection.py
```diff
@@ -0,0 +1,261 @@
...
+        new_aligned_axes = collection_utils._update_aligned_axes(
+                drop_aligned_axes_indices, self.aligned_axes, self._first_key)
+
+        return collection_items, new_aligned_axes
+
+    def copy(self):
+        return copy.deepcopy(self)
+
+    def setdefault(self):
+        raise NotImplementedError("NDCollection does not support setdefault.")
```
Cadair: You also need to add a `__setitem__` which raises `NotImplementedError`.

ndcube/utils/collection.py
```diff
@@ -0,0 +1,142 @@
...
+    *. A tuple of tuples of ints, where each tuple corresponds to a cube
+    in the collection, and each int designates the an aligned axis in numpy order.
+    In this case, the axis represented by the 0th int in the 0th tuple is aligned
+    with the 0th int in the 1st tuple and so on.
+    *. A single tuple of ints if all aligned axes are in the same order.
+    *. A single int if only one axis is aligned and if the aligned axis in each cube
+    is in the same order.
+    """
+    aligned_axes_error_message = "aligned_axes must contain ints or " + \
+        "a tuple of ints for each element in data."
```
Cadair:
```suggestion
    aligned_axes_error_message = ("aligned_axes must contain ints or "
                                  "a tuple of ints for each element in data.")
```

ndcube/utils/collection.py
```diff
@@ -0,0 +1,142 @@
...
+    if isinstance(data[0].dimensions, u.Quantity):
+        cube0_dims = data[0].dimensions[np.array(aligned_axes[0])]
+    elif isinstance(data[0].dimensions, tuple):
+        cube0_dims = np.array(data[0].dimensions, dtype=object)[np.array(aligned_axes[0])]
+    else:
+        raise TypeError("0th cube in collection of unsupported type: {0}".format(type(data[0])))
```
Cadair: What exactly is this doing, do you need the different behaviour on type here, they look kinda similar?

ndcube/utils/collection.py
```diff
@@ -0,0 +1,142 @@
...
+        aligned_axes = (aligned_axes,)
+    if not isinstance(aligned_axes, tuple):
+        raise ValueError(aligned_axes_error_message)
+    # Check type of each element.
+    axes_all_ints = all([isinstance(axis, int) for axis in aligned_axes])
+    axes_all_tuples = all([isinstance(axis, tuple) for axis in aligned_axes])
+    # If all elements are int, duplicate tuple so there is one for each cube.
+    n_cubes = len(data)
+    if axes_all_ints is True:
+
```
Cadair:
```suggestion
    if axes_all_ints:
```

ndcube/utils/collection.py
```diff
@@ -0,0 +1,142 @@
...
+        n_aligned_axes = len(aligned_axes)
+        aligned_axes = tuple([aligned_axes for i in range(n_cubes)])
+
+    # If all elements are tuple, ensure there is a tuple for each cube and
+    # all elements of each sub-tuple are ints.
+    elif axes_all_tuples is True:
```
Cadair:
```suggestion
    elif axes_all_tuples:
```

ndcube/utils/collection.py
```diff
@@ -0,0 +1,142 @@
...
+            max_aligned_axis = max(aligned_axes[i])
+            if n_cube_dims < max([max_aligned_axis, n_aligned_axes]):
+                raise ValueError(
+                    "Each cube in data must have at least as many axes as aligned axes "
+                    "and aligned axis indices must be less than number of cube axes.\n"
+                    f"Cube number: {i};\n"
+                    f"Number of cube dimensions: {n_cube_dims};\n"
+                    f"No. aligned axes: {n_aligned_axes};\n"
+                    f"Highest aligned axis: {max_aligned_axis}")
+            subtuple_types = [False] * n_aligned_axes
```
Cadair: This is redefining a variable on line 54

ndcube/utils/collection.py
```diff
@@ -0,0 +1,142 @@
...
+            cube_lengths_equal = [False] * n_aligned_axes
+            for j, axis in enumerate(aligned_axes[i]):
+                subtuple_types[j] = isinstance(axis, (int, np.integer))
```
Cadair:
```suggestion
                subtuple_types[j] = isinstance(axis, numbers.Integral)
```

ndcube/utils/collection.py
```diff
@@ -0,0 +1,142 @@
+import numpy as np
+import astropy.units as u
+
+
+def _sanitize_user_aligned_axes(data, aligned_axes):
```
Cadair: This is still very long, but eh I suppose if it works... :stuck_out_tongue:

ndcube/ndcollection.py
DanRyanIrish: Interesting.  What's the motivation for this?

Cadair: Well otherwise `print(ndcollection)` gives you something that looks more like a repr and repr gives you something that looks more like it should come out of `print()`.

As for the `object.__repr__` thing, it's very important in debugging that the repr shows you the type of the object and also the memory id, so you can see if it's the instance / type you expect it to be.

ndcube/ndcollection.py
DanRyanIrish: We could probably inherit it.  How do we do that?

ndcube/ndcollection.py
Cadair: huh ok :+1:

ndcube/ndcollection.py
DanRyanIrish: Argh!  We can't do this.  ```__setitem__``` is called under the hood by ```copy```.  So it causes ```NDCollection.copy()``` and copy.deepcopy(my_ndcollection)``` to error.

ndcube/utils/collection.py
DanRyanIrish: One is for the ```NDCube``` case the other is for the ```NDCubeSequence``` case.  Remember that
```type(NDCube.dimensions)``` is ```u.Quantity``` while ```type(NDCubeSequence.dimensions)``` is a ```tuple``` of ```u.Quantity```.

However, I could change this to handle the ```NCubeSequence``` case as special and assume all other cases are compatible with the ```NDCube``` one.

**Comment by DanRyanIrish** [comment]
@Cadair, it's done! ```__setitem__``` raises and error.  ```copy``` works.  And ```pop```'s docstring is deleted.  ```update``` now works even more like ```dict.update``` and there are guide docs.  It's done!  Once approved and merged we can release 1.3!  I'm done for the week!

**Comment by DanRyanIrish** [comment]
*Drop mic!*

File: docs/ndcollection.rst [hunk 22]
```diff
@@ -0,0 +1,201 @@
...
+
+Editing NDCollection
+--------------------
+
+Because `~ndcube.NDCollection` inherits from `dict`, we can edit the 
+collection using many of the same methods.
+These have the same or analagous APIs to the ``dict`` versions and 
+include ``del``, `~ndcube.NDCollection.pop`, and `~ndcube.NDCollection.update`.
+Some `dict` methods may not be implemented on `~ndcube.NDCollection`
+if they are note consistent with its design.
```
Cadair:
```suggestion
if they are not consistent with its design.
```

ndcube/ndcollection.py
```diff
@@ -0,0 +1,261 @@
...
+        """
+        # Enter data and metadata into object.
+        super().__init__(key_data_pairs)
+        self.meta = meta
+
+        # Convert aligned axes to required format.
+        if aligned_axes is not None:
+            keys, data = zip(*key_data_pairs)
+            # Sanitize aligned axes unless hidden kwarg indicates not to.
+            if kwargs.get("sanitize_inputs", True):
```
Cadair: We should probably validate we don't have any other kwargs other than this in `kwargs` as they don't go anywhere.

ndcube/ndcollection.py
```diff
@@ -0,0 +1,261 @@
...
+        return list(self.keys())[0]
+
+    def __str__(self):
+        return (textwrap.dedent("""\
+            NDCollection
+            ------------
+            Cube keys: {keys}
+            Number of Cubes: {n_cubes}
+            Aligned dimensions: {aligned_dims}
+            Aligned world physical axis types: {aligned_axis_types}""".format(
```
Cadair: f-strings are your friend.

ndcube/ndcollection.py
```diff
@@ -0,0 +1,261 @@
...
+        """
+        Returns the lengths of all aligned axes.
+
+        If there are no aligned axes, returns None.
+
+        """
+        if self.aligned_axes is None:
+            return None
+        else:
+            return self[self._first_key].dimensions[np.array(self.aligned_axes[self._first_key])]
```
Cadair:
```suggestion
        if self.aligned_axes is not None:
            return self[self._first_key].dimensions[np.array(self.aligned_axes[self._first_key])]
```

ndcube/ndcollection.py
```diff
@@ -0,0 +1,261 @@
...
+        Returns the physical types of the aligned axes of an ND object in the collection.
+
+        If there are no aligned axes, returns None.
+
+        """
+        if self.aligned_axes is None:
+            return None
+        else:
+            axis_types = np.array(self[self._first_key].world_axis_physical_types)
+            return tuple(axis_types[np.array(self.aligned_axes[self._first_key])])
```
Cadair:
```suggestion
        if self.aligned_axes is not None:
            axis_types = np.array(self[self._first_key].world_axis_physical_types)
            return tuple(axis_types[np.array(self.aligned_axes[self._first_key])])
```

ndcube/ndcollection.py
```diff
@@ -0,0 +1,261 @@
...
+                else:
+                    # Derive item to be applied to each cube in collection and
+                    # whether any aligned axes are dropped by the slicing.
+                    collection_items, new_aligned_axes = self._generate_collection_getitems(item)
+                    # Apply those slice items to each cube in collection.
+                    new_data = [self[key][tuple(cube_item)]
+                                for key, cube_item in zip(self, collection_items)]
+                    # Since item is not strings, no cube in collection is dropped.
+                    # Therefore the collection keys remain unchanged.
+                    new_keys = list(self.keys())
```
Cadair: The else is redundant.
```suggestion
                # Derive item to be applied to each cube in collection and
                # whether any aligned axes are dropped by the slicing.
                collection_items, new_aligned_axes = self._generate_collection_getitems(item)
                # Apply those slice items to each cube in collection.
                new_data = [self[key][tuple(cube_item)]
                            for key, cube_item in zip(self, collection_items)]
                # Since item is not strings, no cube in collection is dropped.
                # Therefore the collection keys remain unchanged.
                new_keys = list(self.keys())
```

ndcube/ndcollection.py
```diff
@@ -54,10 +54,13 @@ def __init__(self, key_data_pairs, aligned_axes=None, meta=None, **kwargs):
         if aligned_axes is not None:
             keys, data = zip(*key_data_pairs)
             # Sanitize aligned axes unless hidden kwarg indicates not to.
-            if kwargs.get("sanitize_inputs", True):
+            if kwargs.pop("sanitize_inputs", True):
                 aligned_axes = collection_utils._sanitize_aligned_axes(keys, data, aligned_axes)
             else:
                 aligned_axes = dict(zip(keys, aligned_axes))
+        if kwargs != {}:
```
Cadair:
```suggestion
        if kwargs:
```

PR #243, opened 2020-03-23
Pixel axis numbers <-> world physical types

### LINKED ISSUE #240: Should world_axis_physical_types report the axis types for each pixel axis?
### Description

Currently ```NDCube.world_axis_physical_types``` (and ```NDCubeSequence.world_axis_physical_types``` and ```NDCubeSequence.cube_like_world_axis_physical_types```) simply return all the physical axis types represented in the WCS.  It does not say which pixel axis corresponds to which physical axis, even though in many cases, a naive one-to-one mapping will give the intuitively correct answer.

It would be highly useful if ```world_axis_physical_types``` or another property were to explicitly provide this mapping, both in the pixel -> physical and the physical -> pixel directions.

### Current Behavior
Say the NDCube's first two dimesions correspond longitude and latitude (i.e. they are dependent) and the third corresponds to an  independent type, e.g. wavelength.

```python
>>> len(NDCube.dimensions)  # See the number of pixel dimensions in the NDCube
3
>>> NDCube.world_axis_physical_types
('custom:pos.helioprojective.lon', 'custom:pos.helioprojective.lat', 'em.wl')
```

Although this simply returns all physical axes, a naive reading of it suggests that the first axis is longitude, the second is latitude, and the third is wavelength.  However in truth, the first two are both longitude and latitude and the 3rd is wavelength.

### Desired Behavior
```python
>>> NDCube.world_to_pixel_axis_names()  # Ordering based on output of NDCube.world_axis_physical_types
((0, 1), (0, 1), 2)
>>> NDCube.pixel_to_world_axis_names()
(('custom:pos.helioprojective.lon', 'custom:pos.helioprojective.lat'), ('custom:pos.helioprojective.lon', 'custom:pos.helioprojective.lat'), 'em.wl')
```
Perhaps these functions could have a similar API to ```NDCube.axis_world_coords()``` and take strings and axis indices to return only the values the user is interested in.
```python
>>> NDCube.world_to_pixel_axis_names('lon')
(0, 1)
>>> NDCube.pixel_to_world_axis_names(0)
('custom:pos.helioprojective.lon', 'custom:pos.helioprojective.lat')
```

These methods could not only search the WCS, but also the ```extra_coords```.

DanRyanIrish:
### Description

This PR supplies two new methods that map between physical axis names and pixel axis numbers.  They differ from ```NDCube.world_axis_physical_types``` in that they explicitly give all pixel axis numbers for each world axis physical type and vice versa.

Fixes #240

To Do:
- [x] Write versions of these methods for ```NDCubeSequence```.
- [ ] Write tests for all new methods.
- [x] Extend methods to include axis names in ```extra_coords```?

### DIFF
--- a/ndcube/ndcube.py
+++ b/ndcube/ndcube.py
@@ -669,9 +669,11 @@ Axis Types of NDCube: {axis_type}
             correspond to the physical types returned by NDCube.world_axis_physical_types.
 
         """
+        wcs_names = self.world_axis_physical_types
+        extra_coords = self.extra_coords
         # Parse user input.
         if axes_names == ():
-            axes_names = self.world_axis_physical_types
+            axes_names = wcs_names
         elif isinstance(axes_names, str):
             axes_names = (axes_names,)
         n_names = len(axes_names)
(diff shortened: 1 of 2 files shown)


ndcube/ndcube.py
```diff
@@ -592,6 +593,137 @@ def explode_along_axis(self, axis):
         # Creating a new NDCubeSequence with the result_cubes and common axis as axis
         return NDCubeSequence(result_cubes, common_axis=axis, meta=self.meta)
 
+    def pixel_axes_to_world_types(self, *axes):
```
DanRyanIrish: Add ```include_extra_coords=True``` kwarg so that users can choose to filter out ```extra_coords``` if they wish?

Thoughts @Cadair?

DanRyanIrish:
@Cadair's suggestion: Much of this can be reimplemented using: ```np.array(w.world_axis_physical_types)[w.axis_correlation_matrix[axes]]```.

DanRyanIrish:
Agreed consensus: Punt this PR to 2.0.  Implement what's needed for ```sunraster``` directly there and replace with ```ndcube 2.0``` version in future versions.

Issue #248, opened 2020-03-30
Simplify NDCubeSequence __getitem__ implementation


DanRyanIrish:
Can ```NDCubeSequence.__getitem__``` be simplified and made independent of ```utils.sequence.slice_sequence```?

Suggested change:
1. Change representation of ```NDCubeSequence.data``` from ```list``` to object ```array```.
2. Replace ```NDCubeSequence.__getitem__``` with:
```python
if isinstance(item, numbers.Integral):
    return self.data[item]
else:
    result = copy.deepcopy(self)
    if isinstance(item, slice):
            result.data = self.data[item]
    else:
        result.data = np.array([cube[item[1:]] for cube in self.data[item[0]]])
        # Also alter result._common_axis here if necessary.
    return result
```

PR #251, opened 2020-04-02
First step to simplify NDCubeSequence slicing implementation

hayesla:
First pass at #248 

To Do
- [x] Add changelog file.
- [x] Fix any failing test(s).


DanRyanIrish:
Hi @hayesla.  I think this PR should also change the type of ``` self.data``` to any array.  I think the most efficient way to do this would be to change [this line](https://github.com/sunpy/ndcube/pull/251/files#diff-6d533a9392fa63eab1371a8997e34419R35) to ```self.data = np.asanyarray(data_list)```.

I don't __think__ should cause any tests to fail, but if it does I think they should be simple enough fixes.

Also change the docstring accordingly, i.e. [this line](https://github.com/sunpy/ndcube/pull/251/files#diff-6d533a9392fa63eab1371a8997e34419R20) to
```
data: 1D `numpy.ndarray`
    Array of cubes
```

DanRyanIrish:
It also looks like there is only one test that's failing: ndcube/tests/test_sequence_plotting.py:482

This PR also needs a ```trivial```-type changelog, i.e. ```251.trivial.rst``` in the changelog directory.  See [here](https://github.com/sunpy/sunpy/blob/master/changelog/README.rst) for more instructions.

DanRyanIrish:
@hayesla, could you move the code in __getitem__ to replace the code in utils.sequence.slice_sequence and then simply have __getitem__ return the output of utils.sequence.slice_sequence. This implementation is needed in sunraster.

hayesla:
so do you mean that getitem just returns the `utils.sequence.slice_sequence` - as this works for ints, slices and tuples?

DanRyanIrish:
Yep, that's right, e.g.
```python
def __getitem__(self, item):
    return utils.sequence.slice_sequence(self, item)
```

DanRyanIrish:
Added a To Do checklist in the PR description to show what's left to do.  As you can see there is not much to do.  I am considering doing a bugfix release soon, perhaps at the start of next week?  Since this is needed for ```sunraster``` we could include it.  The ```_common_axis``` slicing bug can be fixed in a separate PR and released as part of 2.0.

How does this sound?

DanRyanIrish:
[This commit](https://github.com/sunpy/ndcube/pull/251/commits/18f88fbac4aa8f25b964ae6869048e74ee370769) doesn't do things the way I meant.  I think this is my fault for a confusing explanation.  I suggest you wind back this commit and proceed without it.  I'll remove the ```slice_sequence``` step in the above to do list too.

hayesla:
@DanRyanIrish  - I think this is more along the lines you were thinking?

It still don't fix the `common_axis` issue but as you mentioned maybe that could go in another PR?

ndcube/ndcube_sequence.py
```diff
@@ -84,14 +84,21 @@ def cube_like_world_axis_physical_types(self):
         return self.data[0].world_axis_physical_types
 
     def __getitem__(self, item):
+
```
DanRyanIrish, line 87:
```suggestion
```

changelog/251.trivial.rst
```diff
@@ -0,0 +1 @@
+use the `utils.sequence.slice_sequence` for __getitem__
```
DanRyanIrish, line 1:
```suggestion
Simplify and speed up implementation of NDCubeSequence slicing.
```

DanRyanIrish:
> It still don't fix the `common_axis` issue but as you mentioned maybe that could go in another PR?

Yes, let's keep that for a separate PR.

ndcube/ndcube_sequence.py
```diff
@@ -84,14 +84,21 @@ def cube_like_world_axis_physical_types(self):
...
         if isinstance(item, numbers.Integral):
             return self.data[item]
-        elif isinstance(item, slice):
+        else:
             result = copy.deepcopy(self)
-            result.data = self.data[item]
+            if isinstance(item, slice):
+                    result.data = self.data[item]
```
DanRyanIrish, line 93:
```suggestion
                result.data = self.data[item]
```

DanRyanIrish: Over-indented line.

DanRyanIrish:
Hmmm.  I don't understand those test failures...  Perhaps @nabobalis can interpret them for us?

DanRyanIrish:
To get the tests to behave correctly, we have to pin pytest to <5.4.  To do this change [this line](https://github.com/hayesla/ndcube/blob/ndcube_sequence_slicing/setup.cfg#L23) in ```setup.cfg``` to ```pytest < 5.4```.  Then commit and push.

Be sure to make this the only change in the commit.  It makes it easy to ```cherry-pick``` later it I need it for the bugfix release.  Thanks!!

hayesla:
done!

DanRyanIrish:
Great!  Hopefully that'll make the test fails more understandable.

nabobalis:
Seems to be:
```python
===================================================================================== ERRORS ======================================================================================
_________________________________________________ ERROR collecting .tox/py38/Lib/site-packages/ndcube/tests/test_ndcollection.py __________________________________________________
..\..\.tox\py38\Lib\site-packages\ndcube\tests\test_ndcollection.py:77: in <module>
    NDCollection([("seq0", sequence02[:, 1, 1:3]), ("seq1", sequence20[:, 1, 1:3])],
..\..\.tox\py38\Lib\site-packages\ndcube\ndcube_sequence.py:97: in __getitem__
    elif isinstance(item[0], slice) and item[0].stop - item[0].start == 1:
E   TypeError: unsupported operand type(s) for -: 'NoneType' and 'NoneType'
```

hayesla:
> Seems to be:
> 
> ```python
> ===================================================================================== ERRORS ======================================================================================
> _________________________________________________ ERROR collecting .tox/py38/Lib/site-packages/ndcube/tests/test_ndcollection.py __________________________________________________
> ..\..\.tox\py38\Lib\site-packages\ndcube\tests\test_ndcollection.py:77: in <module>
>     NDCollection([("seq0", sequence02[:, 1, 1:3]), ("seq1", sequence20[:, 1, 1:3])],
> ..\..\.tox\py38\Lib\site-packages\ndcube\ndcube_sequence.py:97: in __getitem__
>     elif isinstance(item[0], slice) and item[0].stop - item[0].start == 1:
> E   TypeError: unsupported operand type(s) for -: 'NoneType' and 'NoneType'
> ```
> 
> The mac os build failed due to threading, so we will need to add ` posargs: -n=1` to the last line of the mac os block in the azure template.

I think this may be an actually bug in the code no?

nabobalis:
I would say yes?

ndcube/ndcube_sequence.py
```diff
@@ -84,14 +84,21 @@ def cube_like_world_axis_physical_types(self):
...
-        elif isinstance(item, slice):
+        else:
             result = copy.deepcopy(self)
-            result.data = self.data[item]
+            if isinstance(item, slice):
+                result.data = self.data[item]
+            else:
+                if isinstance(item[0], numbers.Integral):
+                    result = result.data[item[0]][item[1:]]
+                elif isinstance(item[0], slice) and item[0].stop - item[0].start == 1:
```
DanRyanIrish, line 97: Just looked at this again.  I don't think this line and the one below are needed at all?  The line after the ```else``` will do the same job in all remaining cases, I'm pretty sure?  If so, that'll also get rid of the test failure.

hayesla: yeah its needed, as if you just do the list comprehension on the `NDCube` rather on the list of NDCubes it then loops through the arrays in the NDCube if that makes sense

hayesla: i.e. as in for this `result.data = [cube[item[1:]] for cube in result.data[item[0]]]` to work you need result.data[item[0]] to be a list of NDCubes rather than an NDCube. but if item[0] is a 0, or slice(0, 1) this doesn't work ... the way it is currently implemented also follows similar logic to the way the utils.sequence.slice_sequence kind of did it without the need of a `SequenceItem`

DanRyanIrish: Ah!  You're dead right!

DanRyanIrish:
```suggestion
                elif isinstance(item[0], slice):
                    start = 0 if item[0].start is None else item[0].start
                    stop = len(self.data) if item[0].stop is None else item[0].stop
                    if stop - start == 1:
                        result.data = [result.data[item[0].start][item[1:]]]
```

ndcube/ndcube_sequence.py
```diff
@@ -85,12 +85,21 @@ def cube_like_world_axis_physical_types(self):
...
+            if isinstance(item, slice):
+                result.data = self.data[item]
+            else:
+                if isinstance(item[0], numbers.Integral):
+                    result = result.data[item[0]][item[1:]]
+                elif isinstance(item[0], slice):
+                    start = 0 if item[0].start is None else item[0].start
+                    stop = len(self.data) if item[0].stop is None else item[0].stop
+                    if stop - start == 1:
+                        result.data = [result.data[item[0].start][item[1:]]]
```
DanRyanIrish, line 99:
```suggestion
                        result.data = [result.data[start][item[1:]]]
```

ndcube/ndcube_sequence.py
```diff
@@ -85,12 +85,21 @@ def cube_like_world_axis_physical_types(self):
...
+                if isinstance(item[0], numbers.Integral):
+                    result = result.data[item[0]][item[1:]]
+                elif isinstance(item[0], slice):
+                    start = 0 if item[0].start is None else item[0].start
+                    stop = len(self.data) if item[0].stop is None else item[0].stop
+                    if stop - start == 1:
+                        result.data = [result.data[start][item[1:]]]
+                else:
+                    result.data = [cube[item[1:]] for cube in result.data[item[0]]]
             return result
```
DanRyanIrish, line 102:
```suggestion
        return result
```

DanRyanIrish:
So this code passes tests now with the exception of one doctest in ```ndcubesequence.rst```.  However, when I run the same code locally after pulling @hayesla's branch, it gives the right output.  I can't understand why that is!  @nabobalis have you ever seen that before?

Comparison of test output:
Azure case:
```
189 independently!) However, with `~ndcube.NDCubeSequence` this becomes as
190 simple as indexing a single array::
191 
192   >>> regions_of_interest_in_sequence = my_sequence[1:3, 0, 1:3, 1:4]
193   >>> regions_of_interest_in_sequence.dimensions
Expected:
    (<Quantity 2. pix>, <Quantity 2. pix>, <Quantity 3. pix>)
Got:
    (<Quantity 3. pix>, <Quantity 3. pix>, <Quantity 4. pix>, <Quantity 5. pix>)

/home/vsts/work/1/s/docs/ndcubesequence.rst:193: DocTestFailure
```

Output when executing code locally:
```python
>>> regions_of_interest_in_sequence = my_sequence[1:3, 0, 1:3, 1:4]
>>> regions_of_interest_in_sequence.dimensions
(<Quantity 2. pix>, <Quantity 2. pix>, <Quantity 3. pix>)  # This is the same as what the doctest expects.
```

nabobalis:
I am not sure. There has to be some difference but what I can't say for now.

ndcube/ndcube_sequence.py
```diff
@@ -85,12 +85,21 @@ def cube_like_world_axis_physical_types(self):
...
-        elif isinstance(item, slice):
+        else:
             result = copy.deepcopy(self)
-            result.data = self.data[item]
+            if isinstance(item, slice):
+                result.data = self.data[item]
+            else:
+                if isinstance(item[0], numbers.Integral):
+                    result = result.data[item[0]][item[1:]]
+                elif isinstance(item[0], slice):
```
DanRyanIrish, line 95:
```suggestion
                else:
                    item0_is_length1_slice = False
                    if isinstance(item[0], slice):
```

ndcube/ndcube_sequence.py
```diff
@@ -85,12 +85,21 @@ def cube_like_world_axis_physical_types(self):
...
+                    start = 0 if item[0].start is None else item[0].start
```
DanRyanIrish, line 96:
```suggestion
                        start = 0 if item[0].start is None else item[0].start
```

ndcube/ndcube_sequence.py
```diff
@@ -85,12 +85,21 @@ def cube_like_world_axis_physical_types(self):
...
+                    stop = len(self.data) if item[0].stop is None else item[0].stop
```
DanRyanIrish, line 97:
```suggestion
                        stop = len(self.data) if item[0].stop is None else item[0].stop
```

ndcube/ndcube_sequence.py
```diff
@@ -85,12 +85,21 @@ def cube_like_world_axis_physical_types(self):
...
+                    if stop - start == 1:
```
DanRyanIrish, line 98:
```suggestion
                        if stop - start == 1:
```

ndcube/ndcube_sequence.py
```diff
@@ -85,12 +85,21 @@ def cube_like_world_axis_physical_types(self):
...
+                        result.data = [result.data[start][item[1:]]]
```
DanRyanIrish, line 99:
```suggestion
                        item0_is_length1_slice = True
                if item0_is_length1_slice:
                    result.data = [result.data[start][item[1:]]]
```

ndcube/ndcube_sequence.py
```diff
@@ -85,12 +85,21 @@ def cube_like_world_axis_physical_types(self):
...
+                else:
```
DanRyanIrish, line 100:
```suggestion
                    else:
```

ndcube/ndcube_sequence.py
```diff
@@ -85,12 +85,21 @@ def cube_like_world_axis_physical_types(self):
...
+                    result.data = [cube[item[1:]] for cube in result.data[item[0]]]
```
DanRyanIrish, line 101:
```suggestion
                        result.data = [cube[item[1:]] for cube in result.data[item[0]]]
```

ndcube/ndcube_sequence.py
```diff
@@ -85,12 +85,25 @@ def cube_like_world_axis_physical_types(self):
...
+            else:
+                if isinstance(item[0], numbers.Integral):
+                    result = result.data[item[0]][item[1:]]
+                else:
+                    item0_is_length1_slice = False
+                    if isinstance(item[0], slice):
+                        start = 0 if item[0].start is None else item[0].start
+                        stop = len(self.data) if item[0].stop is None else item[0].stop
+                        if stop - start == 1:
+                        item0_is_length1_slice = True
```
DanRyanIrish, line 101:
```suggestion
                            item0_is_length1_slice = True
```

ndcube/ndcube_sequence.py
```diff
@@ -85,12 +85,25 @@ def cube_like_world_axis_physical_types(self):
...
+                if item0_is_length1_slice:
```
DanRyanIrish, line 102:
```suggestion
                    if item0_is_length1_slice:
```

ndcube/ndcube_sequence.py
```diff
@@ -85,12 +85,25 @@ def cube_like_world_axis_physical_types(self):
...
+                    result.data = [result.data[start][item[1:]]]
```
DanRyanIrish, line 103:
```suggestion
                        result.data = [result.data[start][item[1:]]]
```

DanRyanIrish:
Hi @hayesla.  I got this PR working.  Hooray!!!

Before merging, could I ask you to squash all the commits?  If you aren't sure how to do that we can walk through it together.  Making this a 1 commit PR will make cherry-picking it for the release of 1.3.1 much easier.

DanRyanIrish:
Thanks for this successful PR @hayesla!!

hayesla:
whoop whoop first NDCube PR 👯

PR #266, opened 2020-04-13
Support more CTYPES.

### LINKED ISSUE #226: Support more ctypes?
Hi,

Thank you for providing this fantastic package. Do you have any plan to add support of more ctypes? Currently, I have found this package does not support the following ctypes: "CRLT-CEA", "CRLN-CEA" and "HGLT-CEA"/"HGLN-CEA", as well as "HPRZ".

The "CEA" ctypes are quite useful, as they are widely used in e.g., the HMI vector magnetic field data, which are usually used to perform non-linear force free field extrapolations and forced field extrapolations.

Thank you.

DanRyanIrish:
### Description

This PR adds more mappings from FITS CTYPES to IVOA words.

Fixes #226


DanRyanIrish:
@Cadair, any objections or suggestions for including these CTYPE mappings?

DanRyanIrish:
@PyDL, here is a PR to add more CTYPES to ndcube.  Any objections to these?

Cadair:
We should in general be adding these to astropy: https://github.com/astropy/astropy/blob/master/astropy/wcs/wcsapi/fitswcs.py#L31

Certainly the obvious lon/lat ones. I need to take a look at the spec on the words a little more.

DanRyanIrish:
> We should in general be adding these to astropy: https://github.com/astropy/astropy/blob/master/astropy/wcs/wcsapi/fitswcs.py#L31

Oh right!  I didn't know about that file.  ndcube 1.x doesn't inherit those from astropy, though right?  That's an APE14-related changed?  I was going to add this PR to the 1.3.1 release, but if it's all going to be done in astropy WCS, maybe it's better to just wait until 2.0.

> Certainly the obvious lon/lat ones. I need to take a look at the spec on the words a little more.

Yeh, this was best interpretation of the spec.  Maybe there is a more correct way to do it though.

ndcube/utils/wcs.py
```diff
@@ -32,8 +32,20 @@ def inv(self):
...
+    "DSUN": "custom:pos.distance.sunToObserver",
+    "HGLT": "custom:pos.heliographic.stonyhurst.lon",
+    "CRLT": "custom:pos.heliographic.carrington.lat",
+    "HGLN": "custom:pos.heliographic.stonyhurst.lat",
+    "CRLN": "custom:pos.heliographic.carrington.lon",
+    "HECR": "custom:pos.distance.fromSunCenter",
+    "HECR": "custom:pos.distance.fromSunSurface",
+    "SOLX": "custom.pos.distance.heliocentric.west",
+    "SOLY": "custom.pos.distance.heliocentric.north",
+    "SOLZ": "custom.pos.distance.heliocentric.sunToObserver",
```
DanRyanIrish, line 44: Should this be ```custom.pos.distance.heliocentric.alt```?

Same question for ```"HPRZ"```.

ndcube/utils/wcs.py
```diff
@@ -32,8 +32,20 @@ def inv(self):
...
     "HPLT": "custom:pos.helioprojective.lat",
     "HPLN": "custom:pos.helioprojective.lon",
+    "HPRZ": "custom:pos.distance.helioprojective.sunToObserver",
```
DanRyanIrish, line 47: Does above comments on ```"SOLZ"``` mean ```"HPRZ"``` should be ```"custom:pos.helioprojective.z"```?  Perhaps ```"custom:pos.helioprojective.alt"```?

ndcube/utils/wcs.py
DanRyanIrish: In fact, ```"SOLX"```, ```"SOLY"```, ```"SOLZ"``` should probably be ```"custom:pos.heliocentric.x"```, ```"custom:pos.heliocentric.y"```, ```"custom:pos.heliocentric.z"```.

Cadair: I *think* we can drop the custom for these?

ndcube/utils/wcs.py
Cadair: I think "z" personally.

ndcube/utils/wcs.py
DanRyanIrish: In the [list of allowed words](https://www.ivoa.net/documents/REC/UCD/UCDlist-20070402.html#_Toc163276347) I only see ```pos.heliocentric```, not ```pos.heliocentric.x/y/z```.  Whereas I do see ```pos.cartesian.x```.

DanRyanIrish:
This PR is superseded by [astropy PR  10965](https://github.com/astropy/astropy/pull/10965)

PR #271, opened 2020-05-07
Implement Extra Coords for NDCube 2.0

Cadair:
Since #207 the extra coords support in master has been broken, this PR aims to restore it by doing a large refactor and making extra coords closer to a WCS object.

The high level objective here is to turn `extra_coords` into an object which can be represented as an APE 14 wcs and then combined with whatever WCS class is passed by the user when constructing the cube.

To do this I have implemented two classes `ExtraCoords` and `LookupTableCoord`, the `ExtraCoords` class will be the type used by `NDCube.extra_coords` (no more dict, or dict like interface). the `ExtraCoords` class uses `LookupTableCoord` to build up a WCS of lookup tables provided by the user.

`ExtraCoords` can operate in one of two modes, either it is given an APE 14 WCS object to wrap, or it is built up of one or more lookup tables. This is to allow use cases like reading extra WCSes from disk (either to provide extra coordinates, or to provide different transforms (like ICRS in STEREO HI data for instance).

The plan is to add a `combined_wcs` property to `NDCube` which uses the `CompoundWCS` wrapper class to present the extra coords object and the base WCS to the user as a single WCS, the main use of this would be to provide `pixel_to_world` operations. This `combined_wcs` will also be used by the plotting to allow plotting of the `extra_coords` with `WCSAxes` (back to where it all started).

Finally, this PR includes the meaty contents of astropy/astropy#9736 to test out and experiment with the utility of those wrapper classes.

My plan is to use gWCS for the implementation of the lookup table WCSes, so ndcube will depend on gwcs after this, which I don't really think is an issue. Also as currently implemented this will depend on the as yet unreleased astropy 4.1. We can decide if we want to relax that later.

**This PR needs sunpy master, astropy 4.1 and gwcs master** to work correctly.

TODO (before merge):

- [x] More tests
- [x] Docs
- [x] Finish and merge dependant PRs: https://github.com/spacetelescope/gwcs/pull/311 https://github.com/spacetelescope/gwcs/pull/325 ~https://github.com/astropy/astropy/pull/10484~ ~https://github.com/sunpy/sunpy/pull/4270~

TODO (after merge):

- [ ] Fix all the wonderful sequence issues.


Cadair:
Extra coords `axis_world_coords` (where do we get array shape) and does `ExtraCoords.wcs.array_shape` change when slicing the `ExtraCoords`.

Cadair:
@DanRyanIrish I think this is ready for a review. I have rigged the devdeps build to pass by installing my custom branch of gwcs.

ndcube/extra_coords/extra_coords.py
```diff
@@ -0,0 +1,324 @@
...
+        new_extra_coords._lookup_tables = tuple(new_lookup_tables)
+        return new_extra_coords
+
+    def _getitem_wcs(self, item):
+        item = sanitize_slices(item, self.wcs.pixel_n_dim)
+
+        # It's valid to slice down the EC such that there is nothing left,
+        # which is not a valid way to slice the WCS
+        if len(item) == self.wcs.pixel_n_dim and all(isinstance(i, Integral) for i in item):
+            return type(self)()
```
DanRyanIrish, line 300: Doesn't slicing all axes with an int leave the coordinate values at the single pixel?  So there should be something left?  Or would these values get transferred to `GlobalCoords`?

ndcube/extra_coords/extra_coords.py
```diff
@@ -0,0 +1,324 @@
...
+
+        subwcs = self.wcs[item]
+
+        new_mapping = []
+        for i, subitem in enumerate(item):
+            if not isinstance(subitem, Integral):
+                new_mapping.append(self.mapping[i])
```
DanRyanIrish, line 307:
```suggestion
        new_mapping = [self.mapping[i]  for i, subitem in enumerate(item) if not isinstance(subitem, Integral)]
```

DanRyanIrish (commented):
A couple initial line comments.  Will look further tomorrow.

A couple general comments too:
1. Do we want the capability to slice `ExtraCoords` by multiple strings?  e.g. returning coupled coordinates like longitude and latitude?
2. From my reading it seems  you can initialize an `ExtraCoords` with a WCS or a lookup table.  What if you initialize with a WCS but then add a coordinate in lookup table form?  Is this possible?  The slicing code seems to aae coords are either wcs or lookup table but not a combination.

ndcube/extra_coords/lookup_table_coord.py
```diff
@@ -0,0 +1,317 @@
...
+    """
+    A wrapper to create a lookup table model on demand.
+    """
+
+    def __init__(self, lookup_table, model_function, mesh=True):
+        self.lookup_table = lookup_table
+        self.model_function = model_function
+        self.mesh = mesh
+
+    def __call__(self):
```
DanRyanIrish, line 298: Should``` mesh``` be incorporated in this function?  Or is it assumed that it is already accounted for because ```model_function``` is always a ``partial``` function that sets ```mesh``` and ```self.mesh``` is simply a record of that value?

ndcube/mixins/plotting.py
```diff
@@ -157,7 +185,23 @@ def _animate_cube(self, wcs, plot_axes=None, axes_coordinates=None,
             for axis_unit, coord_name in zip(axes_units, wcs.world_axis_physical_types):
```
DanRyanIrish, line 185: Isn't `axes_units` entered in array order?  If so, are they reversed relative to `wcs.world_axis_physical_types`?

ndcube/mixins/plotting.py
```diff
@@ -157,7 +185,23 @@ def _animate_cube(self, wcs, plot_axes=None, axes_coordinates=None,
...
                 coord_params[coord_name] = {'format_unit': axis_unit}
 
+        # TODO: `_plot_2D_cube` transposes the array if 'y' is before 'x'.
+        # Investigate if this should be the case here, or if it shouldn't be done at all.
```
DanRyanIrish, line 189: Is this done?

ndcube/tests/test_ndcubesequence.py
```diff
@@ -1,264 +0,0 @@
-import datetime
```
DanRyanIrish, line 1: Why is this file deleted?

DanRyanIrish (commented):
More comments.  Still yet to look through the wcs wrappers.

ndcube/wcs/wrappers/compound_wcs.py
```diff
@@ -0,0 +1,213 @@
...
+            world_arrays_sub = world_arrays[:w.world_n_dim]
+            world_arrays = world_arrays[w.world_n_dim:]
+            pixel_arrays_sub = w.world_to_pixel_values(*world_arrays_sub)
+            if w.pixel_n_dim > 1:
+                pixel_arrays.extend(pixel_arrays_sub)
+            else:
+                pixel_arrays.append(pixel_arrays_sub)
+
+        pixel_arrays = tuple(pixel_arrays)
+        # for i, ix in enumerate(self.mapping.mapping):
```
DanRyanIrish, line 131: Can these commented out lines be removed?

ndcube/extra_coords/extra_coords.py
```diff
@@ -0,0 +1,324 @@
...
+    @wcs.setter
+    def wcs(self, wcs):
+        if self._wcs is not None:
+            raise AttributeError(
+                "Can't set wcs if a WCS has already been specified."
+            )
+
+        if self._lookup_tables:
+            raise AttributeError(
+                "Can't set mapping manually when ExtraCoords is built from lookup tables."
```
Cadair, line 251:
```suggestion
                "Can't set wcs manually when ExtraCoords is built from lookup tables."
```

ndcube/extra_coords/extra_coords.py
Cadair: It would be copied into `GlobalCoords` when that's implemented.

ndcube/extra_coords/lookup_table_coord.py
```diff
@@ -0,0 +1,317 @@
...
+        A list of all the lookup tables contained by this object.
+    frames: `list` of `gwcs.coordinate_frames.CoordinateFrame`
+        A list of all the gwcs coordinate frame objects corresponding to the
+        lookup tables.
+
+    Notes
+    -----
+
+    The constructor to this class expects a single coordinate to be passed,
+    i.e. one coordinate frame (i.e. Celestial), although this could be built
```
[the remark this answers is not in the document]
Cadair: the "single coordinate"

ndcube/extra_coords/lookup_table_coord.py
```diff
@@ -0,0 +1,317 @@
...
+    from multiple lookup tables.
+
+    The implementation of this class, however, allows the representation of
+    many coordinate frames in a single instance. This is primarily to allow the
+    ``&`` operator to work between instances of this class, and therefore build
+    up a multi-dimensional WCS object.
+
+    If you wish to build a multi-dimensional `.LookupTableCoord` the correct
+    approach is to construct two instances and then join them into a new
+    combined instance with the ``&`` operator.
```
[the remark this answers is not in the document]
Cadair: This approach was chosen to simplify the constructor of the class. There is minimal overhead in doing it this way, and to my mind it makes the API a lot easier to reason with.

ndcube/extra_coords/lookup_table_coord.py
```diff
@@ -0,0 +1,317 @@
...
+        return cf.CoordinateFrame(naxes, axes_type, axes_order, unit=unit,
+                                  axes_names=names, name=name, axis_physical_types=physical_types)
+
+    def _from_time(self, lookup_tables, mesh=False, names=None, physical_types=None, **kwargs):
+        if len(lookup_tables) > 1:
+            raise ValueError("Can only parse one time lookup table.")
+
+        time = lookup_tables[0]
+        deltas = (time[1:] - time[0]).to(u.s)
+        deltas = deltas.insert(0, 0)
```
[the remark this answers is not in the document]
Cadair: This works as-is probably because of a quirk in units, will change it to be more explicit.

ndcube/extra_coords/lookup_table_coord.py
Cadair: mesh is passed to the constructor of the class, so that when call is called the information about if mesh is being used is already in the class state.

ndcube/mixins/plotting.py
Cadair: it is not done, it's still an open question. I am not sure it's easy to do in the animator, and it feels weird to have it supported in 2D but not the animator?

ndcube/mixins/plotting.py
Cadair: yes, but the utils function flips them into world order so all the plot methods here have them in WCSAxes order.

ndcube/wcs/wrappers/compound_wcs.py
Cadair: hmm I should probably get back to this.

Cadair:
> Do we want the capability to slice ExtraCoords by multiple strings? e.g. returning coupled coordinates like longitude and latitude?

If you index with a single string it will return all coordinates in a matching lookup coordinate. There isn't a way to split them into bits. I don't know if it's worth adding multiple string slicing? I doubt it would be that hard?

> From my reading it seems you can initialize an ExtraCoords with a WCS or a lookup table. What if you initialize with a WCS but then add a coordinate in lookup table form? Is this possible? The slicing code seems to aae coords are either wcs or lookup table but not a combination.

It's not possible, if we ever decided that we really need it we could add it, but it would add a decent amount of complexity.

ndcube/extra_coords/extra_coords.py
kakirastern: Am working on this now

ndcube/extra_coords/extra_coords.py
```diff
@@ -0,0 +1,321 @@
...
+       The mapping between the array dimensions and pixel dimensions in the
+       extra coords object. This is an iterable of ``(array_dimension, pixel_dimension)`` pairs
+       of length equal to the number of pixel dimensions in the extra coords.
+
+    """
+
+    @abc.abstractmethod
+    def __init__(self,
+                 *,
+                 wcs: BaseLowLevelWCS = None,
```
DanRyanIrish, line 40: Should this also accept a `HighLevelWCS`?

ndcube/mixins/plotting.py
DanRyanIrish: I think the more desired behavior is that the data is transposed when `'y'` is before `'x'`.  So the fact that it is difficult/not currently possible with the animator isn't a reason for it not to be done here.  I'd consider that a bug in animator rather than a design choice.  That's my feeling, anyway.

ndcube/ndcube.py
```diff
@@ -6,13 +6,14 @@
 import astropy.nddata
 import astropy.units as u
 import numpy as np
-import sunpy.coordinates
-from astropy.wcs.wcsapi import BaseLowLevelWCS
+import sunpy.coordinates  # pylint: disable=unused-import  # NOQA
```
DanRyanIrish, line 9: This import isn't used.  And if it was it would make `ndcube` dependent on `sunpy` again.  Aren't we trying to avoid that?

ndcube/tests/test_ndcubesequence.py
DanRyanIrish: Is the idea that when I port the new fixture-based sequence tests, they'll slot in here rather than still remaining deleted?  Maybe best to not delete this file as part of this PR and simply have it fixed by another PR?

ndcube/wcs/wrappers/compound_wcs.py
DanRyanIrish: This issues is still open.

ndcube/wcs/wrappers/compound_wcs.py
```diff
@@ -0,0 +1,213 @@
...
+    def inverse(self):
+        mapping = tuple(self.mapping.index(idx)
+                        for idx in range(self.n_inputs))
+        return type(self)(mapping)
+
+    def __repr__(self):
+        return f'<Mapping({self.mapping})>'
+
+
+class CompoundLowLevelWCS(BaseWCSWrapper):
```
DanRyanIrish, line 46: Can this be used to solve the `NDCubeSequence` plotting/collapsing a sequence into a single cube?

ndcube/ndcube.py
Cadair: I think we should probably keep this, maybe with a try block when we actually remove the sunpy dependency (which isn't done yet).

ndcube/extra_coords/extra_coords.py
Cadair: Arguably it could, but I am not sure it's needed. (It would just strip it down to low level anyway).

ndcube/mixins/plotting.py
Cadair: In that case we should probably raise a warning here until we fix it or something.

ndcube/wcs/wrappers/compound_wcs.py
Cadair: Indeed.

ndcube/wcs/wrappers/compound_wcs.py
Cadair: This would allow the one FITS WCS + lookup table to work (which is sorta how we use it in Extra Coords) but it wouldn't enable the lookup table of WCSes to work.

ndcube/mixins/plotting.py
DanRyanIrish: Sounds good.

ndcube/extra_coords/extra_coords.py
DanRyanIrish: :+1:

ndcube/ndcube.py
DanRyanIrish: Sounds good.

docs/ndcube.rst
```diff
@@ -1,3 +1,6 @@
+.. doctest-skip-all::
+    all
```
DanRyanIrish, line 2: Should this still be skipping all the doctests??

docs/ndcubesequence.rst
```diff
@@ -1,3 +1,6 @@
+.. doctest-skip-all::
+    all
+
```
DanRyanIrish, line 3: Should this still be skipping all the doctests?

ndcube/mixins/plotting.py
DanRyanIrish: A warning still needs to be raised here I think?

ndcube/ndcube.py
DanRyanIrish: This still needs a ```try...except...``` on the sunpy import?

docs/ndcube.rst
Cadair: I am skipping all the doctests as we haven't rewritten the docs for the new APIs in NDCube 2 yet.

DanRyanIrish (approved):
Approved pending tests passing.

PR #279, opened 2020-07-02
Add new axis_world_coord_values method to NDCube.

### LINKED ISSUE #276: NDCube.axis_world_coords should return coords for all world types associated with valid pixel axes
@Cadair, this tries to capture a little of what we discussed the other day as well as some thoughts I've had since.

```NDCube.axis_world_coords()``` currently only returns one coord type per axis.  This means that if two world axes are dependent and one of the pixel axes has been sliced away, e.g. ```x``` and ```y``` correspond to ```lon``` and ```lat``` but the x-axis has been sliced away, only ```lat``` is returned.

```NDCube.axis_world_coords``` should be changed so that is returns all world coords that correspond to valid pixel axes.  In the above example, ```lon``` and ```lat``` should be returned as the ```y```-axis corresponds to ```lat``` and ```lon```, even when the ```x```-axis is degenerate.

Furthermore, ```NDCube.axis_world_coords``` currently returns a ```tuple``` of coords, with users expected to decipher which result corresponds to which type by looking at the unit.  But this is not always unique, e.g. ```lon``` and ```lat``` are both in ```degrees```.  Therefore, ```NDCube.axis_world_coords``` should return a ```dict``` where the keys give the physical type of each returned coord.

If users supply world axis types as strings, only those coord types should be returned.  If users supply axis numbers, only coords corresponding to those axes should be returned.

Except in the case where a single axis number is input, this design of ```NDCube.axis_world_coords```means that it is not obvious from the output which coords correspond to which pixel dimensions.  This could be worked out from the ```wcs.axis_correlation_matrix```.  But one could imagine that it would be much easier for there to be a convenience method(s) that returned the pixel axes for each world type and/or the world types for each pixel axis.  Thus a quick visual comparison would make the correspondence immediately obvious.  See Issue #220 (and supplementary issue #240) for more discussion on such a method.

### PR DESCRIPTION
**Pull request by DanRyanIrish**
### Description

This PR adds a new method to ```NDCube```: ```axis_world_coords_values```.  It differs from the current implementation of ```axis_world_coords``` by returning all coords for all the user supplied axes, including dependent axes.  It will differ from future implementation of ```axis_world_coords``` by returning quantity arrays rather than higher level objects, e.g. skycoords.

Fixes #276

### CODE REVIEW

**Comment by DanRyanIrish** [comment]
@Cadair:  Here is my attempt.

File: ndcube/ndcube.py [hunk 1]
```diff
@@ -362,6 +362,93 @@ def axis_world_coords(self, *axes, edges=False):
...
+        # Reduce duplication across independent dimensions for each coord.
+        for i, axis_coord in enumerate(axes_coords):
+            slices = np.array([slice(None)] * self.wcs.world_n_dim)
+            slices[np.invert(self.wcs.axis_correlation_matrix[i])] = 0
+            axis_coords[i] = axis_coord[tuple(slices)].T
+
+        world_axis_physical_types = self.wcs.world_axis_physical_types
+        # If user has supplied axes, extract only the
+        # world coords that correspond to those axes.
+        if axes != ():
```
DanRyanIrish, line 422: Replace with
```python
if axes:
```

ndcube/ndcube.py
```diff
@@ -362,6 +362,93 @@ def axis_world_coords(self, *axes, edges=False):
...
+            Real world coords for axes requested by user.
+            Returned in same order as axis_names.
+
+        Example
+        -------
+        >>> NDCube.all_world_coords_values(('lat', 'lon')) # doctest: +SKIP
+        >>> NDCube.all_world_coords_values(2) # doctest: +SKIP
+
+        """
+
```
DanRyanIrish, line 399:
```suggestion
```

ndcube/ndcube.py
```diff
@@ -362,6 +362,93 @@ def axis_world_coords(self, *axes, edges=False):
...
+        if edges:
+            wcs_shape = tuple(np.array(wcs_shape) + 1)
+            pixel_inputs = np.meshgrid(*[np.arange(i) - 0.5 for i in wcs_shape], indexing='ij')
+        else:
+            pixel_inputs = np.meshgrid(*[np.arange(i) for i in wcs_shape], indexing='ij')
+
+        # Get world coords for all axes and all pixels.
+        axes_coords = self.wcs.pixel_to_world(*pixel_inputs)
+
+        # Reduce duplication across independent dimensions for each coord.
```
DanRyanIrish, line 413:
```suggestion
        # Reduce duplication across independent dimensions for each coord
        # and transpose to make dimensions mimic numpy array order rather than WCS order.
```

ndcube/ndcube.py
```diff
@@ -620,3 +707,51 @@ def __init__(self, data, wcs, uncertainty=None, mask=None, meta=None,
...
+            # Get world axes that correspond to the WCS pixel axis.
+            axes_numbers += list(get_world_axes_from_wcs_pixel_axis(wcs, axis))
+        elif isinstance(axis, str):
+            axes_numbers += get_wcs_world_axes_from_axis_name(wcs, axis)
+        else:
+            raise TypeError(f"Unrecognized axis type: {axis, type(axis)}. "
+                            "Must be of type (numbers.Integral, str)")
+    return np.array(list(set(axes_numbers)), dtype=int)
+
+
```
DanRyanIrish, line 729:
```suggestion
```

ndcube/ndcube.py
```diff
@@ -620,3 +707,51 @@ def __init__(self, data, wcs, uncertainty=None, mask=None, meta=None,
...
+def get_world_axes_from_wcs_pixel_axis(wcs, pixel_axis):
+    return np.arange(wcs.world_n_dim)[wcs.axis_correlation_matrix[:, pixel_axis]]
+
+
+def get_wcs_world_axes_from_axis_name(wcs, axis_name):
+    # Define required attributes from wcs.
+    world_axis_physical_types = wcs.world_axis_physical_types
+    axis_correlation_matrix = wcs.axis_correlation_matrix
+    world_n_dim = wcs.world_n_dim
+
```
DanRyanIrish, line 747:
```suggestion
```

ndcube/ndcube.py
```diff
@@ -620,3 +707,51 @@ def __init__(self, data, wcs, uncertainty=None, mask=None, meta=None,
...
+    # Find world axis described by axis_name.
+    widx = np.where(world_axis_physical_types == axis_name)[0]
```
DanRyanIrish, line 749:
```suggestion
    widx = np.where(wcs.world_axis_physical_types == axis_name)[0]
```

ndcube/ndcube.py
```diff
@@ -620,3 +707,51 @@ def __init__(self, data, wcs, uncertainty=None, mask=None, meta=None,
...
+    # If axis name as given does not correspond to a physical axis,
+    # check if it is a substring of any physical types.
+    if len(widx) == 0:
+        widx = [axis_name in physical_type for physical_type in world_axis_physical_types]
```
DanRyanIrish, line 753:
```suggestion
        widx = [axis_name in physical_type for physical_type in wcs.world_axis_physical_types]
```

ndcube/ndcube.py
```diff
@@ -620,3 +707,51 @@ def __init__(self, data, wcs, uncertainty=None, mask=None, meta=None,
...
+        widx = np.arange(world_n_dim)[widx]
```
DanRyanIrish, line 754:
```suggestion
        widx = np.arange(wcs.world_n_dim)[widx]
```

ndcube/ndcube.py
Cadair: surely you mean `if axes`?

ndcube/ndcube.py
```diff
@@ -620,3 +707,51 @@ def __init__(self, data, wcs, uncertainty=None, mask=None, meta=None,
...
+
```
DanRyanIrish, line 755:
```suggestion
```

ndcube/ndcube.py
DanRyanIrish: I do indeed!

ndcube/ndcube.py
```diff
@@ -362,6 +362,93 @@ def axis_world_coords(self, *axes, edges=False):
...
+        """
+        # Create meshgrid of all pixel coordinates.
+        # If user, wants edges, set pixel values to pixel edges.
+        # Else make pixel centers.
+        wcs_shape = self.data.shape[::-1]
+        if edges:
+            wcs_shape = tuple(np.array(wcs_shape) + 1)
+            pixel_inputs = np.meshgrid(*[np.arange(i) - 0.5 for i in wcs_shape], indexing='ij')
+        else:
+            pixel_inputs = np.meshgrid(*[np.arange(i) for i in wcs_shape], indexing='ij')
```
Cadair, line 407: This needs some thought, at the moment I am pretty sure this is memory dense.

DanRyanIrish: OK.  It could well be.  I didn't really understand how it could not be.  If you have a way to ensure memory efficiency is maximized that would be great.  But would it be a case of changing this one line?  Or would the whole approach need to be refactored?

ndcube/ndcube.py
```diff
@@ -362,6 +362,93 @@ def axis_world_coords(self, *axes, edges=False):
...
+            slices[np.invert(self.wcs.axis_correlation_matrix[i])] = 0
+            axis_coords[i] = axis_coord[tuple(slices)].T
+
+        world_axis_physical_types = self.wcs.world_axis_physical_types
+        # If user has supplied axes, extract only the
+        # world coords that correspond to those axes.
+        if axes:
+            # Convert input axes to WCS world axis indices.
+            if not isinstance(axes, tuple):
+                axes = (axes,)
```
Cadair, line 425: axes will always be a tuple if the function sig is `*axes`.
```suggestion
```

ndcube/ndcube.py
```diff
@@ -362,6 +362,93 @@ def axis_world_coords(self, *axes, edges=False):
...
+        else:
+            pixel_inputs = np.meshgrid(*[np.arange(i) for i in wcs_shape], indexing='ij')
+
+        # Get world coords for all axes and all pixels.
+        axes_coords = self.wcs.pixel_to_world(*pixel_inputs)
+
+        # Reduce duplication across independent dimensions for each coord
+        # and transpose to make dimensions mimic numpy array order rather than WCS order.
+        for i, axis_coord in enumerate(axes_coords):
+            slices = np.array([slice(None)] * self.wcs.world_n_dim)
```
Cadair, line 415: why?
```suggestion
            slices = [slice(None)] * self.wcs.world_n_dim
```

ndcube/ndcube.py
```diff
@@ -362,6 +362,93 @@ def axis_world_coords(self, *axes, edges=False):
...
+            axis_coords[i] = axis_coord[tuple(slices)].T
+
+        world_axis_physical_types = self.wcs.world_axis_physical_types
+        # If user has supplied axes, extract only the
+        # world coords that correspond to those axes.
+        if axes:
+            # Convert input axes to WCS world axis indices.
+            if not isinstance(axes, tuple):
+                axes = (axes,)
+            world_indices = []
```
Cadair, line 426: why not make this a set up front?
```suggestion
            world_indices = set()
```

ndcube/ndcube.py
```diff
@@ -362,6 +362,93 @@ def axis_world_coords(self, *axes, edges=False):
...
+            for axis in axes:
+                if isinstance(axis, numbers.Integral):
+                    # If axis is int, it is a numpy order array axis.
+                    # Convert to pixel axis in WCS order.
+                    axis = array_axis_to_wcs_pixel_axis(axis, wcs.pixel_n_dim)
+                    # Get WCS world axis indices that correspond to the WCS pixel axis
+                    # and add to list of indices of WCS world axes whose coords will be returned.
+                    world_indices += list(get_world_axes_from_wcs_pixel_axis(wcs, axis))
```
Cadair, line 434: If it's a set:
```suggestion
                    world_indices.update(get_world_axes_from_wcs_pixel_axis(wcs, axis))
```

ndcube/ndcube.py
```diff
@@ -362,6 +362,93 @@ def axis_world_coords(self, *axes, edges=False):
...
+                elif isinstance(axis, str):
+                    # If axis is str, it is a physical type or substring of a physical type.
+                    # Use world_axis_physical_types to infer the its WCS world index.
+                    world_indices += get_wcs_world_axes_from_axis_name(wcs, axis)
```
Cadair, line 438:
```suggestion
                    world_indices.update(get_wcs_world_axes_from_axis_name(wcs, axis))
```

ndcube/ndcube.py
```diff
@@ -362,6 +362,93 @@ def axis_world_coords(self, *axes, edges=False):
...
+                else:
+                    raise TypeError(f"Unrecognized axis type: {axis, type(axis)}. "
+                                    "Must be of type (numbers.Integral, str)")
+            # Remove duplicate world indices then use to extract the desired coord values
+            # and corresponding physical types.
+            world_indices = np.array(list(set(world_indices)), dtype=int)
```
Cadair, line 444:
```suggestion
            world_indices = np.array(list(world_indices), dtype=int)
```

ndcube/ndcube.py
DanRyanIrish: Because indexing a list with an array and setting the values doesn't always work.  At least it's thrown errors for me which I was able to fix by setting the list to an array.

ndcube/ndcube.py
DanRyanIrish: Hadn't thought of that!

DanRyanIrish:
> I think before we pour too much more time into this we need to look at the memory ramifications of this implementation.

Agreed.

> Also I am still sceptical about the `_values` method needing to return the physical types. I can't see a use case where that is useful to a library function, it seems to me a library would know what axes they were asking for already and don't need to be told on the way out. I don't really see this function being useful for users, they should be using `axis_world_coords()`, not the `values()` variant.

I guess I was hoping to allow users to access it until axis_world_coords is refactored for 2.0.  Because as it stands, the fact that ```axis_world_coords``` does not return longitude if you slice away your x dimension is such a back design choice it's actually a bug.  That design choice was on me so I don't mind bashing it.  Alternatively we could remove the axis names from the output of  ```axis_world_coords_values``` and have ```axis_world_coords``` return the axis names + the output of  ```axis_world_coords_values```.  However, this would be a signifiant breaking change.  Basically we are in a bind because the behaviour of ```axis_world_coords``` is flat out wrong and needs to be fixed for asap.

ndcube/ndcube.py
DanRyanIrish: I compared the outputs of ```np.meshgrid``` using the ```sparse``` kwargs.  For ```wcs_shape=(32, 1024, 48, 1)``` and the same number of pixel and world dimensions, I got the following sizes in bytes:
| ```sparse``` | Meshgrid List |  Meshgrid[0] |  Meshgrid[1] |  Meshgrid[2] |  Meshgrid[3] |
|--------------|:--------------:|-------------:|-------------:|-------------:|--------------:|
| ```False```   |  104                  | 12583056     | 12583056    | 12583056     | 12583056      |
| ```True```    |   104                 | 400                | 8336            | 528                | 152                 |

Meanwhile the sizes for the outputs of ```pixel_to_world``` was the same independent of whether ```sparse``` was ```True``` or ```False```.  The checked the sizes of the output ```Quantities``` and the ```Quantity.values```.  They were:
|                 | coord0 | coord1 | coord2 | coord3 |
|----------|:-------:|-------:|--------:|-------:|
| Quantity  | 176       | 176      | 176       | 64       |
| Values     | 144       | 144     | 144       | 12583056|

For clarity, these were the sizes before redundant dimensions were removed.  And coord3 was a ```Time``` object which I imagine explains the much large size of its ```values```.

Is this sufficient to assuage memory concerns or is there are way to make the inputs or outputs sparser?  Or is the concern that even if the input is sparse, it must be made dense as part of the calculations and so this doesn't save us much?

An alternative approach would be to perform the transformations separately for sets of dependent axes.  In the above example, axes ```0```, ```1```, and ```3``` are dependent, so we would call ```pixel_to_world``` twice, once for a pixel grid of size ```(32, 1024, 1, 1)``` and once for a grid of ```(1, 1, 48, 1)```.  However, this seems like it would reintroduce the code complexity of the old version of ```axes_world_coords```.  Since as well as sorting out the user input, breaking up the pixel grid was the main source of code complexity.

DanRyanIrish:
ping @Cadair.  ```NDCube.axis_world_coord_values``` now returns a named tuple and I believe all issues of controversy have been addressed.  I believe this is now ready to merge.

ndcube/ndcube.py
```diff
@@ -262,105 +263,111 @@ def axis_world_coords(self, *axes, edges=False):
...
+            axes_coords = np.array(axes_coords)[world_indices]
+            world_axis_physical_types = tuple(np.array(world_axis_physical_types)[world_indices])
+
+        # Return in array order.
+        # First replace characters in physical types forbidden for namedtuple identifiers.
+        identifiers = []
+        for physical_type in world_axis_physical_types[::-1]:
+            identifier = physical_type.replace(":", "_")
+            identifier = identifier.replace(".", "_")
+            identifier = identifier.replace("-", "__")
```
Cadair, line 367: Why replace this one with two?

ndcube/ndcube.py
```diff
@@ -262,105 +262,105 @@ def axis_world_coords(self, *axes, edges=False):
...
+            High level object giving the real world coords for the axes requested by user.
+            For example, SkyCoords.
 
         Example
         -------
         >>> NDCube.all_world_coords(('lat', 'lon')) # doctest: +SKIP
         >>> NDCube.all_world_coords(2) # doctest: +SKIP
+
+        """
+        raise NotImplementedError()
```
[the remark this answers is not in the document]
DanRyanIrish: This is not meant to be a deliberate policy.  It simply stems from the fact that I renamed ```axis_world_coords``` to ```axis_world_coord_values``` but wanted and placeholder for ```axis_world_coords``` for the future version that returns coordinate objects.  Before 2.0 this should be written, if you haven't written it already...

ndcube/ndcube.py
DanRyanIrish: Good question.  The idea was that ```_``` would denote an IVOA word separator.  But ```-``` is used to show a range, e.g. ```100-200nm```, not separate words.  So I thought it would be a good idea for string manipulations to not overload the meaning of ```_```.  Perhaps there is a better replacement for ```-``` than ```__```?

ndcube/ndcube.py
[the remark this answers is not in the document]
Cadair: ok, makes sense on the 2.0 branch :+1:

ndcube/ndcube.py
Cadair: not that I can think of :+1:

PR #280, opened 2020-07-03
Utils refactor

DanRyanIrish:
### Description

This PR does two things:
  1. Removes unused util functions.
  2. Refactors the functions that convert between data and WCS axes.  In doing some new conversions are created.

The second a point is needed since the introduction of APE14 for astropy WCS.  Among other things, this draws a much more explicit distinction between array axes, WCS pixel axes (the reverse order of array axes) and WCS world axes which correspond to the world axis physical types.  This is because there is not always a 1-to-1 mapping from pixel axes to world axes/physical types.  Previously, ```ndcube``` code only converted between array (data) axes and WCS axes which is now ambiguous, or at the very least confusing because it wrapped the array-to-pixel (WCS order) mapping with the pixel-to-physical type mapping.  These new functions make this clearer and more explicit.

Nine new functions have been created, a couple of which are just rebranded old functions to match the new array-pixel-world naming convention and minimize the complexity of required inputs.  But the rest are either new or proper refactors.  They are described below and are grouped in 3 categories: conversions between array and WCS pixel axes, conversions between WCS pixel axes, world axes, and physical types, and dependent axes functions.

#### Array/WCS pixel Conversions
* ```reflect_axis_index()```: Converts between array (data) axes and WCS pixel axes.  (Effectively reflecting the axis number through about the mean of the number of array/pixel dimensions.  This function works in either direction as the mapping is symmetric.  It is also vectorized so an array of array or pixel axes can be input at one time.

#### WCS pixel/world/physical types Conversions
These functions convert between pixel axis indices (defined in the WCS ordering convention), world axes (which are the indices of the relevant physical types in ```wcs.world_axis_physical_types```), and physical types.  As different pixel axes will have different numbers of world axes/physical types corresponding to them (and vice versa), these functions are not vectorized.  The fact that more then one axis can be returned from a single input axes is reflected by the fact that the input is axis type in the function name is singular, while the output type is plural.
* ```pixel_axis_to_world_axes```: Gives the world axes that are dependent on the supplied pixel axis.
* ```world_axis_to_pixel_axes```: Gives the pixel axes that are dependent on the supplied world axis.
* ```pixel_axis_to_physical_types```: Gives the physical types that are dependent on the supplied pixel axis.
* ```physical_type_to_pixel_axes```: Gives the pixel axes that are dependent on the supplied physical type.
* ```physical_type_to_world_axis```: Gives the world axes that are dependent on the supplied physical type.  This is called by ```physical_type_to_pixel_axes``` and is not trivial because it allows unique substrings to be supplied.

The mapping from ```world_axis``` to ```physical_type``` is trivial and so not included.

#### Dependent Axes
* ```get_dependent_pixel_axes```: Given a pixel axis index (WCS ordering), return all WCS pixel axis indices dependent on it, including itself.
* ```get_dependent_array_axes```: Given an array axis index (numpy ordering), return all array axis indices dependent on it, including itself.
* ```get_dependent_world_axes```: Given a world axis index (WCS ordering), return all WCS world axis indices dependent on it, including itself.

The corresponding ```get_dependent_physical_types``` is not supplied but could be.  However, this should simply be 
```python
def get_dependent_physical_types(physical_type, wcs):
    world_axis = physical_type_to_world_axis(physical_type, wcs.world_axis_physical_types)
    dependent_world_axes = get_dependent_world_axes(world_axis, wcs.axis_correlation_matrix)
    dependent_physical_types = np.array(wcs.world_axis_physical_types)[dependent_world_axes]
    return dependent_physical_types
```

Fixes #<Issue Number>

### DIFF
--- a/changelog/280.breaking.rst
+++ b/changelog/280.breaking.rst
@@ -1,1 +1,1 @@
-Remove unused util functions and the ndcube WCS class.  Refactor util functions for converting between between data and WCS indices to reflect the APE14 nomenclature that distinguishes between array, pixel and world axes. 
+Remove unused util functions and the ndcube WCS class.  Refactor util functions for converting between between data and WCS indices to reflect the APE14 nomenclature that distinguishes between array, pixel and world axes.
--- a/changelog/280.breaking.rst
+++ b/changelog/280.breaking.rst
@@ -1,1 +1,1 @@
-Refactor utils.
+Remove unused util functions and the ndcube WCS class.  Refactor util functions for converting between between data and WCS indices to reflect the APE14 nomenclature that distinguishes between array, pixel and world axes. 
--- a/changelog/280.breaking.rst
+++ b/changelog/280.breaking.rst
@@ -1,0 +1,1 @@
+Refactor utils.
--- a/ndcube/utils/cube.py
+++ b/ndcube/utils/cube.py
@@ -159,7 +159,7 @@ def _format_input_extra_coords_to_extra_coords_wcs_axis(extra_coords, pixel_keep
             raise ValueError(coord_0_format_error.format(coord))
         # Check coord axis number is valid and convert to a WCS-order axis number.
         if coord[1] is None:
-            new_coord_axis = None
+            wcs_coord_axis = None
         else:
             if isinstance(coord[1], numbers.Integral):
                  wcs_coord_axis = wcs_utils.reflect_axis_index(np.array([coord[1]]), naxes)[0]
--- a/ndcube/utils/cube.py
+++ b/ndcube/utils/cube.py
@@ -7,7 +7,7 @@ import numbers
 import numpy as np
 from astropy.units import Quantity
 
-from ndcube.utils.wcs import _pixel_keep, get_dependent_wcs_axes
+import ndcube.utils.wcs as wcs_utils
 
 __all__ = [
     '_pixel_centers_or_edges',
@@ -147,7 +147,7 @@ def _format_input_extra_coords_to_extra_coords_wcs_axis(extra_coords, pixel_keep
     coord_0_format_error = ("1st element of extra coordinate tuple must be a "
                             "string giving the coordinate's name.")
     coord_1_format_error = ("2nd element of extra coordinate tuple must be None "
-                            "or an int giving the data axis "
+                            "or an int or tuple of int giving the data axis "
                             "to which the coordinate corresponds.")
     coord_len_error = ("extra coord ({0}) must have same length as data axis "
                        "to which it is assigned: coord length, {1} != data axis length, {2}")
@@ -157,14 +157,18 @@ def _format_input_extra_coords_to_extra_coords_wcs_axis(extra_coords, pixel_keep
             raise ValueError(coord_format_error.format(coord))
         if not isinstance(coord[0], str):
             raise ValueError(coord_0_format_error.format(coord))
-        if coord[1] is not None and not isinstance(coord[1], numbers.Integral):
-            raise ValueError(coord_1_format_error)
+        # Check coord axis number is valid and convert to a WCS-order axis number.
+        if coord[1] is None:
+            new_coord_axis = None
+        else:
+            if isinstance(coord[1], numbers.Integral):
+                 wcs_coord_axis = wcs_utils.reflect_axis_index(np.array([coord[1]]), naxes)[0]
+            elif hasattr(coord[1], "__len__") and all([isinstance(c, numbers.Integral) for c in coord[1]]):
+                wcs_coord_axis = tuple(wcs_utils.reflect_axis_index(np.array(coord[1]), naxes))
+            else:
+                raise ValueError(coord_1_format_error)
 
-        # Determine wcs axis corresponding to data axis of coord
-
-        extra_coords_wcs_axis[coord[0]] = {
-            "wcs axis": data_axis_to_wcs_ape14(coord[1], pixel_keep, naxes),
-            "value": coord[2]}
+        extra_coords_wcs_axis[coord[0]] = {"wcs axis": wcs_coord_axis, "value": coord[2]}
     return extra_coords_wcs_axis
 
 
--- a/ndcube/utils/cube.py
+++ b/ndcube/utils/cube.py
@@ -17,7 +17,7 @@ __all__ = [
     'wcs_axis_to_data_ape14',
     'unique_data_axis']
 
-# Deprecated in favor of utils.wcs.reflect_axis_index
+# Deprecated in favor of utils.wcs.convert_between_array_and_pixel_axes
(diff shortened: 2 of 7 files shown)


ndcube/utils/wcs.py
```diff
@@ -300,3 +152,257 @@ def _pixel_keep(wcs_object):
...
+    axis_correlation_matrix: `numpy.ndarray` of `bool`
+        2D boolean correlation matrix defining the dependence between the pixel and world axes.
+        Format same as `astropy.wcs.BaseLowLevelWCS.axis_correlation_matrix.
+
+    Returns
+    -------
+    pixel_axes: `numpy.ndarray`
+        The pixel axis indices corresponding to the world axis.
+    """
+    return np.arange(axis_correlation_matrix.shape[1])[axis_correlation_matrix[world_axis]]
```
Cadair, line 244: why not make this one function with an option to select direction, that's a lot of LOC you are saving :wink:

DanRyanIrish: Doesn't that go against what you said about my functions allowing too many options? ;P  And I would actually have to add a couple lines of code and make this API a little more confusing for the sake of saving on a docstring.

ndcube/utils/cube.py
```diff
@@ -95,27 +78,7 @@ def data_axis_to_wcs_ape14(data_axis, pixel_keep, naxes, old_order=False):
...
-                len(missing_axes)))
-        if missing_axes[wcs_axis]:
-            result = None
-        else:
-            data_ordered_wcs_axis = len(missing_axes) - wcs_axis - 1
-            result = data_ordered_wcs_axis - sum(missing_axes[::-1][:data_ordered_wcs_axis])
-    return result
-
-
+# Deprecated in favor of utils.wcs.reflect_axis_index
```
Cadair, line 81: :fire:

ndcube/utils/cube.py
```diff
@@ -265,7 +206,7 @@ def convert_extra_coords_dict_to_input_format(extra_coords, pixel_keep, naxes):
         result.append((name, axis, extra_coords[name]["value"]))
     return result
 
-
+# Deprecated in favor of utils.wcs.physical_type_to_pixel_axes + util.wcs.reflect_axis_index
 def get_axis_number_from_axis_name(axis_name, world_axis_physical_types):
```
Cadair, line 210: :fire:

DanRyanIrish:
@Cadair All tests are now passing.  Are you happy to approve this?

ndcube/utils/wcs.py
```diff
@@ -300,3 +158,270 @@ def _pixel_keep(wcs_object):
     if hasattr(wcs_object, "_pixel_keep"):
         return wcs_object._pixel_keep
     return np.arange(wcs_object.pixel_n_dim)
+
+
+def reflect_axis_index(axis, naxes):
+    """Reflects axis index about center of number of axes.
```
Cadair, line 164: I find this description confusing. Maybe rename it to something other than reflect?

ndcube/utils/wcs.py
```diff
@@ -300,3 +158,270 @@ def _pixel_keep(wcs_object):
...
+
+    This is used to convert between array axes in numpy order and pixel axes in WCS order.
+    Works in both directions.
+
+    Parameters
+    ----------
+    axis: `numpy.ndarray` of `int`
+        The axis number(s) before reflection.
+
```
Cadair, line 173:
```suggestion
```

ndcube/utils/wcs.py
```diff
@@ -300,3 +158,270 @@ def _pixel_keep(wcs_object):
...
+                "Input does not uniquely correspond to a physical type."
+                f" Expected unique substring of one of {world_axis_physical_types}."
+                f"  Got: {physical_type}")
+    # Return axes with duplicates removed.
+    return widx[0]
+
+
+def get_dependent_pixel_axes(pixel_axis, axis_correlation_matrix):
+    """
+    Given a WCS pixel axis index, return indices of dependent WCS pixel axes.
```
Cadair, line 320: Does the return include the input?

```suggestion
    Given a WCS pixel axis index, return indices of all dependent WCS pixel axes.
```

This is a little weird to reason about as well. Because pixel axes aren't dependant on other pixel axes, the pixel axes are always separable (they just might have more than one world axis associated with them). So I think this description really needs rewriting to make it clear what's happening.

ndcube/utils/wcs.py
DanRyanIrish: Fair point.  Unless you have a preferred suggestion, I'll rename it ```convert_between_array_and_pixel_axes```.

ndcube/utils/wcs.py
DanRyanIrish: That's fair enough.  I'll change it to the following:
```
"""
Return indices of all pixel axes associated with the world axes that correspond to the input pixel axis.

For example, say the input pixel axis is 0 and it is associated with two world axes corresponding to longitude and latitude.
Let's also say that pixel axis 1 is also associated with longitude and latitude.
Thus, this function would return pixel axes 0 and 1.
On the other hand let's say pixel axis 2 is associated with only one world axis, say wavelength, which does not depend on any other pixel axis (i.e. it is independent).
In that case this function would only return pixel axis 2.
"""
```

DanRyanIrish:
> Generally looks good. Most of the new utils are sufficiently general I wonder if we should PR them to astropy?

I think that's a good idea.  Although I also think we should merge them here for now.  If they get merged into astropy before ndcube 2.0, we can remove them.  If not, we can deprecated them later.  This way we aren't held up in ndcube but are still able to depend on them from astropy when the time comes.

DanRyanIrish:
@Cadair, your requested changed have been pushed.  I'll merge as soon as the tests pass.

changelog/280.breaking.rst
```diff
@@ -0,0 +1 @@
+Refactor utils.
```
DanRyanIrish, line 1:
```suggestion
Remove unused util functions and the ndcube WCS class.  Refactor util functions for converting between between data and WCS indices to reflect the APE14 distinctions of array, pixel and world nomenclature. 
```

Cadair:
Oh yeah we should merge them here as well.

changelog/280.breaking.rst
```diff
@@ -0,0 +1 @@
+Remove unused util functions and the ndcube WCS class.  Refactor util functions for converting between between data and WCS indices to reflect the APE14 nomenclature that distinguishes between array, pixel and world axes.
```
DanRyanIrish, line 1:
```suggestion
Remove unused util functions and the ndcube WCS class.  Refactor util functions for converting between between data and WCS indices to reflect the APE14 nomenclature that distinguishes between array, pixel and world axes.
```

PR #287, opened 2020-09-23
Fix a bug in world_axis_coord_values when there is only one dimension in the WCS

Cadair:
For reasons of being nice to the user but unpleasant to the library developer the return type of `pixel_to_world` and `pixel_to_world_values` changes if n_dim == 1. This fixes a bug where the number of axes in the input WCS is 1, and adds a regression test.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -33,7 +33,6 @@ jobs:
   parameters:
     submodules: false
     coverage: codecov
-    posargs: -n=4
     envs:
       - macos: py37
         name: py37_test_macos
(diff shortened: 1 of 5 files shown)


DanRyanIrish (approved):
Approved, pending tests passing and whether you feel a changelog and milestone should also be added.

Cadair:
I added another commit, which ensures this method always returns Quantity objects not high-level objects like SkyCoord etc. I assume that we want to return Quantities here and not numpy arrays like the `wcs.*_values` methods do?

Cadair:
@DanRyanIrish Did we backport some or all of this code?

Cadair:
dev_deps fail is because of #288

DanRyanIrish (approved):

DanRyanIrish:
> @DanRyanIrish Did we backport some or all of this code?

We did backport at least some of this code.  ```axis_world_coord_values``` exists on the 1.3 branch but is not released.

Issue #288, opened 2020-09-23
NDCubeSequence plotting is broken with SunPy 2.1.dev


Cadair:
The old ImageAnimatorWCS has been removed as it was deprecated by ArrayAnimatorWCS in 1.1. We need to update our code.

PR #294, opened 2020-10-08
Change NDCubeSequence ImageAnimators parent class to ArrayAnimatorWCS.

DanRyanIrish:
This fixes bug when using sunpy master, described in Issue #288. ```NDCubeSequence``` was using ```ImageAnimatorWCS``` in its plot infrastructure which has now been removed from sunpy master.  This PR changes the infrastructure to use the replacement animator, ```ArrayAnimatorWCS```.

### DIFF
--- a/changelog/294.bugfix.rst
+++ b/changelog/294.bugfix.rst
@@ -1,0 +1,1 @@
+Remove NDCubeSequence animation dependence of deprecated sunpy ImageAnimator and LineAnimator classes in favour of ArrayAnimatorWCS class.
--- a/ndcube/mixins/sequence_plotting.py
+++ b/ndcube/mixins/sequence_plotting.py
@@ -218,9 +218,9 @@ class NDCubeSequencePlotMixin:
                                                        axes_units, data_unit, **kwargs)
             else:
                 # Else produce an image or line animation depending number of plot axes.
-                ax = NDCubeSequenceCubeLikeAnimator(
+                ax = NDCubeSequenceAnimator(
                     self, plot_axis_indices=plot_axis_indices, axes_units=axes_units,
-                    data_unit=data_unit, **kwargs)
+                    data_unit=data_unit, animate_as_cube=True, **kwargs)
         return ax
 
     def _plot_1D_sequence(self, axes_coordinates=None,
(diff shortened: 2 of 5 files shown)


DanRyanIrish:
@Cadair, this PR moves ```ImageAnimatorNDCubeSequence``` and ```ImageAnimatorCubeLikeNDCubeSequence``` to inherit from ```ArrayAnimatorWCS```.  Both NDCubeSequence ImageAnimators act as an API translator between ```NDCubeSequence.plot``` (and ```NDCubeSequence.plot_as_cube```) and ```ArrayAnimatorWCS```.  A wider question of whether the API of ```NDCubeSequence.plot``` should change is beyond the scope of this PR.

This PR is only valid for master.  I don't think it can be simply backported to the 1.3/1.4 branch.  For example, ```missing_axes``` must still be supported by ```ndcube 1.x```.  I will open another PR for the analogous change for that branch.

ndcube/mixins/sequence_plotting.py
```diff
@@ -748,26 +727,29 @@ def __init__(self, seq, wcs=None, axes=None, plot_axis_indices=None,
...
+        # So invert plot_axis_indices.
+        n_seq_dims = len(seq.dimensions)
+        slices_indices = utils.wcs.convert_between_array_and_pixel_axes(
+                np.array(plot_axis_indices), n_seq_dims)
+        slices = [0] * n_seq_dims
+        slices[slices_indices[-1]] = 'y'
+        slices[plot_axis_indices[0]] = 'x'
+        # Construct coord params input
+        if axes_units is not None:
+            coord_params = None  # In future construct axis unit helpers.
```
DanRyanIrish, line 745: @Cadair, I tried taking the ```axes_units``` input and construct ```CoordinateHelper``` objects to control the axes' unit, but couldn't make it work.  The online documentation I found didn't help a lot.  Any tips on how to do this?

DanRyanIrish:
> @Cadair, this PR moves `ImageAnimatorNDCubeSequence` and `ImageAnimatorCubeLikeNDCubeSequence` to inherit from `ArrayAnimatorWCS`. Both NDCubeSequence ImageAnimators act as an API translator between `NDCubeSequence.plot` (and `NDCubeSequence.plot_as_cube`) and `ArrayAnimatorWCS`. A wider question of whether the API of `NDCubeSequence.plot` should change is beyond the scope of this PR.
> 
> This PR is only valid for master. I don't think it can be simply backported to the 1.3/1.4 branch. For example, `missing_axes` must still be supported by `ndcube 1.x`. I will open another PR for the analogous change for that branch.

I may have changed my mind...  Maybe this could be backported but a change would be needed to handle ```missing_axes```.

changelog/294.bugfix.rst
```diff
@@ -0,0 +1 @@
+Remove NDCubeSequence animation dependence of deprecated sunpy ImageAnimator and LineAnimator classes in favour of ArrayAnimatorWCS class.
```
Cadair, line 1: needs an extra changelog about the breaking change of no extra coords

Cadair:
I am a little unclear on what the scope / goal of this PR is. If the goal here is to fix up for 1.4 then I think it is probably fine and should land and get backported. However, this PR as a thing to master needs a lot of updates to standardise the plotting interface to the new API for NDCube on master.

DanRyanIrish:
Agreed that there should be harmonisation between sequence and cube plotting before 2.0.  However, this fix is needed for 1.4 and the changes to the underlying infrastructure are also needed for master.  The question of the API should be addressed in a separate PR.

Cadair (approved):
Needs the extra changelog but otherwise :+1:

PR #303, opened 2020-10-13
[1.3] Remove features and prepare for new bugfix release

Cadair:


ndcube/tests/helpers.py
```diff
@@ -35,15 +35,15 @@ def assert_cubes_equal(test_input, expected_cube):
     assert np.all(test_input.mask == expected_cube.mask)
     assert_wcs_are_equal(test_input.wcs, expected_cube.wcs)
     assert test_input.missing_axes == expected_cube.missing_axes
-    if type(test_input.uncertainty) is not type(expected_cube.uncertainty):
+    if not isinstance(expected_cube.uncertainty, type(test_input.uncertainty)):
```
DanRyanIrish, line 38: There was a reason I compared types rather than used ```isinstance```.  I can't remember the exact details, but there was a case where I wanted to ensure uncertainties were of the same specific type.  But ```isinstance``` would be true when they were both any kind of general uncertainty type.  Whereas comparing the types was a stricter comparison.

DanRyanIrish (approved):

PR #304, opened 2020-10-15
A re-write of the NDCubeABC Class to facilitate discussion on a upcoming SEP

Cadair:
I went full type hints on the ABC, I don't really expect us to pick them up elsewhere at the moment, but they are useful for API specification.

### DIFF
--- a/ndcube/ndcube.py
+++ b/ndcube/ndcube.py
@@ -4,33 +4,72 @@ import numbers
 import textwrap
 import warnings
 from collections import namedtuple
+from typing import Any, Union, Mapping, Iterable, Tuple
 
 import astropy.nddata
 import astropy.units as u
 import numpy as np
-import sunpy.coordinates
-from astropy.utils.misc import InheritDocstrings
-from astropy.wcs.wcsapi import BaseHighLevelWCS, BaseLowLevelWCS, HighLevelWCSWrapper, SlicedLowLevelWCS
-from astropy.wcs.wcsapi.fitswcs import SlicedFITSWCS, custom_ctype_to_ucd_mapping
+import sunpy.coordinates  # pylint: disable=unused-import # noqa
+from astropy.wcs.wcsapi import BaseHighLevelWCS, BaseLowLevelWCS
 
 import ndcube.utils.wcs as wcs_utils
 from ndcube import utils
 from ndcube.mixins import NDCubePlotMixin, NDCubeSlicingMixin
 from ndcube.ndcube_sequence import NDCubeSequence
 
-__all__ = ['NDCubeABC', 'NDCubeBase', 'NDCube', 'NDCubeOrdered']
+__all__ = ['NDCubeABC', 'NDCubeBase', 'NDCube']
 
 
 class NDCubeABC(astropy.nddata.NDDataBase):
 
-    def __init__(self, data, uncertainty=None, mask=None, wcs=None,
-                 meta=None, unit=None, copy=False, extra_coords=None):
+    @abc.abstractmethod
+    def __init__(self,
+                 data: "numpy.typing.ArrayLike",
+                 uncertainty: Any = None,
+                 mask: Any = None,
+                 wcs: Union[BaseLowLevelWCS, BaseHighLevelWCS] = None,
+                 meta: Mapping = None,
+                 unit: Union[str, u.Unit] = None,
+                 copy: bool = False,
+                 extra_coords: "ExtraCoordsABC" = None):
         pass
 
+    @abc.abstractmethod
+    def __getitem__(self, item: Union[int, slice, Iterable[Union[slice, int]]]) -> "NDCubeABC":
+        """
+        An ndcube object must be sliceable in a similar way to a numpy array,
+        however, support for a `step` is not required and no "fancy indexing is allowed".
+
+        When a slice is applied to a NDCube object it must be applied to the following attributes:
+
+          * data
+          * wcs
+          * uncertainty
+          * mask
+          * extra_coords
+
+        the following attributes must be copied unaltered to the returned NDCube:
+
+          * unit
+          * meta
+
+        """
+
+    @abc.abstractmethod
     @property
-    def extra_coords(self) -> ExtraCoordsABC:
+    def extra_coords(self) -> "ExtraCoordsABC":
         """
         The extra coordinates associated with this cube.
+        """
+
+    @abc.abstractmethod
+    @property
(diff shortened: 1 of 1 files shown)


ndcube/ndcube.py
```diff
@@ -4,74 +4,195 @@
...
+        >>> NDCube.all_world_coords(2) # doctest: +SKIP
+
+        """
+
+    @abc.abstractmethod
+    def crop(self,
+             lower_corner: Iterable[Union[u.Quantity, float]],
+             upper_corner: Iterable[Union[u.Quantity, float]] = None,
+             units: Iterable[u.Unit] = None,
+             wcs: BaseHighLevelWCS = None):
```
Cadair, line 162: This needs updating.

ndcube/ndcube.py
```diff
@@ -4,74 +4,195 @@
...
             The length of the iterable must equal the number of data dimensions
             and must have the same order as the data.
 
+        wcs
+            The WCS to use to calculate the pixel coordinates based on the input.
+
         Returns
         -------
         result: NDCube
         """
```
DanRyanIrish, line 194:
```suggestion
abc.abstractmethod
    def crop_by_values(self, lower_corner, upper_corner, units=None, wcs=None):
        """
        Crops an NDCube given lower and upper real world bounds for each real world axis.
        
        Parameters
        ----------
        lower_corner: iterable whose elements are None, `astropy.units.Quantity` or `float`
            An iterable of length-1 Quantities or floats, representing
            the real world coordinate values of the lower corner of
            the region of interest.
            These are input to `astropy.wcs.WCS.world_to_array_index_values`
            so their number and order must be compatible with the API of that method,
            i.e. they must be in world axis order.
            Alternatively, None, can be provided instead of a Quantity or float.
            In this case, the corresponding array axes will be cropped starting from
            0th array index.
        upper_corner: iterable whose elements are None, `astropy.units.Quantity` or `float`
            An iterable of length-1 Quantities or floats, representing
            the real world coordinate values of the upper corner of
            the region of interest.
            These are input to `astropy.wcs.WCS.world_to_array_index_values`
            so their number and order must be compatible with the API of that method,
            i.e. they must be in world axis order.
            Alternatively, None, can be provided instead of a Quantity or float.
            In this case, the corresponding array axes will be cropped to include
            the final array index.
        units: iterable of `astropy.units.Unit`
            The unit of the corresponding entries in lower_corner and upper_corner.
            Must therefore be the same length as lower_corner and upper_corner.
            Only used if the corresponding type is not a `astropy.units.Quantity`.
        wcs: `astropy.wcs.wcsapi.BaseLowLevelWCS`
            The WCS object to used to convert the world values to array indices.
            Although technically this can be any valid WCS, it will typically be
            self.wcs, self.extra_coords.wcs, or self.combined_wcs, combing both
            the WCS and extra coords.
            Default=self.wcs
        
        Returns
        -------
        result: `ndcube.NDCube`
        """
```

ndcube/ndcube.py
```diff
@@ -4,74 +4,195 @@
...
+                 copy: bool = False,
+                 extra_coords: "ExtraCoordsABC" = None):
+        pass
 
-class NDCubeABC(astropy.nddata.NDData, metaclass=NDCubeMetaClass):
+    @abc.abstractmethod
+    def __getitem__(self, item: Union[int, slice, Iterable[Union[slice, int]]]) -> "NDCubeABC":
+        """
+        An ndcube object must be sliceable in a similar way to a numpy array,
+        however, support for a `step` is not required and no "fancy indexing is allowed".
```
hayesla, line 41: fancy indexing?

ndcube/ndcube.py
```diff
@@ -4,74 +4,195 @@
...
+             upper_corner: Iterable[Union[u.Quantity, float]] = None,
+             units: Iterable[u.Unit] = None,
+             wcs: BaseHighLevelWCS = None):
         """
         Crops an NDCube given minimum values and interval widths along axes.
 
         Parameters
         ----------
-        lower_corner: iterable of `astropy.units.Quantity` or `float`
+        lower_corner
```
hayesla, line 168: are lower_corner, upper_corner etc going to be astropy quantities (or tuples of such)?

ndcube/ndcube.py
hayesla: The WCS object to used to convert  -> The WCS object _is_ used to convert

hayesla: I wonder where the `crop_by_values`  idea came from? I remember it being discussed before. Is it due to the fact you crop by read world values rather than data array indices? What about  `crop_by_wcs`  or something? or maybe that doesn't make sense.

ndcube/ndcube.py
```diff
@@ -4,74 +4,195 @@
...
+
+        Example
+        -------
+        >>> NDCube.all_world_coords(('lat', 'lon')) # doctest: +SKIP
+        >>> NDCube.all_world_coords(2) # doctest: +SKIP
+
+        """
+
+    @abc.abstractmethod
+    def axis_world_coords_values(self,
```
hayesla, line 124: Am I missing something but what is the major difference between `axis_world_coords` and `axis_world_coords_values` - its not super clear from the docstrings

DanRyanIrish: `axis_world_coords` returns high level objects, e.g. astropy `Time`, `SkyCoord`, `SpectralCoord`, etc. which may or may not combine world axes into single objects, e.g. lat, lon in a `SkyCoord`.  `axis_world_coords_values` returns a `Quantity` for each world axis.  So lat and lon are separate.  `axis_world_coords`'s object can have extra contextual information like what frame you're in.  While `_values` returns Quantities that may need extra information to interpret correctly, but can be easier to handle.

DanRyanIrish:
The APIs of `GlobalCoords` and `ExtraCoords` should be harmonized, specifically for adding and removing coordinates.  At the moment we have `ExtraCoords.add_coordinate` and `GlobalCoords.add`/`GlobalCoords.remove`.  We could change the `ExtraCoords` or `GlobalCoords` API, but the main thing is that the APIs are consistent.  I propose we change to `ExtraCoords.add`.

ndcube/ndcube.py
Cadair: I think I mean https://numpy.org/doc/stable/reference/arrays.indexing.html#advanced-indexing

Cadair:
I am going to close this. It's out of date and also we should be having all this discussion over on https://github.com/sunpy/sunpy-SEP/pull/51

PR #312, opened 2020-11-05
Make figure tests work with pytest-mpl 0.12

Cadair:


ndcube/tests/helpers.py
```diff
@@ -3,22 +3,74 @@
...
+    sunpy_version = "dev" if "dev" in sunpy.__version__ else sunpy.__version__.replace('.', '')
+    astropy_version = "dev" if "dev" in astropy.__version__ else astropy.__version__.replace('.', '')
+    return f"figure_hashes_mpl_{mpl_version}_ft_{ft2_version}_astropy_{astropy_version}_sunpy_{sunpy_version}.json"
+
+
+def figure_test(test_function):
+    """
+    A decorator for a test that verifies the hash of the current figure or the
+    returned figure, with the name of the test function as the hash identifier
+    in the library. A PNG is also created in the 'result_image' directory,
```
DanRyanIrish, line 46: Is the result_image a tmp dir or is it actually in the repo?

tox.ini
```diff
@@ -36,15 +26,35 @@ changedir = .tmp/{envname}
...
     test
 deps =
     pytest-xdist
-    devdeps: git+https://github.com/sunpy/sunpy
     devdeps: git+https://github.com/astropy/astropy
+    devdeps: git+https://github.com/sunpy/sunpy
+
+    # Figure tests need a tightly controlled environment
+    figure: matplotlib==3.3.2
+    figure-!devdeps: astropy==4.1.0
```
DanRyanIrish, line 51: What does ```-!``` mean?

DanRyanIrish (approved):

tox.ini
Cadair: Not that factor so `-figure` would match but `-figure-devdeps` would not.

ndcube/tests/helpers.py
Cadair: This docstring is well out of date, will nuke.

PR #314, opened 2020-11-05
Make pyplot color bar work with 2D output of NDCube.plot and enable NDCube to be passed to WCSAxes as a projection

### LINKED ISSUE #206: Add _as_mpl_axes to NDCube
In APE14 land it should be possible to provide an NDCube to matplotlib as a projection in the same way you can a Map. This requires adding a `_as_mpl_axes` method.

See here: https://github.com/sunpy/sunpy/blob/master/sunpy/map/mapbase.py#L509 for the sunpy map implementation.

### LINKED ISSUE #217: No easy way to add a colorbar for a 2D ndcube plot
Currently `NDCube.plot()` only returns the `WCSAxes` instance and not the plotted image, which means it's not possible to add a colorbar post-plotting. I think it should be possible to do something like this SunPy example with NDCube objects: https://docs.sunpy.org/en/stable/generated/gallery/plotting/map_editcolormap.html#sphx-glr-generated-gallery-plotting-map-editcolormap-py

DanRyanIrish:
Fixes #206 
Fixes #217


DanRyanIrish:
@Cadair this is ready for a review.  Assigning this to you now.

DanRyanIrish:
Is that an approval?

Cadair (approved):

Cadair:
Would be good to have tests for both, for `_as_mpl_axes` do something like this:
```
def test():
  ax = plt.subplot(projecton=cube)
  assert isinstance(ax, WCSAxes)
```

ndcube/tests/test_plotting.py
```diff
@@ -99,3 +99,11 @@ def test_animate_cube_from_slice(ndcube_4d, cslice, kwargs, bugged):
...
 
     return ax.fig
+
+
+@figure_test
+@pytest.mark.parametrize(("ndcube_4d", "cslice"), [("simple", np.s_[:, :, 0, 0])])
+def test_mpl_axes(ndcube_4d, cslice):
+    ndcube_2d = ndcube_4d[cslice]
+    ax = plt.subplot(projection=ndcube_2d)
+    assert isinstance(ax, WCSAxes)
```
Cadair, line 109:
```suggestion
@pytest.mark.parametrize(("ndcube_4d", "cslice"), [("simple", np.s_[:, :, 0, 0])])
def test_mpl_axes(ndcube_4d, cslice):
    ndcube_2d = ndcube_4d[cslice]
    ax = plt.subplot(projection=ndcube_2d)
    assert isinstance(ax, WCSAxes)
    plt.close()
```

Cadair: This doesn't need to be a figure test, but we do need to close the figure.

changelog/314.feature.1.rst
```diff
@@ -0,0 +1 @@
+Add as_mpl_axes method to NDCube plotting mixin so the an NDCube can be provided to astropy WCSAxes as a projection.
```
Cadair, line 1:
```suggestion
Add a `as_mpl_axes` method to `NDCube` plotting mixin so the an `NDCube` can be provided to astropy `WCSAxes` as a projection.
```

changelog/314.feature.2.rst
```diff
@@ -0,0 +1 @@
+Make pyplot colorbar work with the output on NDCube.plot when it is a 2D image.
```
Cadair, line 1:
```suggestion
Make `~matplotlib.pyplot.colorbar` work with the output on `NDCube.plot` when it is a 2D image.
```

Cadair (approved):

PR #316, opened 2020-11-06
NDCube.crop

### LINKED ISSUE #298: crop!
implement new crop method

DanRyanIrish:
This PR introduces a new method `NDCube.crop` which allows users to crop an `NDCube` by giving real world coordinates of the lower and upper edges of a region of interest.  This method combines and simplifies `NDCube.crop_by_coords` and `NDCube.crop_by_extra_coords`.

Fixes #298 

To Do:
- [x] Rewrite `crop` to mimic current `crop_by_values` API
- [x] Test errors thrown by `crop` methods
- [x] Test more cases:
  - Cropping in case of 1D WCS
- [x] Delete `crop_by_coords`, `crop_by_extra_coords` and their tests.
- [x] Update `crop` docstring


DanRyanIrish:
@Cadair, is this the right direction we want to go in for `NDCube.crop`?

ndcube/ndcube.py
```diff
@@ -33,6 +33,37 @@ class NDCubeABC(astropy.nddata.NDData, metaclass=NDCubeMetaClass):
     def dimensions(self):
         pass
 
+    @abc.abstractproperty
+    def crop(self, lower_corner, upper_corner, coords_set='wcs', crop_by_values=False):
```
Cadair, line 37: I think I would rather have a `crop_by_values` method than a kwarg here. I think it's more inline with the world axis coords methods, and the methods on the WCS objects.

Cadair, line 37: I am not convinced by either `coords` or `coords_set` for this kwarg name, but I haven't got a better option right now (other than `wcs`).

DanRyanIrish: I suspected you would say that.  But the option of cropping by values, whatever the API, is something you think we should offer users?

DanRyanIrish: That's what I started with.  But then I thought `wcs='wcs'` and `wcs='extra coords'` would seem a bit confusing to people who aren't us.

Cadair: I can see the benefit as far as API completeness goes and some people would probably find it useful. That being said, I worry about diluting the benefits of our Coordinates + quantities first API.

DanRyanIrish:
@Cadair, this is ready for a review.  I am still working on testing the case where the NDCube is sliced and/or the WCS is 1D.

ndcube/ndcube.py
```diff
@@ -321,111 +346,69 @@ def extra_coords(self):
...
+    def crop(self, *intervals, wcs=None):
+        # The docstring is defined in NDCubeBase
+        return self._crop(*intervals, wcs=wcs, crop_by_values=False)
+
+    def crop_by_values(self, *intervals, wcs=None):
+        # The docstring is defined in NDCubeBase
+        return self._crop(*intervals, wcs=wcs, crop_by_values=True)
+
+    def _crop(self, *intervals, wcs=None, crop_by_values=False):
+        # If no intervals provided, return NDCube without slicing.
```
Cadair, line 358: would an error not be more appropriate here?

DanRyanIrish: This comment should read `# If all intervals provides are None, return NDCube without slicing.`  I see this as the equivalent to `same_cube = cube[:, ..., :]`

Cadair: Is that a use that we really need to allow here lol? I don't really understand why this would be useful.

DanRyanIrish: Supporting it doesn't change the code in any way except that an error is or isn't raise in line 361.  Checking whether all intervals are `None` is still required.  And if regular slicing allows all axes to be sliced by `slice(None)`, we might as well mimic that behaviour for `crop`.

PR #318, opened 2020-11-09
Refactor tests

Cadair:
The objective of this PR is to cleanup the test suite so all test setup is done inside pytest fixtures.


ndcube/tests/conftest.py
```diff
@@ -13,36 +13,39 @@
...
+        'CDELT4': 5,
+        'CRPIX4': 5,
+        'CRVAL4': 0,
 
-        'CTYPE1': 'TIME    ',
-        'CUNIT1': 'min',
-        'CDELT1': 0.4,
-        'CRPIX1': 0,
-        'CRVAL1': 0,
+        'DATEREF': "2020-01-01T00:00:00"
```
DanRyanIrish, line 42: In theory could this also be `'DATE-OBS`?  Or is `'DATEREF` the key that astropy WCS recognises?

ndcube/tests/test_ndcube.py
```diff
@@ -1,819 +1,277 @@
...
+                             ("ndcube_3d_ln_lt_l", np.s_[..., 0]),
+                             ("ndcube_3d_ln_lt_l", np.s_[1:2, 1:2, 0]),
+                             ("ndcube_3d_ln_lt_l", np.s_[..., 0]),
+                             ("ndcube_3d_ln_lt_l", np.s_[:, :, 0]),
+                             ("ndcube_3d_ln_lt_l", np.s_[1:2, 1:2, 0]),
+                             ("ndcube_4d_ln_lt_l_t", np.s_[:, :, 0, 0]),
+                             ("ndcube_4d_ln_lt_l_t", np.s_[..., 0, 0]),
+                             ("ndcube_4d_ln_lt_l_t", np.s_[1:2, 1:2, 1, 1]),
+                         ),
+                         indirect=("ndc",))
```
DanRyanIrish, line 34: How does this `indirect` thing work?

ndcube/tests/test_ndcube.py
```diff
@@ -1,819 +1,277 @@
...
+    assert isinstance(sndc.wcs, BaseHighLevelWCS)
+    assert isinstance(sndc.wcs.low_level_wcs, SlicedLowLevelWCS)
+    wcs = sndc.wcs
+    assert wcs.pixel_n_dim == 2
+    assert wcs.world_n_dim == 3
+    assert np.allclose(wcs.array_shape, sndc.data.shape)
+    assert set(wcs.world_axis_physical_types) == {"custom:pos.helioprojective.lat",
+                                                  "custom:pos.helioprojective.lon",
+                                                  "em.wl"}
+    assert np.allclose(wcs.axis_correlation_matrix, np.array([[True, False],
```
DanRyanIrish, line 126: Sure this should be a hard assert, rather than `allclose`?

DanRyanIrish (approved):

ndcube/tests/test_ndcube.py
Cadair: It is a froody fixture lookup thing. Basically it lets you lookup a fixture by name. It passes the name specified in the parametrize to the `ndc` fixture which returns the fixture with the given name.

ndcube/tests/test_ndcube.py
Cadair: Don't think it matters much.

ndcube/tests/conftest.py
Cadair: I think dateref is something slightly different. I think this is a hold over from a SpectralCoord bug.

Issue #284, opened 2020-09-22
A concept of "Global Coordinates"


Cadair:
The rework of extra coords in #271 excludes one use case for the extra_coords dict in 1.0, which is the concept of structured metadata describing the coordinates of the whole cube. i.e. "this (x,y,t) cube is at wavelength y nm".

We should consider adding some kind of functionality to NDCube 2.0 to represent these type of coordinates.

(N.B. this is something that could potentially be upstreamed into the APE 14 WCS interface at some point)

Cadair:
A conversation at the PyHC meeting means we should consider this as a way to represent spacecraft location and maybe frame info.

DanRyanIrish:
Would this be attached to the WCS?

kakirastern:
Will start working on this issue shortly...

DanRyanIrish:
Thanks @kakirastern!

kakirastern:
Yeah, this would be a really great feature to have for Astropy conforming to the [APE 14 WCS](https://docs.astropy.org/en/stable/wcs/wcsapi.html)

kakirastern:
Thanks @DanRyanIrish for the assignment. I will take it up later today HKT during the weekly community meeting.

DanRyanIrish:
In the case of slicing out axes, the coord value at which the slice was made will be stored in this ```low_level_wcs``` property from astropy 4.2: https://github.com/astropy/astropy/pull/10195/files#diff-3976bdca12d482d3c3b404875b2db43f9cb04a37f674e17089d3520cced2a72dR157

Cadair:
```python
class GlobalCoords:
    """
    A structured representation of coordinate information applicable to a whole NDCube.
    """
    def __init__(self, ndcube):
        pass

    def add(self, name, physical_type):
        """
        Add a new coordinate to the collection.
        """

    def remove(self, name):
        """
        Remove a coordinate from the collection
        """

    @property
    def names(self):
        """
        A tuple of all the names.
        """

    @property
    def physical_types(self):
        """
        A tuple of all physical types, one per coordinate.
        """

    def keys(self):
        """
        A set-like of all names in this collection.
        """

    def values(self):
        """
        A set-like of all values in this collection
        """

    def __getitem__(self, item):
        """
        Index the collection by a name.
        """
```

kakirastern:
Will start working on this today HKT

kakirastern:
[...]

PR #322, opened 2020-11-10
Remove sequence plotting

### LINKED ISSUE #321: Remove NDCubeSequence plotting infrastructure for 2.0
Following the decision for 2.0 outlined in #315, remove the `NDCubeSequence` plotting infrastructure.

* Have `NDCubeSequence.plot` and `NDCubeSequence.plot_as_cube` return `NotImplementedError`s.
* Write a brief guide in the docs on how users can extract the data and produce their own visualizations.
* Have the error message point users to Issue #315, and the visualization guide.

DanRyanIrish:
Fixes #321 

* Should be merged after #229 and #284 addressed so that docs changes are consistent with those new APIs.
* Should skip doc tests for section of `NDCubeSequence` plotting


docs/ndcubesequence.rst
```diff
@@ -408,41 +391,167 @@ dimensions.shape ``(<Quantity 3.0 pix>, <Quantity 3.0 pix>, <Quantity
...
+sequence plotting, let us know with a comment on that issue.
+
+However, you can still visualize the data in `~ndcube.NDCubeSequence` in a number of ways.
+You can slice out a single `~ndcube.NDCube` and use its `~ndcube.NDCube.plot` method.
+Or you can write your own mixin class to define the plotting methods.
+Or finally, you can extract the data and use the myriad of plotting packages available in
+the Python ecosystem.  Below, we will outline these latter two options in a little more
+detail.
+
+Writing Your Own NDCubeSequence Plot Mixin
```
Cadair, line 412: This feels quite advanced, especially for putting in the docs before the manual plotting step. Are our docs aimed at users or developers of downstream packages?

docs/ndcubesequence.rst
```diff
@@ -408,41 +391,167 @@ dimensions.shape ``(<Quantity 3.0 pix>, <Quantity 3.0 pix>, <Quantity
...
-    >>> print(data.shape)
-    (3, 4, 5)
+Having extracted the data, we can now use matplotlib to visualize it.
+Let's say we want to produce a timeseries of how intensity changes in a
+given pixel at a given wavelength.  We stored time in ``my_sequence.global_coords``
+and associated it with the ``common_axis``.  Therefore, we could do::
+
+    >>> import matplotlib.pyplot as plt
+    >>> # Get intensity at pixel 0, 0, 0 in each cube.
+    >>> intensity = np.array([cube.data[0, 0, 0] for cube in my_sequence])
```
Cadair, line 479: can't we use the stack here rather than the list comprehension?

ndcube/ndcube_sequence.py
```diff
@@ -6,10 +6,15 @@
 import numpy as np
 
 from ndcube import utils
-from ndcube.mixins.sequence_plotting import NDCubeSequencePlotMixin
 
 __all__ = ['NDCubeSequence']
 
+PLOTTING_NOT_SUPPORTED_ERROR = "NDCubeSequence plotting is no longer supported. " + \
```
Cadair, line 12: I would rewrite this to be a single triple quote string?

Cadair (commented):
Do we really need to wait for global coords to be merged before shipping this? Removing this for b1 is the plan but global coords and merging the animators isn't?

kakirastern (approved):

docs/ndcubesequence.rst
DanRyanIrish: I think it is you're right the the manual stuff should come first.  I think moving the Mixin section to last probably mitigates the audience.  As users will not have to read through it to get to what they need, while developers (who may also be users) will not get lost in the manual stuff before reaching the more advanced mixin section.

docs/ndcubesequence.rst
DanRyanIrish: I think we still need list comprehension to extract the data arrays one by one from the cubes in the sequence.  And in this case, since we are extracting scalar values, we don't need stack on top of the list comprehension.

ndcube/ndcube_sequence.py
DanRyanIrish: Good idea!

DanRyanIrish:
> Do we really need to wait for global coords to be merged before shipping this? Removing this for b1 is the plan but global coords and merging the animators isn't?

Fair point.  I guess we can add the parts were we reference `GlobalCoords` to the docs rewrite.

Cadair:
I mean our docs are already seriously out of sync with reality anyway, so I don't see that making them reference things which are not yet implemented is the end of the world?

PR #323, opened 2020-11-11
Global Coords Implementation

### LINKED ISSUE #284: A concept of "Global Coordinates"

kakirastern:
Fixes #284 

Introduces the concept of `Global Coords`, patches the Extra Coords to render it more feature-complete by implementing the remaining unused use case. 

N.B. this is something that could potentially be upstreamed into the APE 14 WCS interface at some point.


kakirastern:
My PyCharm needs some re-configuration... Need to `.gitigore` some files

kakirastern:
Yay, cannot helped but remark that I managed to get most tests passed without adding the code in the codebase yet...

kakirastern:
Wow, I am impressed the project codecov tests has been passed at this stage at all

kakirastern:
PR ready for a review

CHANGELOG.rst
```diff
@@ -1,3 +1,12 @@
+Ndcube v2.0.0
```
DanRyanIrish, line 1: This change should be removed.  The entry in this file is automatically built my `towncrier` from the `323.feature.rst` file at the time of release.

.gitignore
```diff
@@ -149,5 +149,61 @@ docs/generated
 docs/api
 docs/whatsnew/latest_changelog.txt
 ndcube/version.py
-htmlcov/
+#htmlcov/
```
DanRyanIrish, line 152: Presumably the changes in this file are to do with your local setup?  Should probably also be removed for this PR.

changelog/323.feature.rst
```diff
@@ -0,0 +1 @@
+Introduce concept of global coordinates
```
DanRyanIrish, line 1: This should be slightly more descriptive.  For example, `Introduce a new class, GlobalCoords, for holding scalar coordinates that don't apply to any pixel axes.`

ndcube/global_coords/global_coords.py
```diff
@@ -0,0 +1,76 @@
...
+        super().__init__()
+
+        # Set values using the setters for validation
+        self.mapping = {}
+
+    def add(self, name, physical_type):
+        """
+        Add a new coordinate to the collection.
+        """
+        if len(self.mapping) > 1:
```
DanRyanIrish, line 22: I'm not sure this will work.  What if the `GlobalCoords` instance is empty?  Also it looks like you haven't created a link between `physical_type` and `name`.  For instance, if I want to know the `physical_type` of `name`, there doesn't appear to be a way to do that.  Finally, I don't see where the coord value is stored.  Looks like that's been left out?  I'll add some more discussion in the PR conversation.

DanRyanIrish:
Hi @kakirastern.  I was thinking about the implementation of this and think I need to flesh it out a little more.  I think that will help your implementation.  Here are some examples of how I might want to use it.
```python
>>> gc = GlobalCoords()  # Initialze an empty GlobalCoords instance
>>> import astropy.units as u
>>> coord1 = 1 * u.m
>>> coord2 = 2 * u.s
>>> # Add some coordinates
>>> gc.add('name1', 'physical_type1', coord1)
>>> gc.add('name2', 'physical_type2', coord2)
>>> gc.names  # Access the names of the coordinates in the object.
('name1', 'name2')
>>> gc.physical_types  # Return an iterable of the physical types of the coordinates.
('physical_type1', 'physical_type2')
>>> gc.coords  # Return an iterable of the coordinates themselves.
(<Quantity 1 unit='m'>, <Quantity 2 unit='s'>)
>>> gc_name1 = gc['name1']  # Slice the GlobalCoords object by name
>>> gc_name1
'name1'
>>> gc_name1.coords
<Quantity 1, unit='m'>
>>> gc['name1'].physical_type
'physical_type1'
>>> gc.remove('name2')  # Remove a coordinate
>>> gc == gc_name1  # Confirm that gc has all the same values as gc_name1
True
```

ndcube/global_coords/global_coords.py
kakirastern: Yeah, you are right... I think I was just thinking about special cases only. Let me amend this shortly.

changelog/323.feature.rst
kakirastern: Sure, let me fix it soon.

.gitignore
kakirastern: Yeah, probably. But I think I will need these in the "master" `.gitigore` file for it to work though. Let me double check and follow up accordingly.

CHANGELOG.rst
kakirastern: Yup, sure

kakirastern:
> Hi @kakirastern. I was thinking about the implementation of this and think I need to flesh it out a little more. I think that will help your implementation. Here are some examples of how I might want to use it.
> 
> ```python
> >>> gc = GlobalCoords()  # Initialze an empty GlobalCoords instance
> >>> import astropy.units as u
> >>> coord1 = 1 * u.m
> >>> coord2 = 2 * u.s
> >>> # Add some coordinates
> >>> gc.add('name1', 'physical_type1', coord1)
> >>> gc.add('name2', 'physical_type2', coord2)
> >>> gc.names  # Access the names of the coordinates in the object.
> ('name1', 'name2')
> >>> gc.physical_types  # Return an iterable of the physical types of the coordinates.
> ('physical_type1', 'physical_type2')
> >>> gc.coords  # Return an iterable of the coordinates themselves.
> (<Quantity 1 unit='m'>, <Quantity 2 unit='s'>)
> >>> gc_name1 = gc['name1']  # Slice the GlobalCoords object by name
> >>> gc_name1.names
> 'name1'
> >>> gc_name1.coords
> <Quantity 1, unit='m'>
> >>> gc['name1'].physical_types
> 'physical_type1'
> >>> gc.remove('name2')  # Remove a coordinate
> >>> gc == gc_name1  # Confirm that gc has all the same values as gc_name1
> True
> ```
> 
> So as you see, we want to be able to identify the physical type and coordinate by its name. If I understand your current implementation correctly, you can't do that...?
> 
> @Cadair, would you have different views on the above usage?
> 
> Regarding implementation under the hood, we haven't discussed the details. But perhaps one option could be a `numpy.recarray` with three named columns, `name`, `physical_type`, and `coord` (or a better label) of dtypes `str`, `str`, `object`. The columns would then be indexed by name and the rows by `int`/`slice`. This should make it simple to and efficient to perform operations on the columns, e.g. for extracting all the physical types, and for identifying the physical type and coordinate associated with a given name. It also makes the `__getitem__` of `GlobalCoords` relatively simple. What do you guys think?

Yeah, your version looks more like the one we are aiming for, instead of my original which is more like a skeleton.

kakirastern:
Think I have not worked on NDCube for so long I don't recall much about it. But this will change very quickly shortly.

Cadair:
Generally looks good @DanRyanIrish couple of thoughts:

```python
>>> gc.coords  # Return an iterable of the coordinates themselves.
(<Quantity 1 unit='m'>, <Quantity 2 unit='s'>)
```

Why `.coords` rather than `.values()`? `Mapping` will give us `.values()` anyway, so do we also want to add `.coords`?

```
>>> gc_name1 = gc['name1']  # Slice the GlobalCoords object by name
>>> gc_name1.names
'name1'
```
I think this makes sense, but it's a choice here that we are returning a `GlobalCoords` instance from the slicing.

DanRyanIrish:
> Generally looks good @DanRyanIrish couple of thoughts:
> 
> ```python
> >>> gc.coords  # Return an iterable of the coordinates themselves.
> (<Quantity 1 unit='m'>, <Quantity 2 unit='s'>)
> ```
> 
> Why `.coords` rather than `.values()`? `Mapping` will give us `.values()` anyway, so do we also want to add `.coords`?

My original thought was the `values` is a loaded term when it comes to Quantities and coordinates.  But you're right that it is more consistent with the `Mapping` API.  So I'm happy to go with `.values()`.

> ```python
> >>> gc_name1 = gc['name1']  # Slice the GlobalCoords object by name
> >>> gc_name1.names
> 'name1'
> ```
> 
> I think this makes sense, but it's a choice here that we are returning a `GlobalCoords` instance from the slicing.

Yes, that's true it is a choice.  But I would think there should be a way to get the physical type and coord value by indexing by name.  Having the `__getitem__` return a `GlobalCoords` object is, I think, the simplest way.  Another would be for `.physical_types` and `.values()` to return `dict-like` objects which are sliceable by `name`.  But I think this is messier and probably not consistent with `Mapping`.

changelog/323.feature.rst
kakirastern: Updated

CHANGELOG.rst
kakirastern: Done

.gitignore
kakirastern: Think we will need to keep it if I am working with PyCharm... Have removed the duplicated entry though.

DanRyanIrish:
Great!.  @kakirastern sounds like we have consensus on the above API and on the underlying implementation.  Create a hidden attribute that contains a three column `numpy.recarray` that stores the information.  Then have your methods and properties manipulate and extract info from that.

Cadair:
I am not sure we need a recarray, a list of tuples would probably be enough? Iterating over them isn't likely to be a real issue?

kakirastern:
> I am not sure we need a recarray, a list of tuples would probably be enough? Iterating over them isn't likely to be a real issue?

Agreed

kakirastern:
Have been intending to use this new trick for a tuple instead of a list: https://stackoverflow.com/a/45253740/9959073

kakirastern:
Let me test drive the new implementation just submitted to see if I will get what we expected now...

kakirastern:
Probably will need to include some instance type checking... Not sure though since we want generic?

kakirastern:
Think that should do the trick... Am afraid to test it but will try. Please allow some time...

ndcube/global_coords/global_coords.py
kakirastern: Removed the unnecessary constraint

kakirastern: Or condition...

kakirastern:
Am leaning toward using the inherited properties instead of re-inventing the wheel... If this practice is not good please advice.

DanRyanIrish:
> I am not sure we need a recarray, a list of tuples would probably be enough? Iterating over them isn't likely to be a real issue?

I think we want mutable types because of the `add`/`remove` functionality.  So if we go this way we want a list of lists.

kakirastern:
> > I am not sure we need a recarray, a list of tuples would probably be enough? Iterating over them isn't likely to be a real issue?
> 
> I think we want mutable types because of the `add`/`remove` functionality. So if we go this way we want a list of lists.

Oh, okay... Will need some re-working then

kakirastern:
Updated

kakirastern:
I reckon the variables are just dummies so revised the third to `coords`... As long as consistency is kept then the rest follows.

ndcube/global_coords/global_coords.py
```diff
@@ -0,0 +1,66 @@
...
+        super().__init__()
+
+        # Set values using the setters for validation
+        self.mapping = {}
+
+    def add(self, name, physical_type, coords):
+        """
+        Add a new coordinate to the collection.
+        """
+        self.mapping[name] = [name, physical_type, coords]
```
Cadair, line 22: Putting name in the list here seems pointless, given it's the key of the dict.

ndcube/global_coords/global_coords.py
```diff
@@ -0,0 +1,66 @@
...
+
+class GlobalCoords(Mapping):
+    """
+    A structured representation of coordinate information applicable to a whole NDCube.
+    """
+    def __init__(self, NDCube):
+        super().__init__()
+
+        # Set values using the setters for validation
+        self.mapping = {}
```
Cadair, line 16: The internal state modified by add and remove should be hidden, and needs to be separate from the total computed state.
```suggestion
        self._internal_coords = {}
        
    @property
    def _all_coords(self):
        return self._internal_coords
```

If you then use `_all_coords` everywhere you are currently using `.mapping` (with the excpetion of add and remove which should mutate `_internal_coords`) we can modify the property here to extract coords from the wcs and extra coords in followup PRs.

ndcube/global_coords/global_coords.py
```diff
@@ -0,0 +1,66 @@
...
+        """
+        A set-like of all names in this collection.
+        """
+        return set(self.mapping.keys())
+
+    def values(self):
+        """
+        A set-like of all values in this collection
+        """
+        return set(self.mapping.values())
```
Cadair, line 54: These methods will be implemented by `Mapping`
```suggestion
```

ndcube/global_coords/global_coords.py
```diff
@@ -0,0 +1,66 @@
...
+
+    def __getitem__(self, item):
+        """
+        Index the collection by a name.
+        """
+        return self.mapping[item]
```
Cadair, line 60: This needs to return an instance of self as discussed above.

ndcube/global_coords/global_coords.py
```diff
@@ -0,0 +1,66 @@
...
+
+__all__ = ['GlobalCoords']
+
+
+class GlobalCoords(Mapping):
+    """
+    A structured representation of coordinate information applicable to a whole NDCube.
+    """
+    def __init__(self, NDCube):
+        super().__init__()
```
Cadair, line 13:
```suggestion
    def __init__(self, ndcube):
        super().__init__()
        self._ndcube = ndcube
```

ndcube/global_coords/global_coords.py
kakirastern: Sure

ndcube/global_coords/global_coords.py
kakirastern: Sure

ndcube/global_coords/global_coords.py
kakirastern: But I thought that is what @DanRyanIrish wanted

kakirastern:
Reckon I don't remember what I have learned from the `glue-solar` project that well, but will strive to improve.

ndcube/global_coords/global_coords.py
Cadair: I don't know, it seems redundant.

ndcube/tests/test_global_coords.py
```diff
@@ -0,0 +1,25 @@
+import astropy.units as u
+from ndcube.global_coords import GlobalCoords
+
+gc = GlobalCoords()  # Initialze an empty GlobalCoords instance
+
+# Init
+coord1 = 1 * u.m
+coord2 = 2 * u.s
```
Cadair, line 8: You shouldn't have things defined in the root of the file, they should either be in tests or in fixtures.

kakirastern: Okay, I see

kakirastern: Moved to within the test proper

ndcube/global_coords/global_coords.py
kakirastern: Yeah, changed the LHS to `self._internal_coords[str(name)]` for defining the new dict entry instead

kakirastern:
Reduced the tests to truly a minimum...

ndcube/global_coords/global_coords.py
```diff
@@ -0,0 +1,83 @@
...
+        """
+        A mutable list of all physical types, one per coordinate.
+        """
+        physical_types_list = []
+        for item in self._internal_coords.values():
+            physical_types_list.append(item[1])
+        return physical_types_list
+
+    @property
+    def coords(self):
```
Cadair, line 58: I thought we settled on not implementing this? @DanRyanIrish

ndcube/global_coords/global_coords.py
```diff
@@ -0,0 +1,83 @@
...
+    def names(self):
+        """
+        A mutable list of all the names or keys.
+        """
+        return list(self._internal_coords.keys())
+
+    @property
+    def physical_types(self):
+        """
+        A mutable list of all physical types, one per coordinate.
```
Cadair, line 50: these shouldn't be mutable right?

ndcube/tests/test_global_coords.py
```diff
@@ -0,0 +1,19 @@
+import astropy.units as u
+
+import ndcube
+from ndcube.global_coords import GlobalCoords
+
+gc = GlobalCoords(ndcube)  # Initialize an empty GlobalCoords instance
```
Cadair, line 6: This shouldn't be passed the module, it should be passed an *instance* of the `NDCube` class.

```suggestion
@pytest.fixture
def global_coords(ndcube_3d_ln_lt_l):
    return GlobalCoords(ndcube_3d_ln_lt_l)
```

(remember to add the import)
```

File: ndcube/tests/test_global_coords.py [hunk 14]
```diff
@@ -0,0 +1,19 @@
...
+    coord1 = 1 * u.m
+    coord2 = 2 * u.s
+    gc.add('name1', 'physical_type1', coord1)
+    gc.add('name2', 'physical_type2', coord2)
+    assert gc.names == ['name1', 'name2']
+    assert gc.physical_types == ['physical_type1', 'physical_type2']
+    assert gc.coords == [u.Quantity(1., u.m), u.Quantity(2., u.s)]
+
+    gc.remove('name2')
+    assert len(gc) == 1
```
Cadair, line 19:
```suggestion
def test_some_basic_coordinates(global_coords):
    coord1 = 1 * u.m
    coord2 = 2 * u.s
    global_coords.add('name1', 'physical_type1', coord1)
    global_coords.add('name2', 'physical_type2', coord2)
    assert global_coords.names == ['name1', 'name2']
    assert global_coords.physical_types == ['physical_type1', 'physical_type2']
    assert global_coords.coords == [u.Quantity(1., u.m), u.Quantity(2., u.s)]

    global_coords.remove('name2')
    assert len(global_coords) == 1
```

ndcube/global_coords/global_coords.py
kakirastern: Yeah, true... Let me remove the word "mutable"

ndcube/global_coords/global_coords.py
kakirastern: Yeah, I want to know too

kakirastern:
Nice, all lights finally green... Feedback welcome

ndcube/global_coords/global_coords.py
kakirastern: I saw this in the example to be tested supposedly... Maybe I read too much into it.

ndcube/__init__.py
```diff
@@ -8,4 +9,4 @@
 except ImportError:
     __version__ = "unknown"
 
-__all__ = ['NDCube', 'NDCubeSequence', "NDCollection", "ExtraCoords"]
+__all__ = ['NDCube', 'NDCubeSequence', "NDCollection", "ExtraCoords", "GlobalCoords"]
```
Cadair, line 12: I don't think we want this in this file. I don't think we will ever want the user to instantiate one of these themselves.
```suggestion
__all__ = ['NDCube', 'NDCubeSequence', "NDCollection", "ExtraCoords"]
```

ndcube/__init__.py
```diff
@@ -1,4 +1,5 @@
 from .extra_coords import ExtraCoords
+from .global_coords import GlobalCoords
```
Cadair, line 2:
```suggestion
```

changelog/323.feature.rst
```diff
@@ -0,0 +1 @@
+Introduce a new class, GlobalCoords, for holding scalar coordinates that don't apply to any pixel axes.
```
Cadair, line 1:
```suggestion
Introduce a new class, `~ndcube.global_coords.GlobalCoords`, for holding scalar coordinates that don't apply to any pixel axes.
```

ndcube/global_coords/global_coords.py
```diff
@@ -0,0 +1,83 @@
...
+
+    @property
+    def physical_types(self):
+        """
+        A list of all physical types, one per coordinate.
+        """
+        physical_types_list = []
+        for item in self._internal_coords.values():
+            physical_types_list.append(item[1])
+        return physical_types_list
```
Cadair, line 55:
```suggestion
        return tuple(item[0] for item in self._internal_coords.values())
```

ndcube/global_coords/global_coords.py
DanRyanIrish: Let's drop `coords` and stick with `values` which, as I understand it, should be automatically implemented by the `Mapping` parent class.

ndcube/global_coords/global_coords.py
```diff
@@ -0,0 +1,83 @@
...
+        """
+        Establish the _all_coords property with an _internal_coords.
+        """
+        return self._internal_coords
+
+    def add(self, name, physical_type, coords):
+        """
+        Add a new coordinate to the collection.
+        """
+        self._internal_coords[str(name)] = (name, physical_type, coords)
```
Cadair, line 32: I still think we shouldn't have name in the value of this internal dict. I can see no use for it when it's already the key. Also I don't think we need to cast to `str()` for the key, unless we are going to type check all the input.

```suggestion
        self._internal_coords[name] = (physical_type, coords)
```

ndcube/global_coords/global_coords.py
```diff
@@ -0,0 +1,83 @@
...
+
+    def remove(self, name):
+        """
+        Remove a coordinate from the collection
+        """
+        del self._internal_coords[str(name)]
```
Cadair, line 38:
```suggestion
        del self._internal_coords[name]
```

ndcube/global_coords/global_coords.py
```diff
@@ -0,0 +1,83 @@
...
+
+    @property
+    def names(self):
+        """
+        A list of all the names or keys.
+        """
+        return list(self._internal_coords.keys())
```
Cadair, line 45: I think it's probably better if these properties return tuples, it's just an opinion, but seems more natural.
```suggestion
        return tuple(self._internal_coords.keys())
```

ndcube/global_coords/global_coords.py
```diff
@@ -0,0 +1,83 @@
...
+
+    @property
+    def coords(self):
+        """
+        A list of all coords, one per coordinate.
+        """
+        coords_list = []
+        for item in self._internal_coords.values():
+            coords_list.append(item[2])
+        return coords_list
```
Cadair, line 65:
```suggestion
```

Cadair (commented):
I think this is nearly there, would be good to get it merged today and fold it into b1.

kakirastern:
> I think this is nearly there, would be good to get it merged today and fold it into b1.

Awesome, glad to be of some use to `NDCube`

kakirastern:
Updated the tests and removed the one for `coords`

ndcube/global_coords/global_coords.py
```diff
@@ -0,0 +1,70 @@
...
+        Remove a coordinate from the collection
+        """
+        del self._internal_coords[name]
+
+    @property
+    def names(self):
+        """
+        A list of all the names or keys.
+        """
+        return tuple(self._internal_coords.keys())
```
Cadair, line 45:
```suggestion
        return tuple(self._all_coords.keys())
```

ndcube/global_coords/global_coords.py
```diff
@@ -0,0 +1,70 @@
...
+
+    @property
+    def physical_types(self):
+        """
+        A list of all physical types, one per coordinate.
+        """
+        return tuple(item[0] for item in self._internal_coords.values())
```
Cadair, line 52:
```suggestion
        return tuple(item[0] for item in self._all_coords.values())
```

ndcube/global_coords/global_coords.py
```diff
@@ -0,0 +1,70 @@
...
+
+    def __getitem__(self, item):
+        """
+        Index the collection by a name.
+        """
+        return self._internal_coords[item]
```
Cadair, line 58:
```suggestion
        return self._all_coords[item]
```

ndcube/global_coords/global_coords.py
```diff
@@ -0,0 +1,70 @@
...
+
+    def __iter__(self):
+        """
+        Iterate over the collection.
+        """
+        return iter(self._internal_coords)
```
Cadair, line 64:
```suggestion
        return iter(self._all_coords)
```

ndcube/global_coords/global_coords.py
```diff
@@ -0,0 +1,70 @@
...
+
+    def __len__(self):
+        """
+        Establish the length of the collection.
+        """
+        return len(self._internal_coords)
```
Cadair, line 70:
```suggestion
        return len(self._all_coords)
```

ndcube/global_coords/global_coords.py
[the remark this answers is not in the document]
kakirastern: Could you elaborate?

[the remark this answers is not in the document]
kakirastern: I.e. Has the issue been fixed now?

ndcube/global_coords/global_coords.py
```diff
@@ -0,0 +1,74 @@
...
+    """
+    def __init__(self, ndcube):
+        """
+        Init method.
+        """
+        super().__init__()
+        self._ndcube = ndcube
+
+        # Set values using the setters for validation
+        self._all_coords = OrderedDict()
```
Cadair, line 19: This and the usage in `add` and `remove` as I mentioned before should still be `_internal_coords`. The idea here is that GlobalCoords has multiple sources of data. The one being implemented here is the internal list of coords which users can manually add and remove from. In follow up PRs we will add dynamically getting coords from the wcs and extra_coords of the ndcube.

So we want a property which collates all these sources of coords into one `_all_coords` and one attribute which stores the internal state the user can add and remove from. *Therefore* the `add` and `remove` methods mutate the internal state but all the other methods use the combined set of coords returned by the `_all_coords` property.

kakirastern: Ah, I see. Let me make the changes now...

kakirastern: Fixed... If I have omitted anything, please let me know.

ndcube/global_coords/global_coords.py
```diff
@@ -0,0 +1,77 @@
...
+    A structured representation of coordinate information applicable to a whole NDCube.
+    """
+    def __init__(self, ndcube):
+        """
+        Init method.
+        """
+        super().__init__()
+        self._ndcube = ndcube
+
+        self._all_coords = OrderedDict()
```
Cadair, line 18:
```suggestion
```

ndcube/global_coords/global_coords.py
```diff
@@ -0,0 +1,77 @@
...
+        Establish the length of the collection.
+        """
+        return len(self._all_coords)
+
+    @_all_coords.setter
+    def _all_coords(self, value):
+        """
+        Use setter to set _all_coords
+        """
+        self.__all_coords = value
```
Cadair, line 77:
```suggestion
```

ndcube/global_coords/global_coords.py
```diff
@@ -0,0 +1,77 @@
...
+        """
+        Establish _all_coords as a property returning some _internal_coords.
+        """
+        return self._internal_coords
+
+    def add(self, name, physical_type, coords):
+        """
+        Add a new coordinate to the collection.
+        """
+        self._internal_coords[name] = [physical_type, coords]
```
Cadair, line 32:
```suggestion
        self._internal_coords[name] = (physical_type, coords)
```

ndcube/global_coords/global_coords.py
```diff
@@ -0,0 +1,77 @@
...
+        """
+        A tuple of all physical types, one per coordinate.
+        """
+        return tuple(item[0] for item in self._all_coords.values())
+
+    def __getitem__(self, item):
+        """
+        Index the collection by a name.
+        """
+        return self._all_coords[item]
```
Cadair, line 58: We still need to return a new GlobalCoords instance here
```suggestion
        physical_type, value = self._all_coords[item]
        gc = GlobalCoords(self._ndcube)
        gc.add(item, physical_type, value)
        return gc
```

ndcube/tests/test_global_coords.py
```diff
@@ -0,0 +1,21 @@
...
+def test_some_basic_coordinates(global_coords):
+    coord1 = 1 * u.m
+    coord2 = 2 * u.s
+    global_coords.add('name1', 'physical_type1', coord1)
+    global_coords.add('name2', 'physical_type2', coord2)
+    assert global_coords.names == ('name1', 'name2')
+    assert global_coords.physical_types == ('physical_type1', 'physical_type2')
+
+    global_coords.remove('name2')
+    assert len(global_coords) == 1
```
Cadair, line 21: It would be good to add some extra tests for iterating, slicing and the output of keys and values here as well.

kakirastern: Let me give it a think and add extra tests for it

ndcube/global_coords/global_coords.py
```diff
@@ -0,0 +1,72 @@
...
+        A tuple of all physical types, one per coordinate.
+        """
+        return tuple(item[0] for item in self._all_coords.values())
+
+    def __getitem__(self, item):
+        """
+        Index the collection by a name.
+        """
+        physical_type, value = self._all_coords[item]
+        gc = GlobalCoords(self._ndcube)
```
DanRyanIrish, line 58: I think this should be `gc = type(self)(self._ndcube)`

ndcube/global_coords/global_coords.py
```diff
@@ -0,0 +1,72 @@
...
+        gc.add(item, physical_type, value)
+        return gc
+
+    def __iter__(self):
+        """
+        Iterate over the collection.
+        """
+        return iter(self._all_coords)
```
DanRyanIrish, line 66: I believe we want `iter` to just iterate through the coord values, not the physical types.  So I think this will need some more code.  I think the behaviour we want is:
```python
>>> gc = GlobalCoords()
>>> gc.add('name0', 'type0', 0)  # Where 0 takes the place of an actual coordinate value.
>>> gc.add('name1', 'type1', 1)
>>> for i in gc.items():  # Assuming GlobalCoords will have dictionary's items method. Not sure if it will
...     print(i)
('name0', 0)
('name1', 1)
```

Cadair: > Assuming GlobalCoords will have dictionary's items method. Not sure if it will

yes it should.

ndcube/global_coords/global_coords.py
kakirastern: Sure

ndcube/global_coords/global_coords.py
kakirastern: Okay

ndcube/tests/test_global_coords.py
kakirastern: Done

ndcube/global_coords/global_coords.py
kakirastern: I did something similar

kakirastern:
PR ready for yet another review...

ndcube/global_coords/global_coords.py
DanRyanIrish: This doesn't appear resolved.

ndcube/tests/test_global_coords.py
DanRyanIrish: Could you break out this test so that it tests one functionality at a time?  Namely, `add`, `remove` and `len`.  So break it into 3 tests.

ndcube/global_coords/global_coords.py
```diff
@@ -0,0 +1,72 @@
...
+        """
+        Establish _all_coords as a property returning some _internal_coords.
+        """
+        return self._internal_coords
+
+    def add(self, name, physical_type, coords):
+        """
+        Add a new coordinate to the collection.
+        """
+        self._internal_coords[name] = (physical_type, coords)
```
DanRyanIrish, line 31: I think this should raise an error if there is already a coordinate present of the same name.  Does the `OrderedDict` do that without us have to check that here?

But there should be a way to overwrite an existing coord.  That can of be done by first calling `remove` and then `add`.  but do you want a shortcut?  For example, a `replace=True/False` kwarg?

ndcube/global_coords/global_coords.py
```diff
@@ -0,0 +1,72 @@
...
+
+    def remove(self, name):
+        """
+        Remove a coordinate from the collection
+        """
+        del self._internal_coords[name]
```
DanRyanIrish, line 37: What's the behaviour here if there is no coord called `'name'`?  Does it error or does it just not remove anything?

ndcube/global_coords/global_coords.py
```diff
@@ -0,0 +1,72 @@
...
+
+    @property
+    def names(self):
```
DanRyanIrish, line 40: Do we also have to implement `.keys()`?  If so, it would just be `keys = self._all_coords.keys`.  Or does the magic of `Mapping` do this for us?

DanRyanIrish (changes requested):
Definite progress!

ndcube/global_coords/global_coords.py
kakirastern: Probably going to return `None` by default, but let me double check and add a test for `NoneType` accordingly if that's the case.

ndcube/global_coords/global_coords.py
kakirastern: Think `OrderedDict` guarantees that the order is preserved for the dict, so when iterating over we can rely on the same order every time. We can add an `update` as I have suggested at the very beginning instead of the `remove -> add` sequence as we are dealing with dicts.

ndcube/global_coords/global_coords.py
kakirastern: This is the same as `.keys()` to the best of my knowledge... The `names` can be updated to `keys` or be given any other name for our convenience.

kakirastern:
> This doesn't appear resolved.

Will rewrite in a form more human-understandable soon for better resolution. (I.e., with `.items()` as suggested instead of collecting in order in a `while` loop.)

kakirastern:
> Could you break out this test so that it tests one functionality at a time? Namely, add, remove and len. So break it into 3 tests.

Sure! I have to admit that I have been a bit lazy by lumping all three tests into one as a basic test.

kakirastern:
> Definite progress!

Thanks, I feel the same too!

ndcube/global_coords/global_coords.py
kakirastern: I think having a `'name'` would be like a basic requirement in hindsight...

ndcube/global_coords/global_coords.py
kakirastern: But @Cadair advised against adding an `'update'` function before... What do you guys think?

ndcube/global_coords/global_coords.py
kakirastern: Revised and resolved

ndcube/global_coords/global_coords.py
kakirastern: Otherwise it wouldn't make sense like a person without a name... 😅

kakirastern: **Reality check**: But does it happen in reality though?

ndcube/global_coords/global_coords.py
DanRyanIrish: Maybe we don't want the `update` functionality.  But we still need to know what happens when you try to `add` a coordinate with the same name as an existing coord.  As written, does it error?

ndcube/global_coords/global_coords.py
DanRyanIrish: No.  I mean if you do
```python
>>> gc.keys()  # Only one coordinate called 'Kris'
('Kris',)
>>> gc.remove{'Dan')
???
```
what happens then.

ndcube/global_coords/global_coords.py
DanRyanIrish: Where is the change?

ndcube/global_coords/global_coords.py
kakirastern: Sure, let me try now...

ndcube/global_coords/global_coords.py
kakirastern: Let me add a test to ascertain now... (P.S. We don't want Dan removed)

ndcube/global_coords/global_coords.py
kakirastern: Oh, it is in the test...

kakirastern:
```
for i, gc_item in enumerate(global_coords.keys()):
    if i == 0:
        assert (i, gc_item) == (0, 'name1')
    if i == 1:
        assert (i, gc_item) == (1, 'name2')
```

ndcube/tests/test_global_coords.py
kakirastern: Done, will add `__len__()` test soon

ndcube/global_coords/global_coords.py
kakirastern: We will then get a `KeyError` thrown.

kakirastern: Added an `xfail` (expected to fail) test for this special edge case.

ndcube/global_coords/global_coords.py
kakirastern: The existing coord will get overwritten by the new argument with the same `name`... This is really neat!

kakirastern: Added a test for this. 👍🏼

ndcube/tests/test_global_coords.py
kakirastern: Added a simple `len` test for the case of `length=2`

kakirastern:
Any other changes requested?

ndcube/tests/test_global_coords.py
```diff
@@ -0,0 +1,85 @@
...
+    assert len(global_coords) == 2
+
+
+@pytest.mark.xfail(raises=KeyError)
+def test_global_coords_remove_nonexistent_name(global_coords):
+    coord1 = 1 * u.m
+    coord2 = 2 * u.s
+    global_coords.add('name1', 'physical_type1', coord1)
+    global_coords.add('name2', 'physical_type2', coord2)
+    global_coords.remove('name0')  # x-failing the function on purpose
```
Cadair, line 76: It's better to do this than use xfail:

```suggestion
def test_global_coords_remove_nonexistent_name(global_coords):
    coord1 = 1 * u.m
    coord2 = 2 * u.s
    global_coords.add('name1', 'physical_type1', coord1)
    global_coords.add('name2', 'physical_type2', coord2)
    with pytest.raises(ValueError):
        global_coords.remove('name0')
```

ndcube/global_coords/global_coords.py
DanRyanIrish: This isn't what we need.

ndcube/global_coords/global_coords.py
DanRyanIrish: Please read my initial comment more closely.  I don't think this is the behaviour we want.

ndcube/tests/test_global_coords.py
```diff
@@ -0,0 +1,85 @@
...
+    return GlobalCoords(ndcube_3d_ln_lt_l)
+
+
+def test_adding_global_coords(global_coords):
+    coord1 = 1 * u.m
+    coord2 = 2 * u.s
+    global_coords.add('name1', 'physical_type1', coord1)
+    global_coords.add('name2', 'physical_type2', coord2)
+    assert global_coords.names == ('name1', 'name2')
+    assert global_coords.physical_types == ('physical_type1', 'physical_type2')
```
DanRyanIrish, line 18: Also check the value of the coord here.

ndcube/tests/test_global_coords.py
```diff
@@ -0,0 +1,85 @@
...
+def test_slicing(global_coords):
+    coord1 = 1 * u.m
+    coord2 = 2 * u.s
+    global_coords.add('name1', 'physical_type1', coord1)
+    global_coords.add('name2', 'physical_type2', coord2)
+    assert global_coords['name1']._all_coords == {'name1': ('physical_type1', u.Quantity(1., u.m))}
+    assert global_coords['name2']._all_coords == {'name2': ('physical_type2', u.Quantity(2., u.s))}
+
+
+def test_dict_keys_and_values(global_coords):
```
DanRyanIrish, line 53: This test doesn't test what the test name suggests.

ndcube/tests/test_global_coords.py
```diff
@@ -0,0 +1,85 @@
...
+
+def test_global_coords_remove_nonexistent_name(global_coords):
+    coord1 = 1 * u.m
+    coord2 = 2 * u.s
+    global_coords.add('name1', 'physical_type1', coord1)
+    global_coords.add('name2', 'physical_type2', coord2)
+    with pytest.raises(KeyError):
+        global_coords.remove('name0')
+
+
```
DanRyanIrish, line 78:
```suggestion
```
If a `KeyError` is raised by the `OrderedDict` by default that's fine.  We don't need to test that at this stage.

ndcube/tests/test_global_coords.py
```diff
@@ -0,0 +1,85 @@
...
+    coord2 = 2 * u.s
+    global_coords.add('name1', 'physical_type1', coord1)
+    global_coords.add('name2', 'physical_type2', coord2)
+    global_coords.remove('name2')
+    assert len(global_coords) == 1
+    assert global_coords.names == ('name1',)
+    assert global_coords.physical_types == ('physical_type1',)
+
+
+def test_iterating(global_coords):
```
DanRyanIrish, line 32: See comment in thread above.  As said there, this is not the behaviour we want.

DanRyanIrish (changes requested):

ndcube/global_coords/global_coords.py
kakirastern: @Cadair suggested an improvement over the `xfail` approach.

ndcube/global_coords/global_coords.py
kakirastern: Oh, okay... Think your expected behavior and the test are not entirely mutually exclusive.

ndcube/tests/test_global_coords.py
kakirastern: Yeah, I understand your concerns and they are valid... I was using a by-default naming convention. Let me update this accordingly.

ndcube/tests/test_global_coords.py
kakirastern: Okay, let me erase this from the PR...

ndcube/global_coords/global_coords.py
kakirastern: Let me double check for you...

kakirastern: Should I remove the test? Think the existing code base got everything you needed for this PR covered.

ndcube/global_coords/global_coords.py
kakirastern: So now if we use the same name and update by doing `global_coords.add('name1', 'physical_type2', coord2)`, we will have the existing `global_coords` replaced (or overwritten)... Is this the expected behavior?

kakirastern: This has been done by inheritance from some base properties.

ndcube/tests/test_global_coords.py
kakirastern: Changing from variable `names` to `values` for `keys`...

ndcube/tests/test_global_coords.py
kakirastern: Could you elaborate a bit on the expected behavior? I think I am a bit confused from the discussion previously.

ndcube/global_coords/global_coords.py
kakirastern: The way I proposed to test for `__iter__()` is only one way to do it. There can be others, or none at all.

ndcube/tests/test_global_coords.py
```diff
@@ -0,0 +1,85 @@
...
+            assert (i, gc_item) == (1, 'name2')
+
+
+def test_slicing(global_coords):
+    coord1 = 1 * u.m
+    coord2 = 2 * u.s
+    global_coords.add('name1', 'physical_type1', coord1)
+    global_coords.add('name2', 'physical_type2', coord2)
+    assert global_coords['name1']._all_coords == {'name1': ('physical_type1', u.Quantity(1., u.m))}
+    assert global_coords['name2']._all_coords == {'name2': ('physical_type2', u.Quantity(2., u.s))}
```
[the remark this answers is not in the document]
kakirastern: Done

ndcube/tests/test_global_coords.py
kakirastern: Done

ndcube/tests/test_global_coords.py
kakirastern: Done

kakirastern:
@DanRyanIrish Thanks for the feedback! I have made the changes accordingly.

kakirastern:
I am not sure if the changes I have made are in line with what you have in mind. Please advise if that is not the case. 😀

ndcube/tests/test_global_coords.py
kakirastern: I tried `.items()` already but that would yield some other behavior.

kakirastern: So resorted to using `enumerate()` in the `for` loop...

DanRyanIrish:
@Cadair, I think this PR is good to go now.  We ended up including a `coords` property and making `__getitem__` mimic the slicing of `self._all_coords`, i.e. it returns `(physical_type, coord_value)`, not a `GlobalCoords` instance.  This makes the inherited `Mapping` methods, e.g. `items`, behave more sensibly.  However, this also meant that we did in fact new a `coords` method to get only the coord values.  So that's implemented.  If you would like to give any further feedback, let us know.

DanRyanIrish (approved):

ndcube/global_coords/global_coords.py
```diff
@@ -0,0 +1,81 @@
...
+        A tuple of all the names or keys.
+        """
+        return tuple(self._all_coords.keys())
+
+    @property
+    def physical_types(self):
+        """
+        A tuple of all physical types, one per coordinate.
+        """
+        return tuple(item[0] for item in self._all_coords.values())
```
DanRyanIrish, line 52:
```suggestion
        return dict((name, item[0]) for name, item in zip(self.names, self._all_coords.values()))
```

ndcube/global_coords/global_coords.py
```diff
@@ -0,0 +1,81 @@
...
+
+    @property
+    def coords(self):
+        return tuple(item[1] for item in self._all_coords.values())
```
DanRyanIrish, line 56:
```suggestion
        return dict((name, item[1]) for name, item in zip(self.names, self._all_coords.values()))
```

DanRyanIrish (changes requested):
@kakirastern, One more requested API change.  I think it better if we have `GlobalCoords.physical_types` and `GlobalCoords.coords` return dictionaries so we can index them by `name`.  I've left the changes as "suggested changes" in inline comments.

kakirastern:
> @kakirastern, One more requested API change. I think it better if we have `GlobalCoords.physical_types` and `GlobalCoords.coords` return dictionaries so we can index them by `name`. I've left the changes as "suggested changes" in inline comments. The test will then need to be adjusted accordingly.

No problem. Changes committed. Will need to update the tests next.

ndcube/global_coords/global_coords.py
```diff
@@ -0,0 +1,81 @@
...
+    def physical_types(self):
+        """
+        A tuple of all physical types, one per coordinate.
+        """
+        return dict((key, value[0]) for key, value in self.items())
+
+    @property
+    def coords(self):
+        return dict((key, value[1]) for key, value in self.items())
+
```
DanRyanIrish, line 57:
```suggestion

def get_physical_type(self, name):
    """Return the physical type of a specific coordinate."""
    return self._all_coords[name][0]

def get_coord(self, name):
    """Return value of a specific coordinate."""
    return self._all_coords[name][1]
```

DanRyanIrish (changes requested):

ndcube/global_coords/global_coords.py
DanRyanIrish, line 57:
```suggestion
        return tuple(item[0] for item in self.values())
```

DanRyanIrish (changes requested):

ndcube/tests/test_global_coords.py
```diff
@@ -0,0 +1,92 @@
...
+    coord1 = 1 * u.m
+    coord2 = 2 * u.s
+    gc.add('name1', 'physical_type1', coord1)
+    gc.add('name2', 'physical_type2', coord2)
+    value_list = list(gc.values())
+    for i, element in enumerate(value_list):
+        if i == 0:
+            assert element[1] == coord1
+        if i == 1:
+            assert element[1] == coord2
```
DanRyanIrish, line 92:
```suggestion
def test_get_coord(gc):
    coord1 = 1 * u.m
    gc.add('name1', 'physical_type1', coord1)
    assert gc.get_coord('name1') == coord1
    
    
def test_get_physical_type(gc):
    coord1 = 1 * u.m
    gc.add('name1', 'physical_type1', coord1)
    assert gc.get_phsyical_type('name1') == 'physical_type1'
```

DanRyanIrish (changes requested):

ndcube/tests/test_global_coords.py
```diff
@@ -79,14 +79,13 @@ def test_len(gc):
...
-            assert element[1] == coord1
-        if i == 1:
-            assert element[1] == coord2
+    assert gc.get_coord('name1') == coord1
+    
+    
+def test_get_physical_type(gc):
+    coord1 = 1 * u.m
+    gc.add('name1', 'physical_type1', coord1)
+    assert gc.get_phsyical_type('name1') == 'physical_type1'
```
DanRyanIrish, line 91:
```suggestion
    assert gc.get_physical_type('name1') == 'physical_type1'
```

ndcube/tests/test_global_coords.py
```diff
@@ -0,0 +1,91 @@
...
+    gc.add('name2', 'physical_type2', coord2)
+    assert len(gc) == 2
+
+
+def test_get_coord(gc):
+    coord1 = 1 * u.m
+    gc.add('name1', 'physical_type1', coord1)
+    assert gc.get_coord('name1') == coord1
+    
+
```
DanRyanIrish, line 87:
```suggestion


```

DanRyanIrish (approved):

Cadair (approved):

ndcube/ndcube.py
```diff
@@ -186,11 +187,22 @@ def __init__(self, data, wcs=None, uncertainty=None, mask=None, meta=None,
...
+        self._global_coords = GlobalCoords(self)
 
     @property
     def extra_coords(self):
+        """
+        An `.ExtraCoords` object holding extra coordinates aligned to array axes.
+        """
         return self._extra_coords
 
+    @property
```
DanRyanIrish, line 199: Do we want to expose this directly so users can add and remove coords themselves?

Cadair: they can, this will return the actual object by reference. What this is doing is making a single object always be associated with the NDCube. i.e. you can't set the `.global_coords` property, you can only mutate it in place.

ndcube/tests/test_global_coords.py
```diff
@@ -0,0 +1,82 @@
...
+@pytest.fixture
+def gc():
+    return GlobalCoords()
+
+
+@pytest.fixture
+def gc_coords(gc):
+    coord1 = 1 * u.m
+    coord2 = 2 * u.s
+    gc.add('name1', 'physical_type1', coord1)
```
DanRyanIrish, line 16: These need to be `'custom:physical_typeX'`

ndcube/ndcube.py
DanRyanIrish: :+1:

PR #336, opened 2020-11-16
Remove sunpy dependency 2nd attempt

### LINKED ISSUE #149: Make the sunpy dependency optional
We could make the dependency on the visualisation tools in SunPy optional, i.e only include the plotting mixin if SunPy is importable etc.

Certainly we should consider being able to import a version of ndcube without the plotting addons without SunPy installed.

DanRyanIrish:
Fixes #149


DanRyanIrish:
@Cadair, can you remind me again how to create the hashes for the figure tests?

PR #338, opened 2020-11-18
Update array_axis_physical_types to use combined_wcs.

### LINKED ISSUE #330: NDCube.array_axis_physical_types for 2.0
### Description

The existence of `NDCube.wcs`, `NDCube.extra_coords` and `NDCube.combined_wcs` raise a question about what the API of `NDCube.array_axis_physical_types`.  Initially, this was a property that returned the sequence of tuples, one for each array axis, giving the physical types in `NDCube.wcs` which corresponding to each array axis.  With the advent of the other wcses, what should this property return?

### Options
* `array_axis_physical_types` remains as it, returning the array axis physical types from `NDCube.wcs`.
* `array_axis_physical_types` becomes a method that takes a `wcs` as an keyword argument.  This could default to any of the three wcses.
* Separate properties are created to return the array axis physical types for each wcs.
* `array_axis_physical_types` remains a single property but draws its physical types from the `combined_wcs`.

### Decision
Since the purpose of `array_axis_physical_types` is to show which physical types apply to each array axis., it was decided to take option 4, namely to change the WCS from which ```array_axis_physical_types``` draws its physical types, but otherwise leave the API unchanged.  This will require crystal clear documentation.  Should user feedback tell us their is a need to be able to retrieve `array_axis_physical_types` from different WCSs, this functionality can be implemented then.

DanRyanIrish:
Fixes #330


changelog/338.breaking.rst
```diff
@@ -0,0 +1 @@
+Update NDCube.array_axis_physical_types return physical types from extra coords as well as the WCS.
```
Cadair, line 1:
```suggestion
Update `.NDCube.array_axis_physical_types` return physical types from extra coords as well as the WCS.
```

Cadair (approved):

PR #344, opened 2020-11-19
Replace common_axis_extra_coords with common_axis_coords in NDCubeSequence

### LINKED ISSUE #328: NDCubeSequence.common_axis_coords for 2.0
### Description

The introduction of the `ExtraCoords` and `GlobalCoords` framework for ndcube 2.0 have left the `NDCubeSequence.common_axis_extra_coords` broken.  This issue proposes that it be replaced by a new property, `NDCubeSequence.common_axis_coords` that identifies the coordinates associated with the `common_axis` for all `NDCube`s within the `NDCubeSequence`.  These coordinates should be drawn from both the `NDCube.extra_coords` and `NDCube.wcs`.  A new `dict`-like object should be returned, which for each coordinate holds a list of high-level objects whose length is 1 along the `common_axis`.

This feature may share some implementation with #329

### LINKED ISSUE #343: array_indices_for_world_objects fails when passed a combined_wcs
### Description

When `array_indices_for_world_objects` is passed an `NDCube.combined_wcs` which has a type of `HighLevelWCSWrapper`, it fails with the following error:
```python
if axes:
            world_indices = calculate_world_indices_from_axes(wcs, axes)
        else:
            world_indices = np.arange(wcs.world_n_dim)
    
>       object_names = np.array([wao_comp[0] for wao_comp in wcs.world_axis_object_components])
E       AttributeError: 'HighLevelWCSWrapper' object has no attribute 'world_axis_object_components'

../utils/wcs.py:430: AttributeError
```

### Steps to Reproduce

```python
Code goes here
```

### System Details

- SunPy Version:
- Astropy Version:
- Python Version:
- OS information:

DanRyanIrish:
This PR introduces a new method on `NDCubeSequence` that uses the 2.0 infrastructure to collate all coordinates associated with the common axis across all cubes in a sequence.  This replaces the old `common_axis_extra_coords` method, but it also gathers coordinates stored in the WCS, not just the extra coords.

In addition, this PR fixes two bugs in `utils.wcs.array_indices_for_world_objects`:
1. When the WCS input does not have a `world_axis_object_components` attribute the `low_level_wcs` version is tried.  This bug applies to the case when the input WCS is a `HighLevelWCSWrapper`, e.g. `NDCube.combined_wcs`.
2. An indexerror is fixed when there are dependent axes present.

Fixes #328 
Fixes #343

### DIFF
--- a/ndcube/utils/wcs.py
+++ b/ndcube/utils/wcs.py
@@ -427,9 +427,13 @@ def array_indices_for_world_objects(wcs, axes=None):
     else:
         world_indices = np.arange(wcs.world_n_dim)
 
-    object_names = np.array([wao_comp[0] for wao_comp in wcs.world_axis_object_components])
+    try:
+        object_names = np.array([wao_comp[0] for wao_comp in wcs.world_axis_object_components])
+    except AttributeError:
+        object_names = np.array([wao_comp[0]
+                                 for wao_comp in wcs.low_level_wcs.world_axis_object_components])
 
-    array_indices = [[]] * len(set(object_names))
+    array_indices = [[]] * len(object_names)
     for world_index, oname in enumerate(object_names):
         # If this world index is deselected by axes= then skip
         if world_index not in world_indices:
--- a/ndcube/utils/wcs.py
+++ b/ndcube/utils/wcs.py
@@ -426,22 +426,20 @@ def array_indices_for_world_objects(wcs, axes=None):
         world_indices = calculate_world_indices_from_axes(wcs, axes)
     else:
         world_indices = np.arange(wcs.world_n_dim)
-
-    try:
-        object_names = np.array([wao_comp[0] for wao_comp in wcs.world_axis_object_components])
-    except AttributeError:
-        object_names = np.array([wao_comp[0]
-                                 for wao_comp in wcs.low_level_wcs.world_axis_object_components])
-
+    object_names = np.array([wao_comp[0]
+                             for wao_comp in wcs.low_level_wcs.world_axis_object_components])
     array_indices = [[]] * len(object_names)
     for world_index, oname in enumerate(object_names):
         # If this world index is deselected by axes= then skip
         if world_index not in world_indices:
             continue
-
+        # Select the first occurence of the object name.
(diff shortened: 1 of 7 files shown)


ndcube/utils/wcs.py
```diff
@@ -427,9 +427,13 @@ def array_indices_for_world_objects(wcs, axes=None):
     else:
         world_indices = np.arange(wcs.world_n_dim)
 
-    object_names = np.array([wao_comp[0] for wao_comp in wcs.world_axis_object_components])
+    try:
```
Cadair, line 430: You don't need to do this. NDCube.wcs always is always a high level WCS. So you can always do .low_level_wcs

ndcube/utils/wcs.py
```diff
@@ -427,9 +427,13 @@ def array_indices_for_world_objects(wcs, axes=None):
...
+        object_names = np.array([wao_comp[0] for wao_comp in wcs.world_axis_object_components])
+    except AttributeError:
+        object_names = np.array([wao_comp[0]
+                                 for wao_comp in wcs.low_level_wcs.world_axis_object_components])
 
-    array_indices = [[]] * len(set(object_names))
+    array_indices = [[]] * len(object_names)
```
Cadair, line 436: Pretty sure this needs to be a set still?

DanRyanIrish: Changing that was what fixed the bug.  The elements in which nothing was entered are removed by the tuple comprehension for the for loop.

ndcube/utils/wcs.py
DanRyanIrish: No harm though, right?  Especially since the user has to manually enter the WCS.  This way the function supports High and Low level WCSs.

Cadair: I don't think we need to support both, all the rest of the code is assuming high, and the check just feels like it complicates the code.

ndcube/utils/wcs.py
Cadair: Can you add a comment above like 442 (atleast_1d) saying that it selects the first occurrence of the name.

Cadair (approved):

Issue #342, opened 2020-11-19
Multi-dimensional lookup tables must be passed to ExtraCoords in world order


Cadair:
This is following on from https://github.com/sunpy/ndcube/pull/327#discussion_r526960030. Due to perfectly sensible design decisions in the `LookupTableCoord` layer of the new `ExtraCoords` stack, currently when calling `ExtraCoords.add_coordinate` is called the `lookup_table` argument must be in world order if it is multi-dimensional, i.e transposed with respect to the `NDCube` data array.

We probably want to change this to automatically transpose, but doing that in the `add_coordinate` method is a little fiddily and needs some thought.

DanRyanIrish:
@Cadair, is the main challenges here is that the `ExtraCoords` object, just like a `WCS`, has no knowledge of the data array?  And therefore, there's no obvious way for it to interpret array order inputs?

If so, then why not add an `add_extra_coord` method to `NDCube` which takes the inputs in array order, uses `NDCube`'s knowledge of both the `data` and `extra_coords` to convert them to `ExtraCoords` world order, and then hand them off to `ExtraCoord.add`?

At first glance, this might seem like it's cluttering the `NDCube` API and a self-contained `ExtraCoords` API is preferable.  However, one could argue that cases like this, requiring knowledge of both the data and coordinates, are exactly what `NDCube` is designed for!  And so an `add_extra_coord` is in fact perfectly within the scope of the `NDCube` API.

Cadair:
Note to self: You now have to specify the array axes in world order as well as the array. It's at least consistent but still crazy talk.

Cadair:
Just a note, as of #414 this is now irrelevant until #416 is fixed.

PR #352, opened 2020-11-23
Update docs for 2.0

### LINKED ISSUE #289: Documentation for ndcube 2
This is a generic placeholder for the large documentation rebuild needed for 2.0.

I also wrote this: https://demo.codimd.org/54dm5gPMQmmJux6fV-3AOQ# the other day which might have a home in the docs.

DanRyanIrish:
Fixes #289 

To Do
- [x] Update NDCubeSequence slicing figure
- [x] Make docs render plots in visualization section where appropriate
- [x] Resolve best way to handle duplicate code-blocks
- [x] Determine how to hide repeated code-blocks where needed for doctests but may not be needed for user readability


docs/data_classes.rst
```diff
@@ -0,0 +1,301 @@
+.. _data_classes:
+
+============
+Data Classes
```
Cadair, line 4: I worry a little about the use of "data classes" here given that [`dataclasses`](https://docs.python.org/3/library/dataclasses.html) are now a thing in the standard library.

Should we just call them "classes", "objects" or "types"?

docs/data_classes.rst
```diff
@@ -0,0 +1,301 @@
...
+
+Now let's create an `astropy.wcs.WCS` object.  Let the first world axis be wavelength, the second be helioprojective longitude, the third be helioprojective latitude. Remember that due to convention, the order WCS axes is reversed relative to the data array.  This means that the first two array axes will correspond to helioprojective latitude and longitude and the third to wavelength.
+
+.. code-block:: python
+
+  >>> import astropy.wcs
+  >>> wcs_input_dict = {
+  ... 'CTYPE1': 'WAVE    ', 'CUNIT1': 'Angstrom', 'CDELT1': 0.2, 'CRPIX1': 0, 'CRVAL1': 10, 'NAXIS1': 5,
+  ... 'CTYPE2': 'HPLT-TAN', 'CUNIT2': 'deg', 'CDELT2': 0.5, 'CRPIX2': 2, 'CRVAL2': 0.5, 'NAXIS2': 4,
+  ... 'CTYPE3': 'HPLN-TAN', 'CUNIT3': 'deg', 'CDELT3': 0.4, 'CRPIX3': 2, 'CRVAL3': 1, 'NAXIS3': 3}
```
Cadair, line 32: Would it not be cleaner to use the `WCS` API for this rather than constructing a dict?

docs/data_classes.rst
```diff
@@ -0,0 +1,301 @@
...
+  >>> data0 = np.ones((3, 4, 5))
+  >>> data1 = data0 * 2
+  >>> data2 = data1 * 2
+
+  >>> # Define WCS object for all cubes.
+  >>> import astropy.wcs
+  >>> wcs_input_dict = {
+  ... 'CTYPE1': 'WAVE    ', 'CUNIT1': 'Angstrom', 'CDELT1': 0.2, 'CRPIX1': 0, 'CRVAL1': 10, 'NAXIS1': 5,
+  ... 'CTYPE2': 'HPLT-TAN', 'CUNIT2': 'deg', 'CDELT2': 0.5, 'CRPIX2': 2, 'CRVAL2': 0.5, 'NAXIS2': 4,
+  ... 'CTYPE3': 'HPLN-TAN', 'CUNIT3': 'deg', 'CDELT3': 0.4, 'CRPIX3': 2, 'CRVAL3': 1, 'NAXIS3': 3}
```
Cadair, line 106: same comment as above.

docs/data_classes.rst
```diff
@@ -0,0 +1,301 @@
...
+  >>> my_sequence.cube_like_array_axis_physical_types
+  [('custom:pos.helioprojective.lat', 'custom:pos.helioprojective.lon'),
+   ('custom:pos.helioprojective.lat', 'custom:pos.helioprojective.lon'),
+   ('em.wl',)]
+
+Explode Along Axis
+------------------
+During analysis of some data - say of a stack of images - it may be
+necessary to make some different fine-pointing adjustments to each
+image that isn't accounted for the in the original WCS translations,
```
Cadair, line 168: While this is certainty the point of this method, I do worry about how easy this would be *in general* with APE 14. We should at least test this works with FITS-WCS

docs/data_classes.rst
```diff
@@ -0,0 +1,301 @@
...
+
+Note that an `~ndcube.NDCubeSequence` can be exploded along any axis.  A common axis need not be defined.
+
+To learn how to slice `~ndcube.NDCubeSequence` instances and manipulate sequence coordinates, the :ref:`sequence_slicing` and :ref:`sequence_coordinates` sections.
+
+.. _ndcollection:
+
+NDCollection
+============
+`~ndcube.NDCollection` is a container class for grouping `~ndcube.NDCube` or `~ndcube.NDCubeSequence` instances in an unordered way like in a Python `dict`.  `~ndcube.NDCollection` therefore is differs from `~ndcube.NDCubeSequence` in that the objects contained are not considered to be in any order, are not assumed to represent measurements of the same physical property, and they can have different dimensionalities.  However `~ndcube.NDCollection` is more powerful than a simple `dict` because it enables us to identify axes that are aligned between the objects and hence provides some limited slicing functionality.  (See :ref:`collection_slicing` to for more on slicing.)
```
Cadair, line 205: Python dicts are kina ordered these days.

Cadair, line 205: One line per sentence please.

docs/data_classes.rst
```diff
@@ -0,0 +1,301 @@
...
+
+One possible application of `~ndcube.NDCollection` is linking observations with derived data products.  Let's say we have a 3D `~ndcube.NDCube` representing space-space-wavelength.  Then let's say we fit a spectral line in each pixel's spectrum and extract its linewidth.  Now we have a 2D spatial map of linewidth with the same spatial axes as the original 3D cube.  There is a clear relationship between these two objects and so it makes sense to store them together.  An `~ndcube.NDCubeSequence` is not appropriate here as the physical properties represented by the two objects is different, they do not have an order within their common coordinate space, and they do not have the same dimensionality.  Instead let's use an `~ndcube.NDCollection`.
```
Cadair, line 207: One line per sentence

docs/data_classes.rst
```diff
@@ -0,0 +1,301 @@
...
+
+.. code-block:: python
+
+  >>> my_collection.aligned_axis_physical_types  # doctest: +SKIP
+  [('custom:pos.helioprojective.lon', 'custom:pos.helioprojective.lat'), ('custom:pos.helioprojective.lon', 'custom:pos.helioprojective.lat')]
+
+This returns the a `list` of `tuple` giving the physical types that correspond to each aligned axis.  For each aligned axis, only physical types are associated with all the cubes in the collection are returned.  Note that there is no there is no requirement that all aligned axes must represent the same physical types.  They just have to be the same length.  Therefore, is it possible that this property returns no physical types.
+
+The real power behing `~ndcube.NDCollection.aligned_axes` is that is enables all objects within the `~ndcube.NDCollection` to be sliced along the aligned axes simultaneously form the `~ndcube.NDCollection` level.  This allows users to quickly and accurately crop their entire data set to a region of interest, thereby speeding up their analysis workflow.  See the :ref:`collection_slicing` to see this in action.
```
Cadair, line 296: provide an example of this?

Cadair (commented):
A few general points:

A lot of the method documentation isn't working, I think it's issues with the way our class inheritance is setup. This means both for the class docstrings and the links.

Please use one line per sentence for narrative docs, LaTeX style.

docs/data_classes.rst
DanRyanIrish: I think `Data Objects` is fine.  I wanted to use `Data` to distinguish `NDCube`, `NDCubeSequence` and `NDCollection` from the coordinate objects like `Global Coords`, `ExtraCoords` and other used by `ndcube`.  Another option could be `ND Objects`.  So those three are members of the ND family.

docs/data_classes.rst
DanRyanIrish: This is in API I always use.  Feel free to suggest an alternative.

docs/data_classes.rst
DanRyanIrish: Examples are shown in the slicing section of the docs.

docs/introduction.rst
```diff
@@ -3,49 +3,35 @@ An Introduction to ndcube
...
...el coordinate objects into the realm of ndcube. This includes astropy's SkyCoord object which combines coordinate and reference frame information to give users a full description of their coordinates.  However users can continue to deal with raw coordinate values without reference frame information if they so choose.  ndcube's visualisation code has been rewritten to exclusively use WCSAxes, tremendously simplifying it’s implementation, at the expense of some flexibility.  However, it also allows for a more complete and accurate representation of coordinates along plot axes and animations.  Extra_coords has been completely re-written to serve as an extra WCS, which can be readily constructed from lookup tables.  This enables users to easily combine the extra_coords and WCS coords and to utilize the WCSAxes infrastructure for visualizing extra_coords in their plots.  Finally, a new `~ndcube.GlobalCoords` class can to hold coordinates that do not refer to any axis.  This is particularly useful when the dimensionality of an `~ndcube.NDCube` is reduced by slicing.  The value of a coordinate at the location along the dropped axis at which the `~ndcube.NDCube` was sliced can be retained.
```
hayesla, line 37:
```suggestion
The main feature of ndcube 2.0 is the removal and upstreaming of all the specific WCS handling code to astropy and the use of the astropy's generalised wcsapi.  This has the consequence of bringing high-level coordinate objects into the realm of ndcube. This includes astropy's ~astropy.coordinates.SkyCoord object which combines coordinate and reference frame information to give users a full description of their coordinates.  However users can continue to deal with raw coordinate values without reference frame information if they so choose.  ndcube's visualisation code has been rewritten to exclusively use WCSAxes, tremendously simplifying it’s implementation, at the expense of some flexibility.  However, it also allows for a more complete and accurate representation of coordinates along plot axes and animations.  Extra_coords has been completely re-written to serve as an extra WCS, which can be readily constructed from lookup tables.  This enables users to easily combine the extra_coords and WCS coords and to utilize the WCSAxes infrastructure for visualizing extra_coords in their plots.  Finally, a new `~ndcube.GlobalCoords` class can to hold coordinates that do not refer to any axis.  This is particularly useful when the dimensionality of an `~ndcube.NDCube` is reduced by slicing.  The value of a coordinate at the location along the dropped axis at which the `~ndcube.NDCube` was sliced can be retained.
```

docs/data_classes.rst
Cadair:
```python
wcs = WCS(naxis=3)
wcs.wcs.ctype = 'WAVE', 'HPLT-TAN', 'HPLN-TAN'
wcs.wcs.cunit = 'Angstrom', 'deg', 'deg'
...
```
etc.

docs/introduction.rst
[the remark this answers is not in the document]
hayesla: also link to extra_coords

docs/installation.rst
```diff
@@ -1,6 +1,8 @@
-============
-Installation
-============
+.. _installation::
+
+=================
+Installing ndcube
+=================
 
 ndcube requires Python 3.5+, SunPy 0.9+, astropy and matplotlib.
```
hayesla, line 7: SunPy 0.9+? also what version of Astropy - surely a newer one >4.0?

docs/data_classes.rst
Cadair: I did miss that, but I do wonder if it's worth a very terse example in line?

docs/data_classes.rst
```diff
@@ -0,0 +1,301 @@
+.. _data_classes:
+
+============
+Data Classes
+============
+ndcube provides its features via its data classes: `~ndcube.NDCube`, `~ndcube.NDCubeSequence` and `~ndcube.NDCollection`.  This section describes the purpose of each and how they are structured and instantiated.  To learn how to slicem visualize and perform coordinate transformations with these classes, see the :ref:`slicing`, :ref:`plotting` and :ref:`coordinates` sections.
```
hayesla, line 6:
```suggestion
ndcube provides its features via its data classes: `~ndcube.NDCube`, `~ndcube.NDCubeSequence` and `~ndcube.NDCollection`.  This section describes the purpose of each and how they are structured and instantiated.  To learn how to slice, visualize, and perform coordinate transformations with these classes, see the :ref:`slicing`, :ref:`plotting` and :ref:`coordinates` sections.
```

docs/data_classes.rst
hayesla: I think `Data Objects` sounds good

docs/data_classes.rst
```diff
@@ -0,0 +1,301 @@
...
+
+.. _ndcube:
+
+NDCube
+======
+ndcube's primary data class is `~ndcube.NDCube`.  It's designed for managing a single data array and set of WCS transformations.  `~ndcube.NDCube` provides unified slicing, visualization, coordinate conversion APIs as well as APIs for inspecting the data, coordinate transformations and metadata. `~ndcube.NDCube` does this in a way that is not specific to any number or physical type of axis.  It can therefore be used for any type of data (e.g. images, spectra, timeseries, etc.) so long as those data are represented by an array and a set of WCS transformations. Moreover, `~ndcube.NDCube` is agnostic to the fundamental array type in which the data is stored, as long as it behaves like a numpy array.
```
hayesla, line 12: if data objects then reword here

```suggestion
ndcube's primary data object is `~ndcube.NDCube`.  It's designed for managing a single data array and set of WCS transformations.  `~ndcube.NDCube` provides unified slicing, visualization, coordinate conversion APIs as well as APIs for inspecting the data, coordinate transformations and metadata. `~ndcube.NDCube` does this in a way that is not specific to any number or physical type of axis.  It can therefore be used for any type of data (e.g. images, spectra, timeseries, etc.) so long as those data are represented by an array and a set of WCS transformations. Moreover, `~ndcube.NDCube` is agnostic to the fundamental array type in which the data is stored, as long as it behaves like a numpy array.
```

docs/data_classes.rst
```diff
@@ -0,0 +1,301 @@
...
+Meanwhile, the WCS object can be any class, as long as it adhere's to the AstroPy `wcsapi (APE 14) <https://docs.astropy.org/en/stable/wcs/wcsapi.html>`_ specification. These features and flexibility make `~ndcube.NDCube` ideal for subclassing when creating tools for specific data types of data.  It enables developers and scientists to focus on developing the tools needed for their specific research while leveraging standarized APIs for non-data-type-specific functionalities (e.g. slicing).
```
hayesla, line 13:
```suggestion
Meanwhile, the WCS object can be any class, as long as it adhere's to the AstroPy `wcsapi (APE 14) <https://docs.astropy.org/en/stable/wcs/wcsapi.html>`_ specification. These features and flexibility make `~ndcube.NDCube` ideal for subclassing when creating tools for specific data types of data.  It enables developers and scientists to focus on developing the tools needed for their specific research while leveraging standarized APIs for generic array functionalities (e.g. slicing).
```

docs/data_classes.rst
```diff
@@ -0,0 +1,301 @@
...
+Initialize an NDCube
+--------------------
+To initialize the most basic `~ndcube.NDCube` object, we need is a `numpy.ndarray`-like array containing the data and an APE-14-compliant WCS object (e.g. `astropy.wcs.WCS`) describing the coordinate transformations to and from array-elements. Let's create a 3-D array of data with shape ``(3, 4, 5)`` where every value is 1.
+
+.. code-block:: python
+
+  >>> import numpy as np
+  >>> data = np.ones((3, 4, 5))
+
+Now let's create an `astropy.wcs.WCS` object.  Let the first world axis be wavelength, the second be helioprojective longitude, the third be helioprojective latitude. Remember that due to convention, the order WCS axes is reversed relative to the data array.  This means that the first two array axes will correspond to helioprojective latitude and longitude and the third to wavelength.
```
hayesla, line 24: I think it would be good to give a real world example of what this would be. For example, this could be a data cube which represents 5 images at 5 different wavelengths lalalala...

docs/data_classes.rst
```diff
@@ -0,0 +1,301 @@
...
+  >>> input_wcs = astropy.wcs.WCS(wcs_input_dict)
+
+Now we can create an `~ndcube.NDCube`.
+
+.. code-block:: python
+
+  >>> from ndcube import NDCube
+  >>> my_cube = NDCube(data, input_wcs)
+
+The data array is stored in ``mycube.data`` while the WCS object is stored in ``my_cube.wcs``.  However, when manipulating/slicing the data it is better to slice the object as a whole.  (See section on :ref:`ndcube_slicing`.)  So the ``.data`` attribute should only be used to access specific raw data values.
```
hayesla, line 42: I don't know why but :ref:`ndcube_slicing` isn't rendering a link on readthedocs

docs/data_classes.rst
```diff
@@ -0,0 +1,301 @@
...
+  [('custom:pos.helioprojective.lat', 'custom:pos.helioprojective.lon'),
+   ('custom:pos.helioprojective.lat', 'custom:pos.helioprojective.lon'),
+   ('em.wl',)]
+
+`~ndcube.NDCube.dimensions` returns an `~astropy.units.Quantity` of pixel units giving the length of each dimension in the `~ndcube.NDCube` while `~ndcube.NDCube.array_axis_physical_types` returns tuples of strings denoting the types of physical properties represented by each array axis.  The tuples are arranged in array axis order.  As more than one physical type can be associated with an array axis, the length of each tuple can be greater than 1.  This is the case for the 1st and 2nd array array axes which are associated with the coupled world axes of helioprojective latitude and longitude. The axis names are in accordance with the International Virtual Observatory Alliance (IVOA)
+`UCD1+ controlled vocabulary <http://www.ivoa.net/documents/REC/UCD/UCDlist-20070402.html>`_.
+
+`~ndcube.NDCube` provides many helpful features, specifically regarding coordinate transformations, slicing and visualization.  See the :ref:`cube_coordinates`, :ref:`cube_slicing` and :ref:`cube_plotting` sections.
```
hayesla, line 78: similarly here the `cube_slicing` and `cube_plotting` not linking correctly on readthedocs

docs/data_classes.rst
```diff
@@ -0,0 +1,301 @@
...
...uous data set.  The `~ndcube.NDCube` objects within an `~ndcube.NDCubeSequence` must be have the same shape and physical types associated with each axis.  They must also be arranged in some order.  The direction in which the cubes are ordered is referred to as the "sequence axis".  The sequence axis can be in addition and perpendicular to the cubes' axes, or parallel to one, i.e. they can be arranged in order along one of the cubes' axes.  For example, say we have two images with a shape of 512 x 512 represented by two 2-D `~ndcube.NDCube` objects.  Let's also say they that were taken at two different times, but that their WCS transformations only describe their celestial coordinates. We can place these `~ndcube.NDCube` objects into a `~ndcube.NDCubeSequence` where the sequence axis acts as a 3rd axis representing time.  Thus, the data set has an effective shape of ``(2, 512, 512)``.  However, say those two images are of adjacent regions of the sky.  In that case there is an ordering in one of the spatial axes, say the first.  If we set the common axis to that cube axis (``common_axis=0``), then we can also treat the data as if it were a single cube with a shape of ``(1024, 512)``.
```
hayesla, line 85:
```suggestion
`~ndcube.NDCubeSequence` is a class for handling multiple `~ndcube.NDCube` objects as if they were one contiguous data set.  The `~ndcube.NDCube` objects within an `~ndcube.NDCubeSequence` must have the same shape and physical types associated with each axis.  They must also be arranged in some order.  The direction in which the cubes are ordered is referred to as the "sequence axis".  The sequence axis can be in addition and perpendicular to the cubes' axes, or parallel to one, i.e. they can be arranged in order along one of the cubes' axes.  For example, say we have two images with a shape of 512 x 512 represented by two 2-D `~ndcube.NDCube` objects.  Let's also say that they were taken at two different times, but that their WCS transformations only describe their celestial coordinates. We can place these `~ndcube.NDCube` objects into a `~ndcube.NDCubeSequence` where the sequence axis acts as a 3rd axis representing time.  Thus, the data set has an effective shape of ``(2, 512, 512)``.  However, say those two images are of adjacent regions of the sky.  In that case there is an ordering in one of the spatial axes, say the first.  If we set the common axis to that cube axis (``common_axis=0``), then we can also treat the data as if it were a single cube with a shape of ``(1024, 512)``.
```

hayesla: I think an sketch or schematic here would be really helpful

docs/data_classes.rst
```diff
@@ -0,0 +1,301 @@
...
+------------------
+During analysis of some data - say of a stack of images - it may be
+necessary to make some different fine-pointing adjustments to each
+image that isn't accounted for the in the original WCS translations,
+e.g. due to satellite wobble.  If these changes are not describable
+with a single WCS object, it may be desirable to break up the N-D
+sub-cubes of an `~ndcube.NDCubeSequence` into an sequence of sub-cubes
+with dimension N-1. This would enable a separate WCS object to be
+associated with each image and hence allow individual pointing
+adjustments.
```
hayesla, line 174: again I think a schematic or something here would be very helpful , or a more in-depth high-level example- its a little bit hard to visualize

docs/data_classes.rst
```diff
@@ -0,0 +1,301 @@
...
+
+  >>> # Define derived linewidth NDCube
+  >>> linewidth_data = np.ones((3, 4)) / 2 # dummy data
+  >>> linewidth_wcs_dict = {
+  ...    'CTYPE1': 'HPLT-TAN', 'CUNIT1': 'deg', 'CDELT1': 0.5, 'CRPIX1': 2, 'CRVAL1': 0.5, 'NAXIS1': 20,
+  ...    'CTYPE2': 'HPLN-TAN', 'CUNIT2': 'deg', 'CDELT2': 0.4, 'CRPIX2': 2, 'CRVAL2': 1, 'NAXIS2': 10}
+  >>> linewidth_wcs = astropy.wcs.WCS(linewidth_wcs_dict)
+  >>> linewidth_cube = NDCube(linewidth_data, linewidth_wcs)
+
+To combine these ND objects into an `~ndcube.NDCollection`, simply supply a sequence of
```
hayesla, line 221:
```suggestion
To combine these NDCube objects into an `~ndcube.NDCollection`, simply supply a sequence of
```

docs/data_classes.rst
hayesla, line 296:
```suggestion
The real power behing `~ndcube.NDCollection.aligned_axes` is that it enables all objects within the `~ndcube.NDCollection` to be sliced along the aligned axes simultaneously from the `~ndcube.NDCollection` level.  This allows users to quickly and accurately crop their entire data set to a region of interest, thereby speeding up their analysis workflow.  See the :ref:`collection_slicing` to see this in action.
```

docs/coordinates.rst
```diff
@@ -0,0 +1,302 @@
...
+  ... 'CTYPE1': 'WAVE    ', 'CUNIT1': 'Angstrom', 'CDELT1': 0.2, 'CRPIX1': 0, 'CRVAL1': 10, 'NAXIS1': 5,
+  ... 'CTYPE2': 'HPLT-TAN', 'CUNIT2': 'deg', 'CDELT2': 0.5, 'CRPIX2': 2, 'CRVAL2': 0.5, 'NAXIS2': 4,
+  ... 'CTYPE3': 'HPLN-TAN', 'CUNIT3': 'deg', 'CDELT3': 0.4, 'CRPIX3': 2, 'CRVAL3': 1, 'NAXIS3': 3}
+  >>> input_wcs = astropy.wcs.WCS(wcs_input_dict)
+
+.. _extra_coords::
+
+ExtraCoords
+===========
+In the :ref:`ndcube` section we saw that the WCS object stored at `nducbe.NDCube.wcs` contains the primary set of coordinate transformations that describe the data.  However, what if you have alternative or additional coordinates that are not represented by the WCS?  The `ndcube.ExtraCoords` class provides users with a mechanism of attaching such coordinates to their `~ndcube.NDCube` instances.  Let's start by creating a `~astropy.time.Time` object representing the times corresponding to each position along the 1st axis of the data array defined above.
```
hayesla, line 25: I'd give a list of examples here of what the alternative or additional coordinates are

docs/coordinates.rst
```diff
@@ -0,0 +1,302 @@
...
+  >>> my_extra_coords.add_coordinate('time', (2,), timestamps)  # TO DO: Change the mapping to 0 when bug fixed.
+
+An indefinite number of coordinates can be added in this way.  Alternatively, we can generate an `~ndcube.ExtraCoords` object from a WCS.  The names of the coordinates can be accessed via the `~ndcube.ExtraCoords.keys` method.
+
+.. code-block:: python
+
+  >>> my_extra_coords.keys()
+  ('time',)
+
+The values of the coordinates can be accessed via `~ndcube.ExtraCoords.wcs`.  This property generates a WCS object based on the coordinates stored in the `~ndcube.ExtraCoords` object.  It should be noted that because `~ndcube.ExtraCoords` is a stand alone object, its WCS does not necessarily match the data array of an `~ndcube.NDCube`.  To convert the extra coordinates into a WCS directly applicable to an `~ndcube.NDCube`, see the :ref:`combined_wcs` section below.
```
hayesla, line 48: combined_wcs section not linking correctly in readthedocs

docs/coordinates.rst
```diff
@@ -0,0 +1,302 @@
...
+  <Quantity 1. m>
+
+The coordinate's physical type can be accessed via the `~ndcube.GlobalCoords.physical_types` `dict` property.
+
+.. code-block:: python
+
+  >>> my_cube.global_coords.physical_types['distance']
+  'pos.distance'
+
+Because `~ndcube.GlobalCoords` inherits from `Mapping`, it contains a number of mixin methods similar to those of `dict`.
```
hayesla, line 96: `Mapping`?

docs/coordinates.rst
```diff
@@ -0,0 +1,302 @@
...
+        [(5039.92178073, 4.56894119e-02), (5040.03143627, 1.79995614e+03),
+         (5040.14110851, 3.59986658e+03), (5040.25079745, 5.39950295e+03)]]>)
+
+By default `~ndcube.NDCube.axis_world_coords` returns the coordinates at the
+center of each pixel. However, the pixel edges can be obtained by setting
+the ``edges`` kwarg to ``True``. For example:
+
+.. code-block:: python
+
+  >>> my_cube.axis_world_coords(edges=True)
```
hayesla, line 198: nice!

docs/coordinates.rst
```diff
@@ -0,0 +1,302 @@
...
+
+.. code-block:: python
+
+  >>> combined_ccords = my_cube.axis_world_coords(wcs=my_cube.combined_wcs)
+
+Working with Raw Coordinates
+............................
+
+If users would prefer not to deal with high level coordinate objects, they can elect to use `ndcube.NDCube.axis_world_coords_values`.  The API for this method is the same as `~ndcube.NDCube.axis_world_coords`.  The only difference is that `~astropy.units.Quantity` objects are returned, one for each physical type requested.  In the above case this means that there would be seperate `~astropy.units.Quantity` objects for latitude and longitude, but they would both have the same 2-D shape.  The `~astropy.units.Quantity` objects are returned in world order and correspond to the physical types in the `~astropy.wcs.WCS.world_axis_physical_types`.  The `~astropy.units.Quantity` objects do not contain important contextual information, such as reference frame, which is needed to fully interpret the coordinate values.  However for some use cases this level of completeness is not needed.
```
hayesla, line 223:
```suggestion
If users would prefer not to deal with high level coordinate objects, they can elect to use `ndcube.NDCube.axis_world_coords_values`.  The API for this method is the same as `~ndcube.NDCube.axis_world_coords`.  The only difference is that `~astropy.units.Quantity` objects are returned, one for each physical type requested.  In the above case this means that there would be separate `~astropy.units.Quantity` objects for latitude and longitude, but they would both have the same 2-D shape.  The `~astropy.units.Quantity` objects are returned in world order and correspond to the physical types in the `~astropy.wcs.WCS.world_axis_physical_types`.  The `~astropy.units.Quantity` objects do not contain important contextual information, such as reference frame, which is needed to fully interpret the coordinate values.  However for some use cases this level of completeness is not needed.
```

docs/visualization.rst
```diff
@@ -0,0 +1,215 @@
...
+
+.. code-block:: python
+
+  >>> my_cube.plot(plot_axes=[..., 'y', 'x']) # doctest: +SKIP
+
+`~ndcube.NDCube.plot` uses `~astropy.visualization.wcsaxes.WCSAxes` to produce all plots.  This enables a rigorous representation of the coordinates on the plot, including those that are not aligned to the pixel grid.  It also enables the coordinates along the plot axes to be updated between frames of an animation. `ndcube.NDCube.plot` therefore allows users to decide which WCS object to use, either `~ndcube.NDCube.wcs` or `~ndcube.NDCube.combined_wcs` which also includes the `~ndcube.ExtraCoords`.  In principle, another third-part WCS can be used so long as it is a valid description of all array axes.
+
+.. code-block:: python
+
+  >>> my_cube.plot(wcs=my_cube.combined_wcs)   # doctest: +SKIP
```
hayesla, line 35: it would be nice in general if these actually plot on the docs? just to get a better idea

hayesla (commented):
I think these docs look great! One general comment would be to include the visualizations of the schematics often shown in @DanRyanIrish's talks about NDCube to help users get a better high level view of the differences between the ND objects :)

docs/coordinates.rst
DanRyanIrish: https://docs.python.org/3.8/library/collections.abc.html#collections.abc.Mapping

docs/visualization.rst
Cadair: See https://matplotlib.org/sampledoc/extensions.html#inserting-matplotlib-plots for how to do that.

docs/installation.rst
DanRyanIrish: Great catch!

DanRyanIrish:
@Cadair or @nabobalis, do either of you know how to store code snippets for the docs in different files and then include them in the docs 1st file?  I've tried but haven't been able to figure how to do it in such a way that the doctests are aware of code executed from different files.  So the doctests fail.

This may be the same question, but is there a way to have code execute but not shown in the rendered docs?  For example, if you don't want to show a bit of code that was discussed in a previous section, can you have that code execute for the doctests but not clutter up the code block for the user to read.

nabobalis:
> @Cadair or @nabobalis, do either of you know how to store code snippets for the docs in different files and then include them in the docs 1st file? I've tried but haven't been able to figure how to do it in such a way that the doctests are aware of code executed from different files. So the doctests fail.

In theory I don't see why you could not do a rst include block in the file and sphinx wont embed it.
(I've not tried myself).

> This may be the same question, but is there a way to have code execute but not shown in the rendered docs? For example, if you don't want to show a bit of code that was discussed in a previous section, can you have that code execute for the doctests but not clutter up the code block for the user to read.

Assuming you have run the doctests above the next block of code, it will use any code that has already been run. 
Most of the examples in the sunpy docs are one long broken up doctest run.

Why are you trying to segment all the code into multiple files?

DanRyanIrish:
> Assuming you have run the doctests above the next block of code, it will use any code that has already been run.
Most of the examples in the sunpy docs are one long broken up doctest run.

That's what I assumed.  But if in a subsequent code-block, you call a variable created (or package imported) in a previous code block embedded using `.. include::` or `.. literalinclude::`, the doctests fail.

> Why are you trying to segment all the code into multiple files?

Thanks for the reply @nabobalis.  Two reasons.  Firstly, as with sunpy, ndcube has different sections in different .rst files.  Otherwise the docs would be an unmanageably huge single file.  In different sections I used the same NDCube is used as an example, e.g. in the coordinates section and the slicing section.  But the doctests don't remember code that was run in other .rst files.  They are tested separately.  So to avoid having the same code snippets copied multiple times throughout the docs' code base, it would be much better to simply have that code block in it own file and included in the docs files where needed.

Secondly, because different types of functionalities are shown in different sections (e.g. coordinate transformations, slicing etc.) from where the NDCube was instantiated, it might be helpful to have the option of showing how to instantiate the NDCube that's being used nearby.  This would help users to be clear as to why the result is what it is and also to easily copy and paster the example from one place, rather than having to jump between sections.  Ideally, it would be possible for the instantiation code to be hidden and revealed via a toggle so the doc isn't too cluttered.

nabobalis:
> > Assuming you have run the doctests above the next block of code, it will use any code that has already been run.
> > Most of the examples in the sunpy docs are one long broken up doctest run.
> 
> That's what I assumed. But if in a subsequent code-block, you call a variable created (or package imported) in a previous code block embedded using `.. include::` or `.. literalinclude::`, the doctests fail.

That would imply the include while rendered is not ran by doctest? 
I  assume that something like:
```
.. doctest::
	..include ::
	etc
```
won't work. 
Never tried to nest sphinx directives before.
 
> > Why are you trying to segment all the code into multiple files?
> 
> Thanks for the reply @nabobalis. Two reasons. Firstly, as with sunpy, ndcube has different sections in different .rst files. Otherwise the docs would be an unmanageably huge single file. In different sections I used the same NDCube is used as an example, e.g. in the coordinates section and the slicing section. But the doctests don't remember code that was run in other .rst files. They are tested separately. So to avoid having the same code snippets copied multiple times throughout the docs' code base, it would be much better to simply have that code block in it own file and included in the docs files where needed.
>
> Secondly, because different types of functionalities are shown in different sections (e.g. coordinate transformations, slicing etc.) from where the NDCube was instantiated, it might be helpful to have the option of showing how to instantiate the NDCube that's being used nearby. This would help users to be clear as to why the result is what it is and also to easily copy and paster the example from one place, rather than having to jump between sections. Ideally, it would be possible for the instantiation code to be hidden and revealed via a toggle so the doc isn't too cluttered.

All fair points, I am not personally not aware this is possible with doctest, but maybe @cadair knows.

DanRyanIrish:
@nabobalis, @Cadair: So to clarify this issue a little, I've discovered that the "included" text from other files is run by doctest.  However, the variables coming out of that code are not available to subsequent code blocks in other files.  For example:

main_doc_file.rst
```
Here is an explanation of ``range``.  Now some code.

.. include:: code_example_1.rst

Let's index that.

.. code-block:: python

  >>> print(var[1])
  1
```
code_example_1.rst
```
.. code-block:: python

  >>> var = list(range(3))
  >>> print(var)
  [0, 1, 2]
```
This will fail with doctest saying `var undefined` when it reaches the `print(var[1])` line in the 2nd code block.  So doctest isn't remembering that `var` was generated by the code in `code_example_1.rst`

DanRyanIrish:
Further context: If I change `code_example_1.rst` so it should break, e.g.
```
.. code-block:: python

  >>> var = list(range(3))
  >>> print(a)
  
```
The doctest will fail and say `a` is undefined.  So the code in other files are being run.  They just don't share the same namespace.  That's what I'd like to know how/if it's possible to get around.

docs/installation.rst
```diff
@@ -1,8 +1,10 @@
...
-Installation
-============
+.. _installation:
 
-ndcube requires Python 3.5+, SunPy 0.9+, astropy and matplotlib.
+=================
+Installing ndcube
+=================
+
+ndcube requires Python 3.5+, astropy 4.2+, and matplotlib.
```
Cadair, line 7:
```suggestion
ndcube requires Python 3.7+, astropy 4.2+, and matplotlib.
```

docs/installation.rst
```diff
@@ -1,8 +1,10 @@
...
 
 Installing the Stable Version
 -----------------------------
```
Cadair, line 10: We should follow core's lead and recommend miniconda over anaconda below.

docs/coordinates.rst
```diff
@@ -0,0 +1,423 @@
...
+  >>> wcs = astropy.wcs.WCS(naxis=3)
+  >>> wcs.wcs.ctype = 'WAVE', 'HPLT-TAN', 'HPLN-TAN'
+  >>> wcs.wcs.cunit = 'Angstrom', 'deg', 'deg'
+  >>> wcs.wcs.cdelt = 0.2, 0.5, 0.4
+  >>> wcs.wcs.crpix = 0, 2, 2
+  >>> wcs.wcs.crval = 10, 0.5, 1
+
+.. _extra_coords:
+
+ExtraCoords
```
Cadair, line 27: I feel like this page should start with a dummies guide to WCS?

DanRyanIrish:
Thanks @Cadair.  Except for your request for a dummies guide to WCS, I declare this PR done! (Pending @hayesla taking another look.)

hayesla:
this looks really good @DanRyanIrish ! I think the schematic figures are really helpful

hayesla (approved):

hayesla:
> Thanks @Cadair. Except for your request for a dummies guide to WCS, I declare this PR done! (Pending @hayesla taking another look.)

oh yes a dummies guide to WCS would be class!

Cadair:
> Thanks @Cadair. Except for your request for a dummies guide to WCS, I declare this PR done! (Pending @hayesla taking another look.)

Ok I will give it a proper read through when I get a chance.

docs/coordinates.rst
```diff
@@ -0,0 +1,444 @@
...
+==========================
+Coordinate Transformations
+==========================
+
+Introduction to WCS
+===================
+
+To describe the mapping between array elements/pixels and real world coordinates, ndcube heavily leverages the World Coordinate System (WCS) framework, specifically the tools written by AstroPy that implement this framework in Python.
+WCS allows a wide variety of projections, rotations and transformations be stored and executed.
+Because it allows coordinates transformations to be stored functionally, rather than in memory-heavy lookup tables, and because it caters for both astronomy-specific coordinate systems (e.g. RA & Dec.) as well as simpler, more common ones (e.g. wavelength), WCS has become the most common coordinate transformation framework in astronomy.
```
hayesla, line 12: silly question - coordinate transforms being wavelength?

docs/coordinates.rst
```diff
@@ -0,0 +1,444 @@
...
+  >>> my_2d_cube.global_coords.physical_types['em.wl']  # doctest: +SKIP
+  'em.wl'  # doctest: +SKIP
+  >>> my_2d_cube.global_coords['em.wl']  # doctest: +SKIP
+  <SpectralCoord 1e-9 m>
+
+.. _cube_coordinates:
+
+NDCube Coordinates
+==================
+WCS objects are a powerful and concise way of storing complex functional coordinate transformations.
```
hayesla, line 181: is this astropy WCS objects yeah?

docs/data_classes.rst
```diff
@@ -0,0 +1,427 @@
...
+
+.. _ndcube:
+
+NDCube
+======
+`~ndcube.NDCube` is the primary data class the ndcube package.
+It's designed to manage a single data array and set of WCS transformations.
+`~ndcube.NDCube` provides unified slicing, visualization, coordinate transformation and self-inspection APIs which are independent of the number and physical types of axes.
+It can therefore be used for any type of data (e.g. images, spectra, timeseries, etc.) so long as those data are represented by an object that behaves like a `numpy.ndarray` and the coordinates by an object that adheres to the AstroPy `WCS API (APE 14) <https://docs.astropy.org/en/stable/wcs/wcsapi.html>`_.
+This makes `~ndcube.NDCube` an powerful base class when developing tools for specific data types.
```
hayesla, line 18:
```suggestion
This makes `~ndcube.NDCube` a powerful base class when developing tools for specific data types.
```

docs/data_classes.rst
```diff
@@ -0,0 +1,427 @@
...
+Yellow ovals represent methods for inspecting, visualizing, and analyzing the `~ndcube.NDCube`.
+
+.. image:: images/ndcube_diagram.png
+  :width: 800
+  :alt: Components of an NDCube
+
+
+Initializing an NDCube
+----------------------
+To initialize the most basic `~ndcube.NDCube` object, we need is a `numpy.ndarray`-like array containing the data and an APE-14-compliant WCS object (e.g. `astropy.wcs.WCS`) describing the coordinate transformations to and from array-elements.
```
hayesla, line 42:
```suggestion
To initialize the most basic `~ndcube.NDCube` object, what we need is a `numpy.ndarray`-like array containing the data and an APE-14-compliant WCS object (e.g. `astropy.wcs.WCS`) describing the coordinate transformations to and from array-elements.
```

docs/data_classes.rst
```diff
@@ -0,0 +1,427 @@
...
+In that case the cubes are not only ordered in time, but also along one of their spatial axes.
+Another way of saying this is that the sequence axis is parallel to one of the cubes' axes.
+This cube axis is known as the "common axis".
+Let's say in our example that the common axis is the 2nd axis.
+Thus, we can also treat the data set as if it were a single image with a shape of ``(512, 2048)``.
+See panel b) of the figure above.
+
+Setting a common axis is optional and does not force the user to interact with the data as if it were in configuration b).
+Instead `~ndcube.NDCubeSequence` has different versions of its methods whose names are prefixed with ``cube_like`` that account for the common axis and equivalent non-cube-like methods that do not.
+This allows users to switch back and forth between configurations a) and b) as their use case demands without the requiring the user to have two versions of the same data.
```
hayesla, line 168:
```suggestion
This allows users to switch back and forth between configurations a) and b) as their use case demands without the requiring of the user to have two versions of the same data.
```

docs/data_classes.rst
```diff
@@ -0,0 +1,427 @@
...
+
+  >>> my_sequence.cube_like_array_axis_physical_types
+  [('custom:pos.helioprojective.lat', 'custom:pos.helioprojective.lon'),
+   ('custom:pos.helioprojective.lat', 'custom:pos.helioprojective.lon'),
+   ('em.wl',)]
+
+Explode Along Axis
+------------------
+During analysis of some data --- say of a stack of images --- it may be necessary to make some different fine-pointing adjustments to each image that isn't accounted for the in the original WCS translations, e.g. due to satellite wobble.
+If these changes are not describable with a single WCS object, it may be desirable to break up the N-D sub-cubes of an `~ndcube.NDCubeSequence` into an sequence of sub-cubes with dimension N-1.
```
hayesla, line 272:
```suggestion
If these changes are not describable with a single WCS object, it may be desirable to break up the N-D sub-cubes of an `~ndcube.NDCubeSequence` into a sequence of sub-cubes with dimension N-1.
```

docs/data_classes.rst
```diff
@@ -0,0 +1,427 @@
...
+Note that an `~ndcube.NDCubeSequence` can be exploded along any axis.  A common axis need not be defined.
+
+To learn how to slice `~ndcube.NDCubeSequence` instances and manipulate sequence coordinates, see the :ref:`sequence_slicing` and :ref:`sequence_coordinates` sections.
+
+.. _ndcollection:
+
+NDCollection
+============
+`~ndcube.NDCollection` is a container class for grouping `~ndcube.NDCube` or `~ndcube.NDCubeSequence` instances in an unordered way.
+`~ndcube.NDCollection` therefore is differs from `~ndcube.NDCubeSequence` in that the objects contained are not considered to be in any order, are not assumed to represent measurements of the same physical property, and they can have different dimensionalities.
```
hayesla, line 301:
```suggestion
`~ndcube.NDCollection` therefore differs from `~ndcube.NDCubeSequence` in that the objects contained are not considered to be in any order, are not assumed to represent measurements of the same physical property, and they can have different dimensionalities.
```

docs/data_classes.rst
```diff
@@ -0,0 +1,427 @@
...
+And just like a `dict` we can see the different names available using the ``keys`` method:
+
+.. code-block:: python
+
+  >>> my_collection.keys()
+  dict_keys(['observations', 'linewidths'])
+
+Aligned Axes
+------------
+In the above example, the linewidth object's axes are aligned with the first two axes of observations object.
```
hayesla, line 349:
```suggestion
In the above example, the linewidth object's axes are aligned with the first two axes of the observations object.
```

docs/data_classes.rst
```diff
@@ -0,0 +1,427 @@
...
+
+We can see which axes are aligned by inpecting the ``aligned_axes`` attribute:
+
+.. code-block:: python
+
+  >>> my_collection.aligned_axes
+  {'observations': (0, 1), 'linewidths': (0, 1)}
+
+This gives us the array axes for each ND object separately.
+We should read this as array axis 0 of ``observations`` is aligned with the array axis 0 of ``'linewidths'``, as so on.
```
hayesla, line 369:
```suggestion
We should read this as array axis 0 of ``observations`` is aligned with the array axis 0 of ``'linewidths'``, and so on.
```

docs/data_classes.rst
```diff
@@ -0,0 +1,427 @@
...
+We can also see the physical properties to which the aligned axes correspond by using the `~ndcube.NDCollection.aligned_axis_physical_types` property.
+
+.. code-block:: python
+
+  >>> my_collection.aligned_axis_physical_types  # doctest: +SKIP
+  [('custom:pos.helioprojective.lon', 'custom:pos.helioprojective.lat'), ('custom:pos.helioprojective.lon', 'custom:pos.helioprojective.lat')]
+
+This returns a `list` of `tuple` in array axis order giving the physical types that correspond to each aligned axis.
+For each aligned axis, only physical types associated with all the cubes in the collection are returned.
+Note that there is no there is no requirement that all aligned axes must represent the same physical types.
```
hayesla, line 416:
```suggestion
Note that there is no requirement that all aligned axes must represent the same physical types.
```

docs/data_classes.rst
```diff
@@ -0,0 +1,427 @@
...
+They just have to be the same length.
+Therefore, is it possible that this property returns no physical types.
```
hayesla, line 418:
```suggestion
Therefore, it is possible that this property returns no physical types.
```

docs/introduction.rst
```diff
@@ -3,49 +3,86 @@ An Introduction to ndcube
...
-The fundamental reason to opt for ndcube is a requirement to harness the astronomy-specific World Coordinate System (WCS) tooling provided by packages like Astropy (which now also supports gWCS).
-The data model of xarray centered on the requirements and conventions of the geosciences, which while being very similar to that of astronomy in conception, is sufficently different in construction to cause much friction.
-Utilize the astropy WCS infrastructure enables us to directly read the most common file formats in astronomy (FITS), although the FITS WCS data model is also commonly used outside of FITS files.
+===========
+It's worth addressing the role ndcube plays within the scientific Python ecosystem and why it exists separately from its most similar package, xarray.
+The fundamental reason to opt for ndcube is to harness the astronomy-specific World Coordinate System (WCS).
+The data model of xarray centers on the requirements and conventions of the geosciences.
+Although very similar to those of astronomy in conception, they are sufficently different in construction to cause significant friction.
```
hayesla, line 62: I dont know why, but friction here sounds like theres been an argument between geosciences and astronomy :P - would `challenges` work better?

docs/slicing.rst
```diff
@@ -0,0 +1,360 @@
+.. _slicing:
+
+==================
+Slicing ND Objects
+==================
+Arguably the most useful feature ndcube provides is the slicing of its data classes.
+Users can apply the standard slicing notation to ND objects (or real world coordinates in the case of `ndcube.NDCube.crop`) which alter the data and WCS transformations consistently and simultaneously.
+This enables users to rapidly and reliably identify and extract regions of interest in their data, thereby allowing them to move closer to the speed of thought during their analysis.
```
hayesla, line 8: "speed of thought" :P I want to be able to do this

docs/slicing.rst
```diff
@@ -0,0 +1,360 @@
...
+
+.. code-block:: python
+
+  >>> my_collection['observations']  # doctest: +SKIP
+
+However, also recall that we can mark axes of the member ND objects that are aligned.
+The value in this is that it enables users to slice all the members of the collection simultaneously from the `~ndcube.NDCollection` level.
+This can only be done for aligned axes.
+Non-aligned axes must be sliced separately.
+Nonethless, `~ndcube.NDCollection`'s slicing capability represents one of its greatest advantages over a simple Python `dict`, making it a powerful tool for rapidly and reliably cropping multiple components of a data set to a region of interest.
```
hayesla, line 309:
```suggestion
Nonetheless, `~ndcube.NDCollection`'s slicing capability represents one of its greatest advantages over a simple Python `dict`, making it a powerful tool for rapidly and reliably cropping multiple components of a data set to a region of interest.
```

docs/visualization.rst
```diff
@@ -0,0 +1,332 @@
...
+For for data with two array axes, an image is produced similar to that of `matplotlib.pyplot.imshow`.
+
+.. plot::
+  :context:
+  :include-source:
+
+  >>> ax = my_cube[0].plot()
+  >>> plt.show()
+
+Note that the bottom row is of the image is empty because those array elements are marked ``True`` is ``.mask``.
```
hayesla, line 72:
```suggestion
Note that the bottom row of the image is empty because those array elements are marked ``True`` is ``.mask``.
```

docs/visualization.rst
```diff
@@ -0,0 +1,332 @@
...
+  >>> plt.clf()
+  >>> plt.cla()
+  >>> plt.close()
+
+Visualizing NDCubeSequences
+===========================
+Since ndcube 2.0, the `~ndcube.NDCubeSequence` visualization support has been dropped.
+The rationale for this is outlined in `Issue #321 <https://github.com/sunpy/ndcube/issues/321>`_ on the ndcube GitHub repo.
+If you feel that `~ndcube.NDCubeSequence` visualization should be supported again, please let us know by commenting on that issue and telling us of your use case.
+Better still, if you would like to work on the infrastructure required to support `~ndcube.NDCubeSequence` visualization is a post ndcube 2.0 world let us know by commenting on the issue.
```
hayesla, line 163:
```suggestion
Better still, if you would like to work on the infrastructure required to support `~ndcube.NDCubeSequence` visualization in a post ndcube 2.0 world let us know by commenting on the issue.
```

docs/visualization.rst
```diff
@@ -0,0 +1,332 @@
...
+Despite this the lack of `~ndcube.NDCubeSequence` visualization support, you can still visualize the data in `~ndcube.NDCubeSequence` in a number of ways.
+You can slice out a single `~ndcube.NDCube` and use its `~ndcube.NDCube.plot` method.
+You can extract the data and use the myriad of plotting packages available in the Python ecosystem.
+Finally, if you want to be advanced, you can write your own mixin class to define the plotting methods.
+Below, we will outline these latter two options in a little more detail.
+
+Extracting and Plotting NDCubeSequence Data with Matplotlib
+-----------------------------------------------------------
+In order to produce plots (or perform other analysis) outside of the ``ndcube`` framework, it may be useful to extract the data from the `~ndcube.NDCubeSequence` into single `~numpy.ndarray` instances.
+Let's first define an `~ndcube.NDCubeSequence` with a common axis of 0 and and time as an extra coord stretching across the cube along the common axis.
```
hayesla, line 174:
```suggestion
Let's first define an `~ndcube.NDCubeSequence` with a common axis of 0 and time as an extra coord stretching across the cube along the common axis.
```

docs/visualization.rst
```diff
@@ -0,0 +1,332 @@
...
+    >>> ax.set_xlim(times.mjd[0], times.mjd[-1])  # doctest: +SKIP
+    >>> ax.set_xlabel("Time [Modified Julian Day]")  # doctest: +SKIP
+    >>> ax.set_ylim(wavelength[0].value, wavelength[-1].value)  # doctest: +SKIP
+    >>> ax.set_ylabel(f"Wavelength [{wavelength.unit}]")  # doctest: +SKIP
+    >>> plt.show()  # doctest: +SKIP
+
+Now let's say we want to animate our data, for example, show how the intensity changes over wavelength and time.
+For this we can use `~ndcube.visualization.animator.ImageAnimator`.
+This class is not well suited to displaying the complex relationship between coordinates that we are used to with `~astropy.visualization.wcsaxes.WCSAxes`.
+For example, non-linear coordinates non-independent coordinates.
```
hayesla, line 283:
```suggestion
For example, non-linear coordinates and non-independent coordinates.
```

hayesla (approved):

hayesla:
I think these docs look really good - great work @DanRyanIrish ! I'd definitely suggest merging these once @Cadair is happy and if there is anything more outstanding this should be put in a separate PR (I can't take reading through it all again - jkjk ;))

The inclusion of the figures and examples is really good!  🚀 2.0!

Cadair:
I will give it a read and hit the button. Thanks a lot for reviewing it @hayesla

docs/introduction.rst
DanRyanIrish: Ha!  Oh I see why you made a comment on this before now.  I thought friction described it better because "challenge" makes me think of a more philosophical problem that hasn't been solved.  But now I see where you're coming from I'm not sure.

DanRyanIrish:
Thanks so much @hayesla for wading through this enormous PR...twice...at least!  Really helpful feedback, especially the suggestions about plots and schematics.  I agree that if are still areas where improvements can be made, the docs now closely reflect the 2.0 features and any further changes should be handled in smaller, more manageable subsequent PRs.  Thanks again!

docs/coordinates.rst
```diff
@@ -0,0 +1,444 @@
...
+
+The foundation of the AstroPy WCS implementation is the `~astropy.wcs.WCS` object, which stores the critical information describing the coordinate transformations (e.g. the reference pixel and its corresponding coordinate values, ``crpix`` and ``crval``, and the projection type, ``ctype``).
+It also executes these transformations via methods like `~astropy.wcs.WCS.world_to_pixel` and `~astropy.wcs.WCS.world_to_pixel` which convert between pixel indices and world coordinate values.
+However, these methods are independent of the data array and the `~astropy.wcs.WCS` object carries little or no information about the data itself.
+That is why the ndcube package is needed.
+Nonetheless, AstroPy's WCS implementation is a crucial pillar of ndcube, as is the more generalized offshoot `gWCS <https://gwcs.readthedocs.io/en/stable/>`_ which provides greater generalization for the loss of some performance.
```
Cadair, line 18:
```suggestion
Nonetheless, Astropy's WCS implementation is a crucial pillar of ndcube, as is the more generalized offshoot `gWCS <https://gwcs.readthedocs.io/en/stable/>`_ which provides greater generalization outside of the FITS data model.
```

docs/coordinates.rst
```diff
@@ -0,0 +1,444 @@
...
+Crucially though for ndcube, both implementations adhere to the `AstroPy WCS API (APE 14) <https://docs.astropy.org/en/stable/wcs/wcsapi.html>`_.
```
Cadair, line 19:
```suggestion
Crucially though for ndcube, both implementations adhere to the `Astropy WCS API <https://docs.astropy.org/en/stable/wcs/wcsapi.html>`_.
```

docs/coordinates.rst
```diff
@@ -0,0 +1,444 @@
...
+A familiarity with WCS and the AstroPy and gWCS Python implementations will be helpful (although hopefully not essential) in understanding this guide.
```
Cadair, line 20:
```suggestion
A familiarity with WCS and the Astropy and gWCS Python implementations will be helpful (although hopefully not essential) in understanding this guide.
```

docs/coordinates.rst
```diff
@@ -0,0 +1,444 @@
...
+Now let's create an `~ndcube.ExtraCoords` instance and add our time extra coordinate to it.
+To do this we need to supply the physical type of the coordinate, the array axis to which is corresponds, and the values of the coordinate.
+The number of values should equal the length of the axis and the physical type must be a valid `IVOA UCD1+ controlled words <http://www.ivoa.net/documents/REC/UCD/UCDlist-20070402.html>`_ word.
+If one does not exist for your coordinate, prepend the type with ``custom:``.
+
+.. code-block:: python
+
+  >>> from ndcube import ExtraCoords
+  >>> my_extra_coords = ExtraCoords()
+  >>> my_extra_coords.add_coordinate('time', (2,), timestamps)  # TO DO: Change the mapping to 0 Issue #342 resolved.
```
Cadair, line 73: Also open an issue for this, or add a comment to #342 or we will forget.

```suggestion
  >>> my_extra_coords.add_coordinate('time', (2,), timestamps)  # TODO: Change the mapping to 0 Issue #342 resolved.
```

docs/coordinates.rst
```diff
@@ -0,0 +1,444 @@
...
+  >>> list(my_cube.global_coords.values())  # Returns a list of coordinate values
+  [<Quantity 1. m>]
+  >>> list(my_cube.global_coords.items())  # Returns a list of (name, value) pairs
+  [('distance', <Quantity 1. m>)]
+
+A common use case for `~ndcube.GlobalCoords` is associated with slicing (:ref:`cube_slicing`).
+In addition to tracking and updating the `~ndcube.NDCube.wcs` and `~ndcube.NDCube.extra_coords` objects, `~ndcube.NDCube`'s slicing infrastucture also identifies when the array axes to which a coordinate corresponds are dropped.
+The values of dropped coordinates at the position where the `~ndcube.NDCube` was sliced are stored in the `astropy.wcs.WCS` instance from where `~ndcube.GlobalCoords` can access and return them.
+
+.. code-block:: python
```
Cadair, line 162: if only this example actually worked :stuck_out_tongue:

docs/coordinates.rst
```diff
@@ -0,0 +1,444 @@
...
+  >>> list(my_2d_cube.global_coords.keys())  # doctest: +SKIP
+  ['distance', 'em.wl']  # doctest: +SKIP
+  >>> my_2d_cube.global_coords.physical_types['em.wl']  # doctest: +SKIP
+  'em.wl'  # doctest: +SKIP
+  >>> my_2d_cube.global_coords['em.wl']  # doctest: +SKIP
+  <SpectralCoord 1e-9 m>
+
+.. _cube_coordinates:
+
+NDCube Coordinates
```
Cadair, line 179: Should this not be the first section in this document? It feels weird going from a WCS description to EC and GC. Would this then EC then GC be a better flow?

docs/coordinates.rst
```diff
@@ -0,0 +1,444 @@
...
+      (2159.85889149, 3.59986658e+03), (2159.74920255, 5.39950295e+03)],
+     [(3600.        , 4.56905253e-02), (3600.        , 1.80000000e+03),
+      (3600.        , 3.59995431e+03), (3600.        , 5.39963453e+03)],
+     [(5039.92178073, 4.56894119e-02), (5040.03143627, 1.79995614e+03),
+      (5040.14110851, 3.59986658e+03), (5040.25079745, 5.39950295e+03)],
+     [(6479.70323031, 4.56860725e-02), (6479.92250932, 1.79982456e+03),
+      (6480.14182173, 3.59960344e+03), (6480.36116753, 5.39910830e+03)]]>
+
+It is also possible to request more than one axis's world coordinates by setting ``axes`` to an iterable of data axis number and/or axis type strings.
+The coordinate objects are returned in world axis order in accordance with APE 14.
```
Cadair, line 229:
```suggestion
The coordinate objects are returned in world axis order.
```

docs/coordinates.rst
```diff
@@ -0,0 +1,444 @@
...
+        (4320.05190982, 2699.97783871), (4320.10675212, 4499.81838925),
+        (4320.16160277, 6299.24788597)],
+       [(5759.75658812, -899.79647591), (5759.92104888,  899.95636786),
+        (5760.08553469, 2699.84625127), (5760.25004555, 4499.59909505),
+        (5760.41458147, 6298.94094507)],
+       [(7199.36047891, -899.70880283), (7199.63452676,  899.86866585),
+        (7199.90861634, 2699.58313412), (7200.18274766, 4499.1606028 ),
+        (7200.45692072, 6298.32719784)]]>)
+
+`~ndcube.NDCube.axis_world_coords` also allows the user to pick which WCS object should be used, `ndcube.NDCube.wcs` or `ndcube.NDCube.combined_wcs` by setting the ``wcs=`` keyword.
```
Cadair, line 283: you can also just pass `ndcube.extra_coords` to get the EC tables back.

docs/slicing.rst
```diff
@@ -0,0 +1,360 @@
...
+The `~ndcube.NDCube` slicing infrastructure returns a new `~ndcube.NDCube` with all relevant infomation consistently altered.
+This includes the data, WCS transformations, `~ndcube.ExtraCoords`, uncertainty and mask.
+This is achieved by applying the standard slicing API to the `~ndcube.NDCube`.
+Because many of the inspection properties such `~ndcube.NDCube.dimensions` and `~ndcube.NDCube.array_axis_physical_types` are calculated on the fly, the information they return after slicing is also consistent with the sliced `~ndcube.NDCube`.
+This makes `~ndcube.NDCube`'s slicing infrastructure very powerful.
+
+To slice ``my_cube``, simply do something like:
+
+.. expanding-code-block:: python
+  :summary: Expand to see my_cube instantiated.
```
Cadair, line 23: I think this will work:

```suggestion
  :summary: Expand to see <span class="pre">my_cube</span> instantiated.
```

docs/slicing.rst
```diff
@@ -0,0 +1,360 @@
...
+
+Cropping with Real World Coordinates
+------------------------------------
+In addition to slicing by index, `~ndcube.NDCube` supports slicing by real world coordinates via the `~ndcube.NDCube.crop` method.
+This takes two iterables of high level coordinate objects, e.g. `~astropy.time.Time`, `~astropy.coordinates.SkyCoord`, `~astropy.coordinates.SpectralCoord`, `~astropy,units.Quantity` etc.
+Each iterable describes a single location in the data array in real world coordinates.
+The first iterable describes the lower corner of the region of interest and thus contains the lower limit of each real world coordinate.
+The second iterable represents the upper corner of the region of interest and thus contains the upper limit of each real world coordinate.
+The crop method indentifies the smallest rectangular region in the data array that contains both the lower and upper limits in all the real world coordinates, and crops the `~ndcube.NDCube` to that region.
+It does not rebin or interpolate the data.  The order of the high level coordinate objects in each iterable must be the same as that expected by `astropy.wcs.WCS.world_to_array_index`, namely in world order.
```
Cadair, line 108:
```suggestion
It does not rebin or interpolate the data.  The order of the high level coordinate objects in each iterable must be the same as that expected by `astropy.wcs.wcsapi.BaseHighLevelWCS.world_to_array_index`, namely in world order.
```

docs/slicing.rst
```diff
@@ -0,0 +1,360 @@
...
+.. _sequence_slicing:
+
+Slicing NDCubeSequences
+=======================
+As with `~ndcube.NDCube`, `~ndcube.NDCubeSequence` is sliced by applying the standard slicing API.
+The `~ndcube.NDCubeSequence` slicing infrastrcuture determines which cubes should be kept based on the sequence axis input, then passes the rest of the slicing off to desired NDCubes.
+Thus the data arrays, WCS transformations, masks, uncertainty arrays, and extra coordinates are all altered accorindingly in each relevant sub-cube.
+Say we have three NDCubes in an `~ndcube.NDCubeSequence`, each of shape ``(4, 4, 5)``.
+
+.. expanding-code-block:: python
```
Cadair, line 131: Would it not be better if this were moved to just above the code block where it is used?

docs/slicing.rst
```diff
@@ -0,0 +1,360 @@
...
+Notice that since the sub-cubes are now of different lengths along the common axis, the corresponding `~astropy.units.Quantity` gives the
+lengths of each cube individually.
+
+.. _collection_slicing:
+
+Slicing NDCollections
+=====================
+Recall from the :ref:`ndcollection` section that members of an `~ndcube.NDCollection` can be accessed by slicing it with a string giving the member's name.
+
+.. code-block:: python
```
Cadair, line 301: Should we not have a hidden setup block here as well?

docs/visualization.rst
```diff
@@ -0,0 +1,332 @@
...
+For for data with two array axes, an image is produced similar to that of `matplotlib.pyplot.imshow`.
+
+.. plot::
+  :context:
+  :include-source:
+
+  >>> ax = my_cube[0].plot()
+  >>> plt.show()
+
+Note that the bottom row of the image is empty because those array elements are marked ``True`` is ``.mask``.
```
Cadair, line 72: it isn't empty??

docs/visualization.rst
```diff
@@ -0,0 +1,332 @@
...
+
+.. plot::
+  :nofigs:
+
+  >>> plt.clf()
+  >>> plt.cla()
+  >>> plt.close()
+
+No args are required.
+The necessary information to generate the plot is derived from the data and metadata in the `~ndcube.NDCube`.
```
Cadair, line 101:
```suggestion
Note how no arguments are required, the necessary information to generate the plot is derived from the data and metadata in the `~ndcube.NDCube`.
```

docs/visualization.rst
```diff
@@ -0,0 +1,332 @@
...
+However optional keywords enable customization of the visualization.
+For `~ndcube.NDCube` instances with more than one array axis, the ``plot_axes`` keyword is used to determine which array axes are displayed on which plot axes.It is set to a list with a length equal to the number of array axes.
```
Cadair, line 103:
```suggestion
For `~ndcube.NDCube` instances with more than one array axis, the ``plot_axes`` keyword is used to determine which array axes are displayed on which plot axes.
It is set to a list with a length equal to the number of array axes.
```

docs/visualization.rst
```diff
@@ -0,0 +1,332 @@
...
+
+  >>> plt.clf()
+  >>> plt.cla()
+  >>> plt.close()
+
+`~ndcube.NDCube.plot` uses `~astropy.visualization.wcsaxes.WCSAxes` to produce all plots.
+This enables a rigorous representation of the coordinates on the plot, including those that are not aligned to the pixel grid.
+It also enables the coordinates along the plot axes to be updated between frames of an animation.
+`ndcube.NDCube.plot` therefore allows users to decide which WCS object to use, either `~ndcube.NDCube.wcs` or `~ndcube.NDCube.combined_wcs` which also includes the `~ndcube.ExtraCoords`.
+In principle, another third-part WCS can be used so long as it is a valid description of all array axes.
```
Cadair, line 142:
```suggestion
```

Cadair (commented):
This is really great. It reads really well.

We should keep open an issue after this is merged to update the API docs as they are very ropey right now.

docs/coordinates.rst
DanRyanIrish: Indeed!

docs/coordinates.rst
DanRyanIrish: Possibly.  Will give it some though before merging.

docs/data_classes.rst
```diff
@@ -0,0 +1,427 @@
+.. _data_classes:
+
+==========
+ND Objects
```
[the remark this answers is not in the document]
DanRyanIrish: Suggest an alternative?

docs/data_classes.rst
```diff
@@ -0,0 +1,427 @@
...
+   ('custom:pos.helioprojective.lat', 'custom:pos.helioprojective.lon'),
+   ('em.wl',)]
+
+Explode Along Axis
+------------------
+During analysis of some data --- say of a stack of images --- it may be necessary to make some different fine-pointing adjustments to each image that isn't accounted for the in the original WCS translations, e.g. due to satellite wobble.
+If these changes are not describable with a single WCS object, it may be desirable to break up the N-D sub-cubes of an `~ndcube.NDCubeSequence` into a sequence of sub-cubes with dimension N-1.
+This would enable a separate WCS object to be associated with each image and hence allow individual pointing adjustments.
+
+Rather than manually dividing the cubes up and deriving the corresponding WCS object for each exposure, `~ndcube.NDCubeSequence` provides a useful method, `~ndcube.NDCubeSequence.explode_along_axis`.
```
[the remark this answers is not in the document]
DanRyanIrish: You're right!  Although `NDCubeSequence` has it too.

docs/introduction.rst
```diff
@@ -3,49 +3,86 @@ An Introduction to ndcube
...
+Therefore we use the term "array axes" to refer to pixel axes which have been reversed to reflect the axis order of the numpy data array.
+Take for example a numpy array with three dimensions.
+Since the array axes are simply the reverse of the pixel axes, the first axis of the array corresponds to the 3rd pixel axis.
+And the 2nd array axis corresponds to the 2nd pixel axis.
+If the array had four axes, the first array axis would correspond to the fourth pixel axis and the second array axis would correspond to the third pixel axis and so on.
+
+In ndcube, inputs and outputs are given in either array axis order or world axis order, depending on the types of information.
+Throughout these docs and in the docstrings we will endeavor to highlight which order is relevant.
+However a good rule of thumb is that if you are using a sequence of coordinate objects to describe locations in the data cube -- for example in the input of `ndcube.NDCube.crop` or the output of `ndcube.NDCube.axis_world_coords` -- they should be in world axis order.
+In almost all other cases, array axis order is used.
```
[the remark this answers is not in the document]
DanRyanIrish: In the public API ndcube does not accept or return "pixel" axes or pixel order.  That is an under-the-hood implementation detail.

docs/visualization.rst
DanRyanIrish: No, I've defined the mask in the expanding code block above to demonstrate its affect.

docs/data_classes.rst
```diff
@@ -0,0 +1,427 @@
...
+Instead `~ndcube.NDCubeSequence` has different versions of its methods whose names are prefixed with ``cube_like`` that account for the common axis and equivalent non-cube-like methods that do not.
+This allows users to switch back and forth between configurations a) and b) as their use case demands without requiring the user to have two versions of the same data.
+This flexibility makes `~ndcube.NDCubeSequence` a powerful tool when handling complex N-D dimensional data described by different but comparable coordinate transformations.
+
+Initializing an NDCubeSequence
+------------------------------
+To initialize the most basic `~ndcube.NDCubeSequence`, all you need is a list of `~ndcube.NDCube` instances.
+Let's first define four 3-D NDCubes.
+
+.. code-block:: python
```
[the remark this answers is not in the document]
DanRyanIrish: Yes!

docs/coordinates.rst
```diff
@@ -0,0 +1,444 @@
...
+==========================
+
+Introduction to WCS
+===================
+
+To describe the mapping between array elements/pixels and real world coordinates, ndcube heavily leverages the World Coordinate System (WCS) framework, specifically the tools written by Astropy that implement this framework in Python.
+WCS allows a wide variety of projections, rotations and transformations be stored and executed.
+Because it allows coordinates transformations to be stored functionally, rather than in memory-heavy lookup tables, and because it caters for both astronomy-specific coordinate systems (e.g. RA & Dec.) as well as simpler, more common ones (e.g. wavelength), WCS has become the most common coordinate transformation framework in astronomy.
+
+The foundation of the Astropy WCS implementation is the `~astropy.wcs.WCS` object, which stores the critical information describing the coordinate transformations (e.g. the reference pixel and its corresponding coordinate values, ``crpix`` and ``crval``, and the projection type, ``ctype``).
```
DanRyanIrish, line 14:
```suggestion
The most commonly used WCS implementation in Python is the `~astropy.wcs.WCS` object, which stores the critical information describing the coordinate transformations as described by the FITS data model (e.g. the reference pixel and its corresponding coordinate values, ``CRPIX`` and ``CRVAL``, and the projection type, ``CTYPE`` etc.).
```

docs/data_classes.rst
```diff
@@ -0,0 +1,427 @@
...
+
+  >>> my_collection.aligned_axes
+  {'observations': (0, 1), 'linewidths': (0, 1)}
+
+This gives us the array axes for each ND object separately.
+We should read this as array axis 0 of ``observations`` is aligned with the array axis 0 of ``'linewidths'``, and so on.
+However, the mapping can be more complicated.
+Let's say we reversed the axes of our ``linewidths`` ND object for some reason:
```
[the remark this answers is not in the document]
DanRyanIrish: I think it's fine.

docs/data_classes.rst
```diff
@@ -0,0 +1,427 @@
...
+  >>> mask[0, 0][:] = True  # Now mask some values.
+  >>> # Define uncertainty, metadata and unit.
+  >>> uncertainty = StdDevUncertainty(np.sqrt(np.abs(data)))
+  >>> meta = {"Description": "This is example NDCube metadata."}
+  >>> unit = u.ct
+
+  >>> # Instantiate NDCube with supporting data.
+  >>> my_cube = NDCube(data, wcs=wcs, uncertainty=uncertainty, mask=mask, meta=meta, unit=unit)
+
+Generating `~ndcube.ExtraCoords` and `~ndcube.GlobalCoords` objects and attaching them to your `~ndcube.NDCube` is demonstrated in the :ref:`extra_coords` and :ref:`global_coords` sections.
```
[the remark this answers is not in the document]
Cadair: This is certainly true for GC, less for EC I guess.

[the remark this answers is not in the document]
DanRyanIrish: This would be true for EC too did NDCube was instantiated with an empty EC object by default.  Unless there's an issue with empty EC coord objects or a performance consideration, we should instantiated NDCube with an empty EC object for consistency and ease-of-use.

docs/coordinates.rst
DanRyanIrish: Currently that doesn't work.

docs/getting_help.rst
```diff
@@ -33,16 +33,10 @@ feature, please use the developer mailing list.
...
 ---------
 
-SunPy has a `chat room`_ that uses `Matrix`_ which will open directly in
-your browser. Stop by and say hello. Talk with the SunPy and ndcube
-users and developers and get started.
-
-If you are using IRC hosted on `freenode`_, then you can join the same
-channel (bridged into matrix) by joining the (#sunpy) channel on
-freenode.
+SunPy has a `element.io channel`_ which will open directly in your browser.
```
Cadair, line 36:
```suggestion
SunPy has a `matrix channel`_ which will open directly in your browser.
```

docs/getting_help.rst
```diff
@@ -33,16 +33,10 @@ feature, please use the developer mailing list.
...
+Stop by and say hello.
+Talk with the SunPy and ndcube users and developers and get started.
 
 .. _`general mailing list`: https://groups.google.com/forum/#!forum/sunpy
 .. _`developer mailing list`: https://groups.google.com/forum/#!forum/sunpy-dev
-.. _`chat room`: https://riot.im/app/#/room/#sunpy:openastronomy.org
-.. _`Matrix`: https://matrix.org/blog/home/
-.. _`freenode`: https://freenode.net/
+.. _`element.io channel`: https://openastronomy.element.io/#/room/#sunpy:openastronomy.org
```
Cadair, line 42:
```suggestion
.. _`matrix channel`: https://openastronomy.element.io/#/room/#sunpy:openastronomy.org
```

Cadair (approved):

docs/data_classes.rst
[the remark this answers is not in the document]
DanRyanIrish: Going to leave this as is for this PR.  It certainly be changed in another PR if we come up with a better name.

PR #360, opened 2020-12-03
Include inherited attributes in API docs.

### LINKED ISSUE #289: Documentation for ndcube 2

DanRyanIrish:
### Description

Fixes #289

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -4,6 +4,7 @@ Reference/API
 =============
 
 .. automodapi:: ndcube
+   :inherited-members:
 
 .. automodapi:: ndcube.mixins
 


DanRyanIrish:
This PR raises the question whether `NDCube` should be inheriting from `NDArithmeticMixin` at this time since we don't implement any of those methods...  It also highlights that the docstrings from ABCs don't seem to be carrying over to methods or classes that inherit them.

PR #363, opened 2020-12-03
Add missing docstrings to public attributes

### LINKED ISSUE #362: Add docstrings to all public attributes
### Description

Many public attributes are lacking docstrings.  This should be addressed before 2.0.  In addition, all public facing classes should have class level docstrings.  Classes do not inherit docstrings from parent classes or ABCs as they should perform slightly different roles.  Therefore they must be duplicated and edited if necessary.

DanRyanIrish:
### Description

This PR adds docstrings to all public attributes for `NDCubeSequence` and `NDCollection`.  `NDCube` doctrings will be improved after the `NDCubeABC` is finalized after community consultation.

This PR also restricts the output of `NDCubeSequence.sequence_axis_coords` to coords common to all cubes in sequence.

Fixes #362


Cadair (approved):
Only gave it a quick glance but seems fine.

Issue #366, opened 2020-12-09
HECR not recongnised as a physical type


dstansby:
### Description
Thompson 2006 says I should be able to use `HECR` as a CTYPE to denote "Radial distance from sun center". However, ndcube doesn't seem to recognise this as a physical type:

```python
import astropy.wcs
import numpy as np

from ndcube import NDCube

# Define data array.
data = np.random.rand(4, 4, 5)
# Define WCS transformations in an astropy WCS object.
wcs = astropy.wcs.WCS(naxis=3)
wcs.wcs.ctype = 'HECR', 'HPLT-TAN', 'HPLN-TAN'
wcs.wcs.cunit = 'Angstrom', 'deg', 'deg'
wcs.wcs.cdelt = 0.2, 0.5, 0.4
wcs.wcs.crpix = 0, 2, 2
wcs.wcs.crval = 10, 0.5, 1

# Now instantiate the NDCube
my_cube = NDCube(data, wcs=wcs)
print(my_cube.array_axis_physical_types)
```
gives
```python
[('custom:pos.helioprojective.lat', 'custom:pos.helioprojective.lon'), ('custom:pos.helioprojective.lat', 'custom:pos.helioprojective.lon'), (None,)]
```
Note that the last entry is `None`. I would expect it to be `pos.distance` looking at https://www.ivoa.net/documents/REC/UCD/UCDlist-20070402.html.

This is on the current RC of ndcube.

Cadair:
It would need adding to the dict in utils/wcs.py :)

PR #367, opened 2020-12-09
Add HEC{R,H} to wcs_to_ivoa mapping

dstansby:
Fixes https://github.com/sunpy/ndcube/issues/366. I think "Body related coordinate (altitude on the body)" is correct for both of these; one is measured from the center of the Sun, and one is measured from the solar surface according to Thompson 2006.

### DIFF
--- a/ndcube/utils/wcs.py
+++ b/ndcube/utils/wcs.py
@@ -44,7 +44,9 @@ wcs_to_ivoa = {
     "PIXEL": "instr.pixel",
     "XPIXEL": "custom:instr.pixel.x",
     "YPIXEL": "custom:instr.pixel.y",
-    "ZPIXEL": "custom:instr.pixel.z"
+    "ZPIXEL": "custom:instr.pixel.z",
+    "HECR": "pos.bodyrc.alt",
+    "HECH": "pos.bodyrc.alt",
 }
 wcs_ivoa_mapping = TwoWayDict()
 for key in wcs_to_ivoa.keys():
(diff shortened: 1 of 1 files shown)


DanRyanIrish:
Hi @dstansby.  Thanks so much for this PR.  I think this change should actually be made in astropy, here: https://github.com/astropy/astropy/blob/master/astropy/wcs/wcsapi/fitswcs.py#L88

Also, if these are different CTYPES giving different, if similar, properties, then they should ideally have different IVOA strings.  If needs be we can create one and prepend it with `custom:`.

ndcube/utils/wcs.py
```diff
@@ -44,7 +44,9 @@ def inv(self):
     "PIXEL": "instr.pixel",
     "XPIXEL": "custom:instr.pixel.x",
     "YPIXEL": "custom:instr.pixel.y",
-    "ZPIXEL": "custom:instr.pixel.z"
+    "ZPIXEL": "custom:instr.pixel.z",
+    "HECR": "pos.bodyrc.alt",
```
DanRyanIrish, line 48:
```suggestion
    "HECR": "custom:pos.heliographic.distance",
```

DanRyanIrish: I think `"HECR": "pos.bodyrc.alt"` makes sense but have suggested a change for `"HECR"`

dstansby:
> Hi @dstansby. Thanks so much for this PR. I think this change should actually be made in astropy, here: https://github.com/astropy/astropy/blob/master/astropy/wcs/wcsapi/fitswcs.py#L88

Is there something special about the ones currently in `ndcube` that means they shouldn't be upstream in astropy?

Cadair:
A lot of them are now, we are just waiting for release of astropy 4.3 to include them. We should actually remove some of these from this list now as they are in 4.2.

Cadair (approved):

PR #382, opened 2020-12-17
Fix unit handling by NDCube.crop_by_values.

### LINKED ISSUE #378: NDCube.crop_by_values does not account for units of inputs
### Description

`NDCube.crop_by_values` in effect ignores the units of the corner inputs.  This seems to be because `WCS.world_to_array_index_values` ignores the units and assumes that the values are in the same units native to the WCS.  Therefore, `NDCube.crop_by_values` should convert the corner inputs to the native WCS units before passing the values on.

An issue has been raised in astropy: https://github.com/astropy/astropy/issues/11173
If this is fixed, not action should be required by ndcube.

### Steps to Reproduce

The following two commands should produce the same result without errors.
```python
# The following works fine because input are in degrees.
cube.crop_by_values([(-900*u.arcsec).to(u.deg), (-900*u.arcsec).to(u.deg)],
                    [(0*u.arcsec).to(u.deg), (-200*u.arcsec).to(u.deg)])

# The following crashes because inputs are equivalent but in arcsec.
cube.crop_by_values([-900*u.arcsec, -900*u.arcsec], [0*u.arcsec, -200*u.arcsec])
---------------------------------------------------------------------------
IndexError                                Traceback (most recent call last)
<ipython-input-41-a40911ee6441> in <module>
----> 1 cube.crop_by_values([-900*u.arcsec, -900*u.arcsec], [0*u.arcsec, -200*u.arcsec])

~/repos/ndcube/ndcube/utils/misc.py in wcs_wrapper(*args, **kwargs)
     52         params.arguments['wcs'] = wcs
     53 
---> 54         return func(*params.args, **params.kwargs)
     55 
     56     return wcs_wrapper

~/repos/ndcube/ndcube/ndcube.py in crop_by_values(self, lower_corner, upper_corner, units, wcs)
    473                 upper_corner[i] = upper_corner[i].to(lower_corner[i].unit)
    474 
--> 475         return self._crop(lower_corner, upper_corner, wcs, True)
    476 
    477     def _crop(self, lower_corner, upper_corner, wcs, crop_by_values):

~/repos/ndcube/ndcube/ndcube.py in _crop(self, lower_corner, upper_corner, wcs, crop_by_values)
    523             # then interval does not overlap with cube range. Raise error.
    524             if upper < 0:
--> 525                 raise IndexError("Input real world interval beyond range of NDCube. "
    526                                  f"Axis: {i}; Axis length: {self.dimensions[i]}; "
    527                                  f"Derived array indices: {(lower, upper)}; ")

IndexError: Input real world interval beyond range of NDCube. Axis: 0; Axis length: 1024.0 pix; Derived array indices: (array(511), array(-9223372036854775808));
```

### System Details

- SunPy Version:
- Astropy Version:
- Python Version:
- OS information:

DanRyanIrish:
### Description

Fixes #378

### DIFF
--- a/changelog/382.bugfix.rst
+++ b/changelog/382.bugfix.rst
@@ -1,0 +1,1 @@
+Ensure corner inputs to `ndcube.NDCube.crop` are converted to units stored in WCS as `~astropy.wcs.WCS.world_to_array_index_values` does not handle units.
--- a/ndcube/ndcube.py
+++ b/ndcube/ndcube.py
@@ -495,6 +495,12 @@ class NDCubeBase(NDCubeSlicingMixin, NDCubeABC):
             except AttributeError:
                 world_to_array_index = wcs.low_level_wcs.world_to_array_index_values
                 array_index_to_world = wcs.low_level_wcs.array_index_to_world_values
+            # Convert coordinates to units used by WCS as WCS.world_to_array_index
+            # does not handle quantities.
+            lower_corner = utils.misc.convert_quantities_to_units(lower_corner,
+                                                                  self.wcs.world_axis_units)
+            upper_corner = utils.misc.convert_quantities_to_units(upper_corner,
+                                                                  self.wcs.world_axis_units)
         else:
             world_to_array_index = wcs.world_to_array_index
             array_index_to_world = wcs.array_index_to_world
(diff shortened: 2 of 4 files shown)


DanRyanIrish:
@Cadair, could you take a glance at this and let me know if there's an obviously better way to do this/have missed a consequence of this?  I think it should be fine and ready to go.

PR #383, opened 2020-12-17
Add Colorbar Example to Docs and Clarify Ordering of plot_axes

### LINKED ISSUE #355: X-axis coordinates missing in some Line and Image Plots produced by NDCube.plot
### Description

Found when producing the 2.0 docs.  Coordinates and ticks missing from X-axis when plots produced.

### Steps to Reproduce

```python
import astropy.units as u
import astropy.wcs
import matplotlib.pyplot as plt
import numpy as np
from astropy.nddata import StdDevUncertainty

from ndcube import NDCube

# Define data array.
data = np.random.rand(4, 4, 5)

# Define WCS transformations in an astropy WCS object.
wcs = astropy.wcs.WCS(naxis=3)
wcs.wcs.ctype = 'WAVE', 'HPLT-TAN', 'HPLN-TAN'
wcs.wcs.cunit = 'Angstrom', 'deg', 'deg'
wcs.wcs.cdelt = 0.2, 0.5, 0.4
wcs.wcs.crpix = 0, 2, 2
wcs.wcs.crval = 10, 0.5, 1

# Define mask.  Initially set all elements unmasked.
mask = np.zeros_like(data, dtype=bool)
mask[-1, -1][:] = True  # Now mask some values.
# Define uncertainty, metadata and unit.
uncertainty = StdDevUncertainty(np.sqrt(np.abs(data)))
meta = {"Description": "This is example NDCube metadata."}
unit = u.ct

# Instantiate NDCube with supporting data.
my_cube = NDCube(data, wcs=wcs, uncertainty=uncertainty, mask=mask, meta=meta, unit=unit)

# Now call NDCube.plot to get faulty plot.
ax = my_cube[1, 1].plot()
plt.show()

# Similar behavior with this too.
ax = my_cube[0, 1:3, 1:4].plot()
plt.show()
```

### System Details

- SunPy Version:
- Astropy Version:
- Python Version:
- OS information:

### LINKED ISSUE #368: What order is the plot_axes list in?
`plot` has a `plot_axes` argument, but it's not clear from the [docstring](https://docs.sunpy.org/projects/ndcube/en/v2.0.0b2/api/ndcube.mixins.NDCubePlotMixin.html) whether the order of this list should be `NAXIS1, NAXIS2...` or correspond to the axes of the input data (which are the reverse order of NAXIS1, NAXIS2 etc.).

### LINKED ISSUE #369: Add short narrative example showing how to add a colorbar
Would be nice if there was a plot on https://docs.sunpy.org/projects/ndcube/en/latest/visualization.html showing how to add a colorbar.

### LINKED ISSUE #370: Add a way to have custom coordinate names?
As an example, there are lots of plots on https://docs.sunpy.org/projects/ndcube/en/latest/visualization.html that have axes labelled as `custom.pos.helioprojective.lon`, which is hard to read and very long.

Of course one can go through and manually edit all of these, but it would be nice when creating a `NDCube` if there was a way of supplying 'custom coordinate names', so NDCube automatically uses these coordinate names when labeling plots.

DanRyanIrish:
### Description

Fixes #355 
Fixes #368
Fixes #369 
Fixes #370

### DIFF
--- a/docs/visualization.rst
+++ b/docs/visualization.rst
@@ -33,6 +33,7 @@ Let's define an `~ndcube.NDCube` as with a shape of ``(4, 4, 5)`` and physical a
   >>> wcs.wcs.cdelt = 0.2, 0.5, 0.4
   >>> wcs.wcs.crpix = 0, 2, 2
   >>> wcs.wcs.crval = 10, 0.5, 1
+  >>> wcs.wcs.cname = 'wavelength', 'HPC lat', 'HPC lon'
 
   >>> # Define mask.  Initially set all elements unmasked.
   >>> mask = np.zeros_like(data, dtype=bool)
@@ -55,6 +56,12 @@ The plot method can be called very simply.
   >>> ax = my_cube.plot()
   >>> plt.show()
 
+Note how no arguments are required.
+The necessary information to generate the plot is derived from the data and metadata in the `~ndcube.NDCube`.
+The axis labels are taken from the WCS axis names defined in ``my_cube.wcs.wcs.cname``.
+Defining these when the WCS is instantiated allows users to customize their axis names.
+However if they choose not to, the axis names are derived from the physical types in ``my_cube.wcs.world_axis_physical_types``.
+The type of visualization returned is derived from the dimensionality of the data.
 For a >2 array axes, as is the case above, an animation object is returned displaying either a line or image with sliders for each additional array axis.
 These sliders are used to sequentially update the line or image as it moves along its corresponding array axis, thus animating the data.
 By default and image animation is returned.
@@ -68,9 +75,6 @@ For for data with two array axes, an image is produced similar to that of `matpl
 
   >>> ax = my_cube[0].plot()
   >>> plt.show()
-
-Note that the bottom row of the image is empty because those array elements are marked ``True`` is ``.mask``.
-
 
 For data with one array axis, a line plot is produced, similar to `matplotlib.pyplot.plot`.
 
@@ -97,10 +101,13 @@ Setting the x and y ranges of the plot can be done simply by indexing the `~ndcu
   >>> plt.cla()
   >>> plt.close()
 
-Note how no arguments are required, the necessary information to generate the plot is derived from the data and metadata in the `~ndcube.NDCube`.
-However optional keywords enable customization of the visualization.
+Note that sometimes axis tickmarks are missing.
+This is a caused by a bug in `~astropy.visualization.wcsaes.WCSAxes` whereby the ticks and labels are omitted if the plot extends beyond the valid range of the WCS projection.
+This can happen when `matplotlib` pads the axes and can be overcome by zooming into the image slightly so that the plot boundaries are again within the valid range of the WCS projection.
+
+Visualizations can be customized via the use of kwargs.
 For `~ndcube.NDCube` instances with more than one array axis, the ``plot_axes`` keyword is used to determine which array axes are displayed on which plot axes.
-It is set to a list with a length equal to the number of array axes.
+It is set to a list with a length equal to the number of array axes in array axis order.
 The array axis to be displayed on the x-axis is marked by ``'x'`` in the corresponding element of the ``plot_axes`` list, while the array axis for the y-axis is marked with a ``'y'``.
 If no ``'y'`` axis is provided, a line animation is produced.
 By default the ``plot_axes`` argument is set so that the last array axis to shown on the x-axis and the penultimate array axis is shown on the y-axis.
@@ -153,6 +160,34 @@ It also enables the coordinates along the plot axes to be updated between frames
   >>> plt.clf()
   >>> plt.cla()
   >>> plt.close()
+
+Adding Colorbars
+----------------
+Working with the output of `ndcube.NDCube.plot` is the context of matplotlib figures and axes can be a great way of creating more complex plots.
+Here we will show two examples of home to add a colorbar.
+
+The first is simple and depends on `matplotlib.pyplot`.
+
+.. code-block:: python
+
+  >>> my_cube.plot()  # doctest: +SKIP
+  >>> plt.colorbar()  # doctest: +SKIP
+
+The second example shows how to more intricately play with `~ndcube.NDCube` visualizations and matplotlib figures and axes.
+This includes adding the output of `ndcube.NDCube.plot` to an existing axes object.
+
+.. plot::
+  :context:
+  :include-source:
+
+  >>> fig = plt.figure()  # Create a figure
+  >>> # Create WCSAxes object and then add the NDCube plot by setting the axes kwarg.
+  >>> ax = fig.add_axes([0.1, 0.1, 0.6, 0.6], projection=my_cube[0].wcs)
+  >>> ax = my_cube[0].plot(axes=ax)
+  >>> # Create the colorbar axes object and scale it by the image.
+  >>> cax = fig.add_axes([0.85, 0.1, 0.05, 0.6])
+  >>> im = ax.get_images()[0]  # Retrieve the plot AxesImage by which to scale colorbar.
+  >>> fig.colorbar(im, cax=cax, label="Intensity")
 
 .. _sequence_plotting:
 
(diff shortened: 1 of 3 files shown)


DanRyanIrish:
@dstansby, hopefully these clarify the issues you raised.  Let me know if you think further changes are needed.

docs/visualization.rst
```diff
@@ -97,10 +101,13 @@ Setting the x and y ranges of the plot can be done simply by indexing the `~ndcu
   >>> plt.cla()
   >>> plt.close()
 
-Note how no arguments are required, the necessary information to generate the plot is derived from the data and metadata in the `~ndcube.NDCube`.
-However optional keywords enable customization of the visualization.
+Note that sometimes axis tickmarks are missing.
+This is a caused by a bug in `~astropy.visualization.wcsaes.WCSAxes` whereby the ticks and labels are omitted if the plot extends beyond the valid range of the WCS projection.
```
dstansby, line 105: This isn't a bug, as I discovered a while ago. It's in general not possible to display the ticks/labels if the spine of an axis is outside the range of the WCS projection.

dstansby (approved):
Looks 👍 , apart from one very minor comment

docs/visualization.rst
Cadair: I think there are a few cases where wcsaxes could do better, but yea there are many where it's not a bug.

docs/visualization.rst
```diff
@@ -154,6 +161,34 @@ It also enables the coordinates along the plot axes to be updated between frames
...
+
+.. plot::
+  :context:
+  :include-source:
+
+  >>> fig = plt.figure()  # Create a figure
+  >>> # Create WCSAxes object and then add the NDCube plot by setting the axes kwarg.
+  >>> ax = fig.add_axes([0.1, 0.1, 0.6, 0.6], projection=my_cube[0].wcs)
+  >>> ax = my_cube[0].plot(axes=ax)
+  >>> # Create the colorbar axes object and scale it by the image.
```
Cadair, line 187: This works but the fact you have to go to these lengths is a bug. We should be setting current axes and current image like map does so it's easier to do.

docs/visualization.rst
dstansby: Do you have any cases in mind? I'd be interested in taking a look and trying to improve things if there are things to be improved.

docs/visualization.rst
DanRyanIrish: To be honest, because I don't use a lot of AIA images in my own work, I'm not sure how this is usually done with `Map`.  It was the first way I found that worked.  If you think there's a simpler way based on the `Map` workflow we can put that here instead as long as it works.

docs/visualization.rst
DanRyanIrish:
```suggestion
This is a caused by a behavior in `~astropy.visualization.wcsaes.WCSAxes` whereby the ticks and labels are omitted if the plot extends beyond the valid range of the WCS projection.
```

docs/visualization.rst
Cadair: Our plot method (for 2D) should include this: https://github.com/sunpy/sunpy/blob/master/sunpy/map/mapbase.py#L2181-L2184

docs/visualization.rst
Cadair: > Do you have any cases in mind? I'd be interested in taking a look and trying to improve things if there are things to be improved.

Well the 1D case (without thinking about it too much) seems particularly silly, how do you reach the edge of a projection in 1D??

Also I feel like for some of the whole sky projections where the coordinates are *so close* to the edge of the projection WCSAxes could be smarter, but I think this was the case you looked at before?

docs/visualization.rst
DanRyanIrish: It does.  For 2-D static images at least: https://github.com/sunpy/ndcube/blob/master/ndcube/mixins/plotting.py#L178

Cadair: oh well then a simple call to `plt.colorbar()` should be all you need?

DanRyanIrish: You mean as shown above?

docs/visualization.rst
```diff
@@ -154,6 +161,34 @@ It also enables the coordinates along the plot axes to be updated between frames
...
+  >>> cax = fig.add_axes([0.85, 0.1, 0.05, 0.6])
+  >>> im = ax.get_images()[0]  # Retrieve the plot AxesImage by which to scale colorbar.
+  >>> cbar = fig.colorbar(im, cax=cax, label="Intensity")
```
Cadair, line 190: I think this should work:

```suggestion
  >>> fig = plt.figure()  # Create a figure
  >>> # Create WCSAxes object and then add the NDCube plot by setting the axes kwarg.
  >>> ax = plt.subplot(projection=my_cube[0])
  >>> ax = my_cube[0].plot(axes=ax)
  >>> cbar = plt.colorbar(label="Intensity")
```

Cadair: even the `(axes=ax)` is optional.

docs/visualization.rst
```diff
@@ -154,6 +161,34 @@ It also enables the coordinates along the plot axes to be updated between frames
...
+----------------
+Working with the output of `ndcube.NDCube.plot` is the context of matplotlib figures and axes can be a great way of creating more complex plots.
+Here we will show two examples of home to add a colorbar.
+
+The first is simple and depends on `matplotlib.pyplot`.
+
+.. code-block:: python
+
+  >>> my_cube.plot()  # doctest: +SKIP
+  >>> plt.colorbar()  # doctest: +SKIP
```
DanRyanIrish, line 174:
```suggestion
.. plot::
  :context:
  :include-source:

  >>> ax = my_cube[0].plot()
  >>> cbar = plt.colorbar()
```

docs/visualization.rst
dstansby: I like leaving in `axes=ax`, as it's explicit and shows the user how to choose which axes to plot their cube on.

docs/visualization.rst
```diff
@@ -154,6 +161,36 @@ It also enables the coordinates along the plot axes to be updated between frames
...
+Here we will show two examples of home to add a colorbar.
+
+The first is simple and depends on `matplotlib.pyplot`.
+
+.. plot::
+  :context:
+  :include-source:
+
+  >>> ax = my_cube[0].plot()
+  >>> cbar = plt.colorbar()
```
DanRyanIrish, line 176:
```suggestion
  >>> cbar = plt.colorbar()
  >>> plt.show()
```

docs/visualization.rst
dstansby: Does `fig.subplot` work instead? Again, then it's more explicit.

DanRyanIrish: I've tried to make the above example as short and easy as possible and the second to me much more explicit.

docs/visualization.rst
```diff
@@ -154,6 +161,43 @@ It also enables the coordinates along the plot axes to be updated between frames
...
```
DanRyanIrish, line 176:
```suggestion
  >>> cbar = plt.colorbar(orientation="horizontal")
```

docs/visualization.rst
```diff
@@ -154,6 +161,43 @@ It also enables the coordinates along the plot axes to be updated between frames
...
+  :include-source:
+
+  >>> fig = plt.figure()  # Create a figure
+  >>> # Create WCSAxes object and then add the NDCube plot by setting the axes kwarg.
+  >>> ax = fig.add_axes([0.1, 0.1, 0.6, 0.6], projection=my_cube[0].wcs)
+  >>> ax = my_cube[0].plot(axes=ax)
+  >>> # Create the colorbar axes object and scale it by the image.
+  >>> cax = fig.add_axes([0.85, 0.1, 0.05, 0.6])
+  >>> im = ax.get_images()[0]  # Retrieve the plot AxesImage by which to scale colorbar.
+  >>> cbar = fig.colorbar(im, cax=cax, label="Intensity")
```
DanRyanIrish, line 199:
```suggestion
  >>> cbar = fig.colorbar(im, cax=cax, label="Intensity")
  >>> plt.show()
```

docs/visualization.rst
Cadair: I always feel like `fig.add_axes` is too much for 99% of the time. Why on earth do you want to be manually defining the size of your axes like this? If you really care use gridspec or something.

We could show an example using only the mpl OO API though, if `fig.subplot` does indeed work.

DanRyanIrish: I'm going to leave that for someone who knows about this stuff to make in a separate PR.

PR #393, opened 2021-02-04
Make sunpy dependency optional

### LINKED ISSUE #149: Make the sunpy dependency optional

rosteen:
### Description
In the interest of helping push the 2.0 release along I thought I'd take a crack at this - it's probably not quite the prettiest way to do it, but it seems to work. Likely still needs a couple test-related changes. Currently I have it only throwing a warning about the sunpy dependency if you actually try to use the `plot` method.

Fixes #149

### DIFF
--- a/changelog/393.trivial.rst
+++ b/changelog/393.trivial.rst
@@ -1,0 +1,2 @@
+Make sunpy an optional dependence. Without it, the _animate_cube plotting
+functionality will be disabled.
--- a/ndcube/mixins/plotting.py
+++ b/ndcube/mixins/plotting.py
@@ -6,12 +6,6 @@ import numpy as np
 from astropy.utils.exceptions import AstropyUserWarning
 from astropy.visualization.wcsaxes import WCSAxes
 
-try:
-    from sunpy.visualization.animator import ArrayAnimatorWCS
-    sunpy_available = True
-except ImportError:
-    sunpy_available = False
-
 from . import plotting_utils as utils
 
 __all__ = ['NDCubePlotMixin']
@@ -191,9 +185,11 @@ class NDCubePlotMixin:
     def _animate_cube(self, wcs, plot_axes=None, axes_coordinates=None,
                       axes_units=None, data_unit=None, **kwargs):
 
-        if not sunpy_available:
-            raise ModuleNotFoundError("Sunpy is required for animated "
-                                      "cube plots.")
+        try:
+            from sunpy.visualization.animator import ArrayAnimatorWCS
+        except ImportError:
+            raise ImportError("Sunpy is required for animated "
+                              "cube plots.")
 
         # If data_unit set, convert data to that unit
         if data_unit is None:
--- a/ndcube/mixins/plotting.py
+++ b/ndcube/mixins/plotting.py
@@ -186,7 +186,7 @@ class NDCubePlotMixin:
                       axes_units=None, data_unit=None, **kwargs):
 
         try:
-            from sunpy.visualization.animator import ArrayAnimatorWCS
+            from sunpy.visualization.animator import ArrayAnimatorWCS  # isort:skip
         except ImportError:
             raise ImportError("Sunpy is required for animated "
                               "cube plots.")
--- a/ndcube/mixins/plotting.py
+++ b/ndcube/mixins/plotting.py
@@ -5,6 +5,7 @@ import matplotlib.pyplot as plt
 import numpy as np
 from astropy.utils.exceptions import AstropyUserWarning
 from astropy.visualization.wcsaxes import WCSAxes
+
 try:
     from sunpy.visualization.animator import ArrayAnimatorWCS
     sunpy_available = True
@@ -86,13 +87,10 @@ class NDCubePlotMixin:
             elif naxis == 2 and 'y' in plot_axes:
                 ax = self._plot_2D_cube(plot_wcs, axes, plot_axes, axes_coordinates,
                                         axes_units, data_unit, **kwargs)
-            elif sunpy_available:
+            else:
                 ax = self._animate_cube(plot_wcs, plot_axes=plot_axes,
                                         axes_coordinates=axes_coordinates,
                                         axes_units=axes_units, **kwargs)
-            else:
-                raise ModuleNotFoundError("Sunpy is required for animated "
-                                          "cube plots.")
 
         return ax
 
@@ -192,6 +190,10 @@ class NDCubePlotMixin:
 
     def _animate_cube(self, wcs, plot_axes=None, axes_coordinates=None,
                       axes_units=None, data_unit=None, **kwargs):
+
+        if not sunpy_available:
+            raise ModuleNotFoundError("Sunpy is required for animated "
+                                      "cube plots.")
 
         # If data_unit set, convert data to that unit
         if data_unit is None:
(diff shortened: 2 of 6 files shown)


DanRyanIrish:
Hi @rosteen,  Thanks so much for this PR.  It's relevant to note that ndcube plotting only requires sunpy to make animations.  And these are only produced when the cube is >2D (or  == 2D and there is only 1 plot axis).

@Cadair and I have had several conversations about how best to address this issue. Our idea was to move the sunpy animator classes either into ndcube or their own package.  Your implementation has the potential to decouple progress on ndcube 2.0 from that migration which would be great!  However, I would like to get feedback from @Cadair as there may be considerations in favour of blocking ndcube until that migration is complete.

The one thing I would mention about your implementation is that it quite punitive.  It blocks all plotting (not just animations) if sunpy is not installed.  However, with a little more argument checking, this could probably be relaxed.

Thanks again for this PR.  It's great to have different perspectives on the best way forward and great to have more people interested in contributing to ndcube!

rosteen:
Ah yes, I see that I was a little heavy handed with the scalpel cutting the entire `plot` functionality. It would be easy enough to change to snipping out only the animation that I'll just go ahead and do that, and then you all can decide whether to use this or wait and handle it otherwise.

If any of the other 2.0 MVP issues would be particularly useful for me to take a look at let me know - I decided that it's probably a better use of my time to help ndcube 2.0 along than spend time trying to make my other work compatible with the current 1.x release.

DanRyanIrish:
This looks much more subtle @rosteen.  Let's get feedback from @Cadair before we take this any further to minimize the chance of wasting your effort.

As for other issues, it would be great if you wanted to work on some others.  Let me tag you on a few that I think might be good to tackle next.

> I decided that it's probably a better use of my time to help ndcube 2.0 along than spend time trying to make my other work compatible with the current 1.x release.

I think this is a great strategy and certainly helps us!

ndcube/ndcube.py
```diff
@@ -6,14 +6,17 @@
 import astropy.units as u
 import gwcs
 import numpy as np
-import sunpy.coordinates  # pylint: disable=unused-import  # NOQA
+try:
+    import sunpy.coordinates  # pylint: disable=unused-import  # NOQA
```
Cadair, line 10:
```suggestion
    # Import sunpy coordinates if available to register the frames and WCS functions with astropy
    import sunpy.coordinates  # pylint: disable=unused-import  # NOQA
```

ndcube/mixins/plotting.py
```diff
@@ -82,10 +86,13 @@ def plot(self, axes=None, plot_axes=None, axes_coordinates=None,
             elif naxis == 2 and 'y' in plot_axes:
                 ax = self._plot_2D_cube(plot_wcs, axes, plot_axes, axes_coordinates,
                                         axes_units, data_unit, **kwargs)
-            else:
+            elif sunpy_available:
```
Cadair, line 89: I think it would be cleaner to shuffle this check into the `_animate_cube` method as then someone could overload it to animate a cube not with sunpy.

In fact going one step further than that, you could just put the import into that method.

Cadair (commented):
I don't see any reason not to do this now and worry about a cleaner solution later. I have left a couple of comments in line.

ndcube/mixins/plotting.py
DanRyanIrish: Although having imports in the middle of the code is typically frowned upon, in this case I think that's a great idea!

DanRyanIrish:
> I don't see any reason not to do this now and worry about a cleaner solution later. I have left a couple of comments in line.

OK, great!  Let's push this PR forward then.  I'll add it to the 2.0 project.

@rosteen, we also need a changelog.

DanRyanIrish (approved):

ndcube/mixins/plotting.py
rosteen: Agreed, it's the kind of thing that I don't generally like to do, but it's definitely cleaner here. I kept the bit of added context for the ImportError

ndcube/ndcube.py
```diff
@@ -6,14 +6,18 @@
...
+except ImportError:
+    pass
 from astropy.wcs.wcsapi import HighLevelWCSWrapper
 from astropy.wcs.wcsapi.wrappers import SlicedLowLevelWCS
 
 from ndcube import utils
 from ndcube.extra_coords import ExtraCoords
 from ndcube.global_coords import GlobalCoords
-from ndcube.mixins import NDCubePlotMixin, NDCubeSlicingMixin
+from ndcube.mixins import NDCubeSlicingMixin, NDCubePlotMixin
```
DanRyanIrish, line 20: Actually, is there a reason these imports have been reversed?  It's at least a reason the codestyle is complaining, because they are out of alphabetical order.  Although in the case of this PR, the import within the code might also be a cause it's complaining.

DanRyanIrish:
```suggestion
from ndcube.mixins import NDCubePlotMixin, NDCubeSlicingMixin
```

rosteen: Hm, apparently they do - the tox codestyle is passing on my local branch where I have them reversed, I'll commit that and push it. I didn't think that difference was important, silly me.

ndcube/mixins/plotting.py
```diff
@@ -186,6 +185,12 @@ def _plot_2D_cube(self, wcs, axes=None, plot_axes=None, axes_coordinates=None,
     def _animate_cube(self, wcs, plot_axes=None, axes_coordinates=None,
                       axes_units=None, data_unit=None, **kwargs):
 
+        try:
+            from sunpy.visualization.animator import ArrayAnimatorWCS
```
DanRyanIrish, line 189:
```suggestion
            from sunpy.visualization.animator import ArrayAnimatorWCS  # isort:skip
```
This should make the codestyle test ignore this import and hopefully pass.

rosteen: I thought I saw all the tests pass without this, but it doesn't hurt.

changelog/393.trivial.rst
```diff
@@ -0,0 +1,2 @@
+Make sunpy an optional dependence. Without it, the _animate_cube plotting
+functionality will be disabled.
```
Cadair, line 2: Just an alternative wording:
```suggestion
Make the dependency on sunpy optional. If sunpy is not installed plot will not work on cubes of more than two dimensions.
```

ndcube/mixins/plotting.py
```diff
@@ -186,6 +185,12 @@ def _plot_2D_cube(self, wcs, axes=None, plot_axes=None, axes_coordinates=None,
     def _animate_cube(self, wcs, plot_axes=None, axes_coordinates=None,
                       axes_units=None, data_unit=None, **kwargs):
 
+        try:
+            from sunpy.visualization.animator import ArrayAnimatorWCS  # isort:skip
+        except ImportError:
+            raise ImportError("Sunpy is required for animated "
+                              "cube plots.")
```
Cadair, line 192:
```suggestion
            raise ImportError("Sunpy is required for animated cube plots. "
                              "Either install sunpy or slice your cube before plotting.")
```

DanRyanIrish:
Another great contribution, @rosteen!  We're getting closer to 2.0!

Cadair (approved):

Cadair:
heh lolol

Cadair:
Thanks @rosteen

DanRyanIrish:
Woops!  Sorry @Cadair!

Issue #348, opened 2020-11-22
Rename ExtraCoords.add_coordinate to .add.


DanRyanIrish:
### Description

Rename `ExtraCoords.add_coordinate` and `ExtraCoords.remove_coordinate` should be changed to `.add` and `.remove` to be consistent with `GlobalCoords`.  This represents concensus reached on the SEP thread (sunpy/sunpy-SEP#51).  (See conversation thread below.)

add_coordinate: https://github.com/sunpy/ndcube/blob/master/ndcube/extra_coords/extra_coords.py#L179

DanRyanIrish:
ping @Cadair

Cadair:
We would want `remove_coordinate` as well then?

Cadair:
(This should probably be a comment on the SEP as well)

DanRyanIrish:
Consensus on SEP thread (https://github.com/sunpy/sunpy-SEP/pull/51) is that `ExtraCoords.add_coordinate` and `ExtraCoords.remove_coordinate` should be changed to `.add` and `.remove` to be consistent with `GlobalCoords`.

DanRyanIrish:
@rosteen, this is another conceptually simple issue that needs fixing for 2.0.

DanRyanIrish:
> We would want `remove_coordinate` as well then?

@Cadair, what do you say to making the creation of `ExtraCoords.remove` a 2.1 goal?

DanRyanIrish:
Wait.  `ExtraCoords.remove` might be needed when an axis is sliced away?

DanRyanIrish:
The addition of a `ExtraCoords.remove` method has been moved to #395

PR #394, opened 2021-02-05
Rename ExtraCoords method 'add_coordinates' to 'add'

rosteen:
### Description

Addresses the `add` half of #348. It sounds like there needs to be a `remove` method added as well, and while I assume that needs to do exactly what it sounds like I figured I'd hold off for now in case there are more details or it can wait for 2.1.

### DIFF
--- a/changelog/394.breaking.rst
+++ b/changelog/394.breaking.rst
@@ -1,0 +1,1 @@
+Rename `.ExtraCoords.add` method from previous name "add_coordinate".
--- a/docs/coordinates.rst
+++ b/docs/coordinates.rst
@@ -204,7 +204,7 @@ If one does not exist for your coordinate, prepend the type with ``custom:``.
 
 .. code-block:: python
 
-  >>> my_cube.extra_coords.add_coordinate('time', (2,), timestamps)  # TODO: Change the mapping to 0 Issue #342 resolved.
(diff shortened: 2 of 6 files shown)


DanRyanIrish (changes requested):
Hi @rosteen.  This looks great.  Could you add a changelog file to the changelog directory?  In case you are used to a different way of doing the changelog, ndcube follows the same way as sunpy, outlined here: https://github.com/sunpy/sunpy/blob/master/changelog/README.rst

Also you need to make the codestyle fix I suggested to get the tests passing.  Note the line below my suggestion also needs its indentation fixed.

DanRyanIrish:
Hi @rosteen.  Looks like you failed to include the indentation changes in your latest commit.

rosteen:
Whoops, the good old "forget to git add a file before committing", thanks for the heads up.

DanRyanIrish:
> Whoops, the good old "forget to git add a file before committing", thanks for the heads up.

That's exactly what I suspected happened :)

DanRyanIrish (approved):

DanRyanIrish:
Merged!  Thanks @rosteen for this PR.  Another small step towards 2.0!  If you would like to attack any of the other issues I mentioned to you and have any questions, let me know.

Issue #396, opened 2021-02-08
Better handling of custom plot aspect


dstansby:
Plotting an EIS data cube produces the following plot on the left, which is not very helpful. In an attempt to improve matters, I used `eiscube.plot(aspect=0.1)`, which is shown on the right, which is even worse! 
<img width="263" alt="Screenshot 2021-02-08 at 20 48 42" src="https://user-images.githubusercontent.com/6197628/107279609-2d4e5100-6a4f-11eb-9438-74f2c11eb0eb.png"> <img width="260" alt="Screenshot 2021-02-08 at 20 49 00" src="https://user-images.githubusercontent.com/6197628/107279647-3808e600-6a4f-11eb-87f6-3a178f8d9f27.png">
It looks like there needs to be some cleverer handling of custom plot aspects.

This is on NDCube 2.0-pre, and matplotlib 3.3.4

dstansby:
For reference, `aspect='auto'` works well, but it would still be good to fix handling a custom aspect.

wtbarnes:
In the case of EIS, I've found that setting `aspect=meta['cdelt2']/meta['cdelt1']` works pretty well (you may have been the one that pointed this out to me). Maybe this ratio would be a sensible default?

dstansby:
I'm not too worried about choosing a suitable aspect here, more that the plot gets tiny when a suitable aspect is chosen.

DanRyanIrish:
Hi @dstansby.  Thanks for raising this.  I certainly see your point and appreciate this is a bit of a pain for EIS data.  However, I have a philosophical concern for ndcube which may or may not be an argument against your issue.  ndcube is aimed at any N-D data using WCS coordinates, not just slit spectrograms or even just solar.  Because of this huge scope, ndcube needs to be minimalist and should be only introduce defaults that are common to all (or nearly all) use cases.  This both reduces code maintenance and on average leads to more intuitive behaviour.  This is especially true when the underlying infrastructure and its defaults are commonly used outside of ndcube, as is the case for matplotlib.

[...]

PR #398, opened 2021-02-08
Make 1D plots step plots

dstansby:
Because FITS data are binned, and not samples at a single point, I think it makes sense for 1D plots to be step plots instead of line plots. This is analogous to the 2D case, where we show 'pixelated' images, as opposed to linearly interpolating between the centers of each pixel. Before and after:
before
after

### DIFF
--- a/ndcube/mixins/plotting.py
+++ b/ndcube/mixins/plotting.py
@@ -138,11 +138,22 @@ class NDCubePlotMixin:
             if yerror is not None:
                 yerror = np.ma.masked_array(yerror, self.mask)
 
+        # The centers of the bins are integer values, so the edges are at 0.5
+        x = np.arange(ydata.size + 1) - 0.5
+        # Need to append the final value to get `axes.step` to draw the final
+        # bin. When NDCube depends on matplotlib>=3.4, `axes.stairs` can be
+        # used instead to avoid this extra step.
+        ydata = np.append(ydata, ydata[-1])
+
+        lines = axes.step(x, ydata, where='post', **kwargs)
         if yerror is not None:
-            # We plot against pixel coordinates
-            axes.errorbar(np.arange(len(ydata)), ydata, yerr=yerror, **kwargs)
-        else:
-            axes.plot(ydata, **kwargs)
+            yerror = np.append(yerror, yerror[-1])
+            axes.step(x, ydata + yerror, where='post',
+                      color=lines[0].get_color(),
+                      lw=lines[0].get_linewidth() / 2, **kwargs)
+            axes.step(x, ydata - yerror, where='post',
+                      color=lines[0].get_color(),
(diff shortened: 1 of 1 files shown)


DanRyanIrish:
> Because FITS data are binned, and not samples at a single point, I think it makes sense for 1D plots to be step plots instead of line plots.

Is this really true in general?  I don't think FITS makes any stipulation that this is the case.  It is often true in solar but this is always a point of debate at PyAstro spectroscopy breakouts.  There are a lot on the astro community that think in terms of samples along points of a distribution rather than photons in a bin.

Also, the step example you've shown is not as nice or easy to read as the current version.  Having said that, I agree there are many use cases where step is more appropriate.  Notwithstanding what I said in issue #396, perhaps there's a way we can enable users to choose whether they want step or plot...

dstansby:
Ah fair enough, I assumed that FITS data were always binned. If not the default, it would definitely be nice to have the option to do this though. Do you think a separate method, or a parameter (`plot_type='step'` maybe?) would be best?

DanRyanIrish:
> Do you think a separate method, or a parameter (`plot_type='step'` maybe?) would be best?

I was just pondering that very thing.  Keeping everything in one `plot` method is nice, but do we run the risk of overloading that method?  People are used to `plt.step` so `NDCube.step` might make more sense, although it increases the number of methods and in the `NDCube` case, it may not be obvious that `.step` is a plotting method...  And I think that `NDCube.step` would just be redundant with `NDCube.plot` for 2-D animations and images.  So perhaps a `plot_type='step'` would be best.

If we do this we *should* also do it for the 1-D animator to be consistent.  That would probably merit a new animator class but may be able to inherit most of what it needs from existing animator classes.

ndcube/mixins/plotting.py
```diff
@@ -138,11 +138,22 @@ def _plot_1D_cube(self, wcs, axes=None, axes_coordinates=None, axes_units=None,
...
+        x = np.arange(ydata.size + 1) - 0.5
+        # Need to append the final value to get `axes.step` to draw the final
+        # bin. When NDCube depends on matplotlib>=3.4, `axes.stairs` can be
+        # used instead to avoid this extra step.
+        ydata = np.append(ydata, ydata[-1])
+
+        lines = axes.step(x, ydata, where='post', **kwargs)
         if yerror is not None:
-            # We plot against pixel coordinates
-            axes.errorbar(np.arange(len(ydata)), ydata, yerr=yerror, **kwargs)
```
DanRyanIrish, line 143: If we implement this, I think this still the right line of code for the step case, except that your should add `fmt='none'` so that only the errorbars are plotted.

dstansby:
Fab, I'll have a bit more of a think about API, and hold off until 2.0 is shipped 👍

dstansby:
Going to close this, and will re-think once https://github.com/sunpy/ndcube/pull/401 is done.

PR #399, opened 2021-02-09
Fix dimensions doc

dstansby:
This seems outdated and wrong, and it should inherit the docstring from a parent class, so I've just deleted it.


DanRyanIrish:
Hi @dstansby.  You're right.  Thanks for this PR.  Can you add a `399.trivial.rst` changelong and we'll get this merged.

DanRyanIrish (changes requested):
Perfect!  Just needs a `399.trivial.rst` changelog file.

Issue #334, opened 2020-11-15
Add coordinates from NDCube sliced out axes to GlobalCoords


DanRyanIrish:
### Description

#323 creates a new `GlobalCoords` class.  Now, coordinates corresponding axes sliced out of an `NDCube` (i.e. when slicing reduces the dimensionality of an `NDCube`) must be added to the instance of the `GlobalCoords` attached to the `NDCube`.  Therefore the following is needed:
* ~An API for initialising an `NDCube` with `GlobalCoords` must be created.~
* When an axis is dropped as part of slicing an `NDCube`, the value of each coordinate at the point along the axis where the cube was sliced must be transferred to `GlobalCoords`.  If that coordinate still depends on another axis that has not been sliced away, then it should not be added to `GlobalCoords`.  This should be implemented in `NDCube`'s `__getitem__`.
* The above must be done for coords from both the `wcs` and the `extra_coords`.
* Testing the right coords have been inserted when slicing occurs must be added.

Cadair:
To add a little more detail on how this should be implemented. In astropy 4.2 `SlicedLowLevelWCS` has a [`dropped_world_dimensions`](https://github.com/astropy/astropy/blob/master/astropy/wcs/wcsapi/wrappers/sliced_wcs.py#L157) method which contains all the information needed to generate the elements of the `GlobalCoords` object. This property should be accessed inside `GlobalCoords._all_coords` and the low level objects converted to hight level objects with code similar to `HighLevelWCSWrapper.pixel_to_world`.

I also propose we implement the same system in `ExtraCoords` where when `ExtraCoords.__getitem__` is called a `ExtraCoords.dropped_world_dimensions` property is populated, returning the same information as the `SlicedLowLevel` equivalent.

DanRyanIrish:
:+1:

DanRyanIrish:
@rosteen, you might be interested in tackling this issue.  To start with I would just tackle the case of adding sliced out global coordinates from the WCS object and leave the extra_coords case for a later PR.

Cadair:
This is on the top of my list, but I am happy to tackle the ExtraCoords part if someone else wants to take a swing at the wcs part.

DanRyanIrish:
Splitting this into two PRs might not be the worst idea anyway.  So why don't you, @Cadair, start with the `ExtraCoords` part to give @rosteen a chance to do the WCS part.  The time saved might also allow you to look at one of the other outstanding `ExtraCoords` issues as well?

[...]

Issue #349, opened 2020-11-22
Extra Coords not removed by NDCube slicing when their axis is sliced away


DanRyanIrish:
### Description

NDCube slicing does not remove extra coords if their axis is sliced away.  Instead it is moved to the axis in the sliced cube with the equivalent axis number.

### Steps to Reproduce

```python
>>> import numpy as np
>>> import astropy.wcs
>>> data = np.ones((3, 4, 5))
>>> wcs_input_dict = {
...     'CTYPE1': 'WAVE    ', 'CUNIT1': 'Angstrom', 'CDELT1': 0.2, 'CRPIX1': 0, 'CRVAL1': 10, 'NAXIS1': 5,
...     'CTYPE2': 'HPLT-TAN', 'CUNIT2': 'deg', 'CDELT2': 0.5, 'CRPIX2': 2, 'CRVAL2': 0.5, 'NAXIS2': 4,
...     'CTYPE3': 'HPLN-TAN', 'CUNIT3': 'deg', 'CDELT3': 0.4, 'CRPIX3': 2, 'CRVAL3': 1, 'NAXIS3': 3}
>>> input_wcs = astropy.wcs.WCS(wcs_input_dict)

>>> from astropy.time import Time, TimeDelta
>>> base_time = Time('2000-01-01', format='fits', scale='utc')
>>> timestamps = Time([base_time + TimeDelta(60 * i, format='sec') for i in range(data.shape[0])])

>>> from ndcube import ExtraCoords
>>> my_extra_coords = ExtraCoords()
>>> my_extra_coords.add_coordinate('time', (0,), timestamps)

>>> from ndcube import NDCube
>>> my_cube = NDCube(data, input_wcs, extra_coords=my_extra_coords)

>>> my_cube.array_axis_physical_types
[('custom:pos.helioprojective.lat', 'custom:pos.helioprojective.lon'),
 ('custom:pos.helioprojective.lat', 'custom:pos.helioprojective.lon'),
 ('em.wl', 'time')]

>>> my_cube[:, :, 0].array_axis_physical_types
[('custom:pos.helioprojective.lat', 'custom:pos.helioprojective.lon'),
 ('custom:pos.helioprojective.lat', 'custom:pos.helioprojective.lon', 'time')]
```

### System Details

- SunPy Version:
- Astropy Version:
- Python Version:
- OS information:

Cadair:
Just re-tested this with master and it's fixed.

Cadair:
[...]

PR #403, opened 2021-02-17
Fixing bugs in LookupTableCoord slicing

Cadair:
In the process of attempting to address the extra coords part of #334 I noticed there were some issue with the way we handled slicing in the `LookupTableCoord` stack, which would be very problematic when it came to dropping dimensions.

I haven't added any dimension dropping in this PR, axes in the table just become length one, but it hopefully lays the ground work for a follow on PR.

The main change introduced in this PR is to move the generation of the gWCS frame into what was the "delayed model" layer. So the three layers of the stack are now `ExtraCoords > LookupTableCoord > BaseTableCoordinate` where there is a subclass of `BaseTableCoordinate` for each type of table we support. This means all the frame generation and slicing logic are all in a type-specific class and the generation of the frame and model is delayed until the `.wcs` property of the `LookupTableCoord` class is called. This means that all the slicing operations now happen on the actual object the user passes to us, we only convert to a model at the last moment.


DanRyanIrish (commented):
Looks good!  A few minor questions and comments.

DanRyanIrish:
Also, there's no changelog.  I know this isn't changing currently available code, but does alter code described in previous changelogs.  Do we need a changelog here then?

ndcube/extra_coords/lookup_table_coord.py
```diff
@@ -14,6 +14,31 @@
...
+    axes_order = tuple(range(naxes))
+
+    name = None
+    axes_type = "CUSTOM"
+
+    if isinstance(unit, (u.Unit, u.IrreducibleUnit, u.CompositeUnit)):
+        unit = tuple([unit] * naxes)
+
+    if all([u.m.is_equivalent(un) for un in unit]):
+        axes_type = "SPATIAL"
```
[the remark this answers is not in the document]
Cadair: Temporal and spectral have their own frames, this is a generic frame distance, pixel or unknown units.

ndcube/extra_coords/lookup_table_coord.py
```diff
@@ -67,28 +92,26 @@ class LookupTableCoord:
...
             if not all(isinstance(lt, type(lt0)) for lt in lookup_tables):
                 raise TypeError("All lookup tables must be the same type")
 
             type_map = {
-                u.Quantity: self._from_quantity,
-                Time: self._from_time,
-                SkyCoord: self._from_skycoord
+                u.Quantity: QuantityTableCoordinate,
+                SkyCoord: SkyCoordTableCoordinate,
+                Time: TimeTableCoordinate,
```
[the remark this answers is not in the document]
Cadair: Yes, but I have never implemented that, it should be done but not in this PR.

ndcube/extra_coords/lookup_table_coord.py
```diff
@@ -214,113 +270,201 @@ def _generate_compound_model(cls, *lookup_tables, mesh=True):
...
         components = tuple(getattr(sc.data, comp) for comp in sc.data.components)
         ref_frame = sc.frame.replicate_without_data()
         units = list(c.unit for c in components)
 
-        if names and len(names) != 2:
+        names = self.names
+        if self.names and len(self.names) != 2:
             names = None
 
         # TODO: Currently this limits you to 2D due to gwcs#120
```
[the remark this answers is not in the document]
Cadair: yep

ndcube/tests/test_lookup_table_coord.py
```diff
@@ -94,85 +130,54 @@ def test_2d_nout_1_no_mesh():
...
 
-    sub_ltc = ltc[1:3, 1:2]
-    assert sub_ltc.delayed_models[0].lookup_table[0].shape == (2, 1)
-    assert sub_ltc.delayed_models[0].lookup_table[1].shape == (2, 1)
+    # assert isinstance(ltc.wcs, gwcs.WCS)
+    #
+    # sub_ltc = ltc[0:4, 0:5, 0:6]
+    # assert sub_ltc.delayed_models[0].lookup_table[0].shape == (4, )
+    # assert sub_ltc.delayed_models[0].lookup_table[1].shape == (5, )
+    # assert sub_ltc.delayed_models[0].lookup_table[2].shape == (6, )
```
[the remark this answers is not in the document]
Cadair: It's been open a long time on gWCS, so we are probably going to have to do it ourselves.

ndcube/tests/test_lookup_table_coord.py
```diff
@@ -209,7 +201,8 @@ def test_join_3d():
     assert isinstance(world[0], SkyCoord)
     assert u.allclose(world[1], 0 * u.nm)
 
-    assert u.allclose(ltc.wcs.world_to_pixel(*world), (0, 0, 0))
+    # TODO: Investigate this, something about inverse model
+    # assert u.allclose(ltc.wcs.world_to_pixel(*world), (0, 0, 0))
```
Cadair, line 205: This is a real issue.

DanRyanIrish (approved):

Cadair:
> Also, there's no changelog. I know this isn't changing currently available code, but does alter code described in previous changelogs. Do we need a changelog here then?

It doesn't alter anything I would consider "user facing" it all behaves the same as before but it just works differently, at least that was my logic for skipping the changelog.

ndcube/tests/test_lookup_table_coord.py
```diff
@@ -5,41 +5,89 @@
...
+    return TimeTableCoordinate(data)
+
+
+@pytest.fixture
+def lut_1d_wave():
+    # TODO: Make this into a SpectralCoord object
+    return QuantityTableCoordinate(range(10) * u.nm)
+
+
+# def test_exceptions(lut_1d_distance):
```
Cadair, line 73: fix me

ndcube/extra_coords/extra_coords.py
```diff
@@ -298,10 +312,13 @@ def _getitem_lookup_tables(self, item):
...
             lut_axes = (lut_axis,) if not isinstance(lut_axis, tuple) else lut_axis
             lut_slice = tuple(item[i] for i in lut_axes) if isinstance(item, tuple) else item
+            if isinstance(lut_slice, tuple) and len(lut_slice) == 1:
+                lut_slice = lut_slice[0]
 
             sliced_lut = lut[lut_slice]
-            if sliced_lut:
+            if not isinstance(lut_slice, Integral):
                 new_lookup_tables.add((lut_axis, sliced_lut))
+            # TODO: Handle dropped table here
```
DanRyanIrish, line 321: Is this to be fixed in this PR or is this where a new PR should address #349?

ndcube/extra_coords/lookup_table_coord.py
```diff
@@ -9,318 +9,431 @@
...
-                SkyCoord: self._from_skycoord
-            }
-            delayed_model, frame = type_map[type(lt0)](lookup_tables,
-                                                       mesh=mesh,
-                                                       names=names,
-                                                       physical_types=physical_types)
-            self.delayed_models = [delayed_model]
-            self.frames = [frame]
+    ndim = lookup_table.ndim
+    TabularND = tabular_model(ndim, name=f"Tabular{ndim}D")
```
DanRyanIrish, line 49: Is this returning an class or an instance of a class?  If the latter, perhaps it shouldn't be camelcase.

DanRyanIrish:
This updated implementation looks good.  Barring further work on the tests, am happy if you feel the core code is done.

ndcube/extra_coords/lookup_table_coord.py
Cadair: It returns a class.

ndcube/extra_coords/extra_coords.py
Cadair: Probably a follow on one.

rosteen:
Looks good to me.

PR #404, opened 2021-02-18
Enable initializing an NDCube from an existing NDCube

### LINKED ISSUE #385: NDcube should be able to be instantiated from another NDCube
If this doesn't already work it should. This would allow you to instantiate a sunraster object from a DKIST user tools object for instance, which is probably something that people are going to want to do.

rosteen:
### Description

Fixes #385 . As far as I could tell, the only things not handled by the underlying `NDData` machinery were pulling in `global_coords` and `extra_coords` from the input `NDCube`. If there are other `NDCube` attributes that I missed, let me know. I think the way I handled the `copy` cases is consistent with the way `NDData` handles it for the rest of the attributes.


Cadair (approved):
Looks uncontroversial to me. Thanks.

ndcube/ndcube.py
```diff
@@ -186,13 +187,27 @@ def __init__(self, data, wcs=None, uncertainty=None, mask=None, meta=None,
...
+            global_coords = data._global_coords
+        else:
+            global_coords = GlobalCoords(self)
+
+        if copy:
+            extra_coords = deepcopy(extra_coords)
+            global_coords = deepcopy(global_coords)
+
         self._extra_coords = extra_coords
-        self._global_coords = GlobalCoords(self)
```
DanRyanIrish, line 195: If we are allowing the user to provide a separate `ExtraCoords` instance when NOT instantiating from another `NDCube`, should we also allow the user to supply a separate `GlobalCoords` instance via a `global_coords` kwarg?  Or am I forgetting that `GlobalCoords` needs a soft copy of the `NDCube` and not allowing this is a safety feature?  @Cadair?

DanRyanIrish: @Cadair: Could you clarify this before I merge?

Cadair: I would say that adding a `global_coords` kwarg is out of scope for this PR, we should probably open an issue and come to a decision there. I am not sure what shape that API should take.

PR #410, opened 2021-03-04
Implement SpectralCoordTableCoordinate

### LINKED ISSUE #389: Add SpectralCoord support to ExtraCoords
At the moment there is only quantity support and the units alone are not enough to distinguish between spatial (length) dimensions and spectral dimensions.

rosteen:
### Description

Not done, but intended to close #389 - I'm opening a PR at this point to get feedback and input. I'm not that familiar with gWCS and reference frames so I've honestly mostly bumbled my way through this so far based on the other lookup table classes. Things I'm unclear on:
- I think that the information that `SpectralCoord` stores in its `observer` attribute corresponds to what `SpectralFrame` wants for its `reference_frame` parameter, but I haven't found much documentation for SpectralFrame to confirm this.
- Right now `SpectralCoordTableCoordinate.model` errors out with `ValueError: Lookup table must have at least one dimension`. I haven't spent enough time to fully comprehend what the model machinery is doing here to fix this. I suspect it might be related to point 3:
- You mentioned that gWCS might need work to support returning a `SpectralCoord`. I see from testing that I'm getting back what appears to be a `Quantity` when slicing an instance of the new class, rather than getting back a `SpectralCoord` as one gets e.g. a `SkyCoord` when slicing a `SkyCoordTableCoord`. It's not immediately clear to me where (in `SpectralFrame` / somewhere else in gWCS / in the lookup table code itself) is the place to fix this.

Unfortunately at this point I need to switch over to working on a big task for `specutils` that's due next week, but if you can add some clarity on any of the above, @Cadair, I could probably do some more work on this soon.

### DIFF
--- a/ndcube/extra_coords/lookup_table_coord.py
+++ b/ndcube/extra_coords/lookup_table_coord.py
@@ -5,12 +5,13 @@ import astropy.units as u
 import gwcs
 import gwcs.coordinate_frames as cf
 import numpy as np
-from astropy.coordinates import SkyCoord
+from astropy.coordinates import SkyCoord, SpectralCoord
 from astropy.modeling import models
 from astropy.modeling.models import tabular_model
 from astropy.time import Time
 
-__all__ = ['TimeTableCoordinate', 'SkyCoordTableCoordinate', 'QuantityTableCoordinate']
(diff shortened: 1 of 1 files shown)


Cadair:
Thanks for this @rosteen I will have a look at it tomorrow.

PR #415, opened 2021-03-25
Link ExtraCoords to the parent Cube

### LINKED ISSUE #349: Extra Coords not removed by NDCube slicing when their axis is sliced away

### LINKED ISSUE #408: Supply global_coords during NDCube Initialization?
### Description

In order to stay consistent with `extra_coords`, should it be possible to supply a `GlobalCoords` instance while initializing and `NDCube`?  If so, what should the API look like?

Currently when an `NDCube` is initialized, an empty `GlobalCoords` instance is created and attached via the `extra_coords` attribute.  Global coords must then be added subsequently.

### LINKED ISSUE #417: Fix the ordering of array_coordinates when adding ExtraCoords
Ensure that array_dimensions is in array order not world order in `ExtraCoords.add`

Cadair:
Once upon a time I decided to throw up a quick fix for #349. Then I realised that to fix that issue the `ExtraCoords` object needed to  know the dimensionality of the array, which means it needs a reference to the cube like `GlobalCoords`. So this PR addresses this by making the following changes:

* Change the constructor of `ExtraCoords` to take one optional argument `ndcube`.
* Remove the extra_coords kwarg from `NDCube` the canonical way to add coords to the `ExtraCoords` or `GlobalCoords` is `NDCube.xxxx_coords.add`.
* Write a descriptor for the `_extra_coords` and `_global_coords` attributes which when a coords object is assigned to the attribute sets the `._ndcube` attribute of the object. This means that all our internal code which sets the hidden attributes automatically ensures the parent reference is correct.
* Fix the cause of #349 which was `ExtraCoords.mapping` using array coordinates not pixel coordinates. Add a regression test for this as well.

This therefore fixes #349 and also fixes #417

It also closes #408 by deciding no for both `GlobalCoords` and `ExtraCoords`; however, it makes it real easy to change our minds in the future, by changing the descriptors to be not prefixed with `_`.

### DIFF
--- a/ndcube/ndcube.py
+++ b/ndcube/ndcube.py
@@ -137,7 +137,16 @@ class NDCubeLinkedDescriptor:
         self._property_name = None
 
     def __set_name__(self, owner, name):
+        """
+        This function is called when the class the descriptor is attached to is initialized.
+
+        The *class* and not the instance.
+        """
+        # property name is the name of the attribute on the parent class
+        # pointing at an instance of this descriptor.
         self._property_name = name
+        # attribute name is the name of the attribute on the parent class where
+        # the data is stored.
         self._attribute_name = f"_{name}"
 
     def __get__(self, obj, objtype=None):
@@ -155,7 +164,9 @@ class NDCubeLinkedDescriptor:
         elif issubclass(value, self._default_type):
             value = value(obj)
         else:
-            raise ValueError("wat")
+            raise ValueError(
+                f"Unable to set value for {self._property_name} it should "
+                f"be an instance or subclass of {self._default_type}")
 
         setattr(obj, self._attribute_name, value)
 
@@ -206,6 +217,7 @@ class NDCubeBase(NDCubeSlicingMixin, NDCubeABC):
         Default is False.
 
     """
+    # Instances of Extra and Global coords are managed through descriptors
     _extra_coords = NDCubeLinkedDescriptor(ExtraCoords)
     _global_coords = NDCubeLinkedDescriptor(GlobalCoords)
 
--- a/ndcube/ndcube.py
+++ b/ndcube/ndcube.py
@@ -128,6 +128,38 @@ class NDCubeABC(astropy.nddata.NDData, metaclass=NDCubeMetaClass):
         """
 
 
+class NDCubeLinkedDescriptor:
+    """
+    A descriptor which gives the property a reference to the cube to which it is attached.
+    """
+    def __init__(self, default_type):
+        self._default_type = default_type
+        self._property_name = None
+
+    def __set_name__(self, owner, name):
+        self._property_name = name
+        self._attribute_name = f"_{name}"
+
+    def __get__(self, obj, objtype=None):
+        if obj is None:
+            return
+
+        if getattr(obj, self._attribute_name, None) is None and self._default_type is not None:
+            self.__set__(obj, self._default_type)
+
+        return getattr(obj, self._attribute_name)
+
+    def __set__(self, obj, value):
+        if isinstance(value, self._default_type):
+            value._ndcube = obj
+        elif issubclass(value, self._default_type):
+            value = value(obj)
+        else:
+            raise ValueError("wat")
+
+        setattr(obj, self._attribute_name, value)
+
+
 class NDCubeBase(NDCubeSlicingMixin, NDCubeABC):
     """
     Class representing N-D data described by a single array and set of WCS transformations.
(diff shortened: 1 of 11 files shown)


Cadair:
Some of the tests are janky because of gWCS versions and fixes... I will investigate.

Cadair:
@DanRyanIrish This is good to go when you get a chance.

ndcube/ndcube.py
```diff
@@ -128,6 +128,38 @@ def crop_by_values(self, lower_corner, upper_corner, units=None, wcs=None):
...
+
+        return getattr(obj, self._attribute_name)
+
+    def __set__(self, obj, value):
+        if isinstance(value, self._default_type):
+            value._ndcube = obj
+        elif issubclass(value, self._default_type):
+            value = value(obj)
+        else:
+            raise ValueError("wat")
```
DanRyanIrish, line 158: More helpful error meesage here ;)

ndcube/ndcube.py
```diff
@@ -128,6 +128,38 @@ def crop_by_values(self, lower_corner, upper_corner, units=None, wcs=None):
...
+    """
+    A descriptor which gives the property a reference to the cube to which it is attached.
+    """
+    def __init__(self, default_type):
+        self._default_type = default_type
+        self._property_name = None
+
+    def __set_name__(self, owner, name):
+        self._property_name = name
+        self._attribute_name = f"_{name}"
```
DanRyanIrish, line 141: Might be helpful to add a comment here explaining different between `_property_name` and `_attribute_name`.

ndcube/ndcube.py
```diff
@@ -174,9 +206,11 @@ class NDCubeBase(NDCubeSlicingMixin, NDCubeABC):
         Default is False.
 
     """
+    _extra_coords = NDCubeLinkedDescriptor(ExtraCoords)
```
DanRyanIrish, line 209: Add comment here like, `# Initializing empty ExtraCoords and GlobalCoords instances on NDCube.`

By the by, why are these lines floating on their own and not in `__init__`.  Is it for the same reason as in astropy modeling (whatever that reason actually is?)

DanRyanIrish (approved):
Approved pending a couple minor comments

DanRyanIrish:
Does this fix #417 as well?

ndcube/ndcube.py
Cadair: These are class level attributes not instance level attributes (as they would be if they were `self.whatever` in the constructor). All instances of this class share the one instance of this descriptor object.

PR #433, opened 2021-06-22
Initial implementation for validating two WCS

### LINKED ISSUE #428: Validate WCS Function
### Description

Create a function that accepts two WCS objects, `target_wcs` and `initial_wcs`, and verifies that `target` is the same/equivalent to `initial`.  (Better variable names here are welcome.)

Initially, the `target_wcs` must describe all axes of the `initial_wcs`.  Future versions of this function should handle invariant axes.

The use case for this function is as part of verifying the input to the future `NDCube.resample` method.

adwaitbhope:
### Description

Fixes #428.

An initial implementation of the validate function for checking if two WCS objects are compatible with each other to reproject an `NDCube` on top of another.

### DIFF
--- a/ndcube/utils/wcs.py
+++ b/ndcube/utils/wcs.py
@@ -437,8 +437,7 @@ def validate_wcs(source_wcs, target_wcs):
 
     Parameters
     ----------
-    source_wcs : `astropy.wcs.WCS` or any other object that implements
-                    `astropy.wcs.wcsapi.BaseHighLevelWCS` or `astropy.wcs.wcsapi.BaseLowLevelWCS`
+    source_wcs : `astropy.wcs.wcsapi.BaseHighLevelWCS` or `astropy.wcs.wcsapi.BaseLowLevelWCS`
         The WCS which is currently in use, usually `self.wcs`.
 
     target_wcs : `astropy.wcs.WCS` or any other object that implements
--- a/ndcube/utils/wcs.py
+++ b/ndcube/utils/wcs.py
@@ -431,7 +431,7 @@ def array_indices_for_world_objects(wcs, axes=None):
     return tuple(ai for ai in array_indices if ai)
 
 
-def validate_wcs(source_wcs, target_wcs):
+def compare_wcs_physical_types(source_wcs, target_wcs):
     """
     Checks if two WCS objects are comptible with each other for reprojecting an NDCube on another.
 
@@ -440,9 +440,8 @@ def validate_wcs(source_wcs, target_wcs):
     source_wcs : `astropy.wcs.wcsapi.BaseHighLevelWCS` or `astropy.wcs.wcsapi.BaseLowLevelWCS`
         The WCS which is currently in use, usually `self.wcs`.
 
-    target_wcs : `astropy.wcs.WCS` or any other object that implements
-                    `astropy.wcs.wcsapi.BaseHighLevelWCS` or `astropy.wcs.wcsapi.BaseLowLevelWCS`
-        The WCS object on which the NDCube is to reprojected.
+    target_wcs : `astropy.wcs.wcsapi.BaseHighLevelWCS` or `astropy.wcs.wcsapi.BaseLowLevelWCS`
+        The WCS object on which the NDCube is to be reprojected.
 
     Returns
     -------
--- a/ndcube/utils/wcs.py
+++ b/ndcube/utils/wcs.py
@@ -8,7 +8,7 @@ import numbers
 from collections import UserDict
 
 import numpy as np
-from astropy.wcs.wcsapi import low_level_api, BaseHighLevelWCS, BaseLowLevelWCS
+from astropy.wcs.wcsapi import BaseHighLevelWCS, BaseLowLevelWCS, low_level_api
 
 __all__ = ['array_indices_for_world_objects', 'convert_between_array_and_pixel_axes',
            'calculate_world_indices_from_axes', 'wcs_ivoa_mapping',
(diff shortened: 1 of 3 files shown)


ndcube/utils/wcs.py
```diff
@@ -429,3 +429,36 @@ def array_indices_for_world_objects(wcs, axes=None):
         array_index = convert_between_array_and_pixel_axes(pixel_index, wcs.pixel_n_dim)
         array_indices[oinds] = tuple(array_index[::-1])  # Invert from pixel order to array order
     return tuple(ai for ai in array_indices if ai)
+
+
+def validate_wcs(source_wcs, target_wcs):
```
Cadair, line 434: I would name this something more descriptive like `compare_wcs_physical_types`

ndcube/utils/wcs.py
```diff
@@ -429,3 +429,36 @@ def array_indices_for_world_objects(wcs, axes=None):
...
+    """
+    Checks if two WCS objects are comptible with each other for reprojecting an NDCube on another.
+
+    Parameters
+    ----------
+    source_wcs : `astropy.wcs.WCS` or any other object that implements
+                    `astropy.wcs.wcsapi.BaseHighLevelWCS` or `astropy.wcs.wcsapi.BaseLowLevelWCS`
```
Cadair, line 441: I would simplify both of these to just this:
```suggestion
    source_wcs : `astropy.wcs.wcsapi.BaseHighLevelWCS` or `astropy.wcs.wcsapi.BaseLowLevelWCS`
```

Cadair (commented):
logic looks fine. Couple of small comments and needs tests :smile:

adwaitbhope:
I've added test cases for this function, but for the positive test case, we have the same object as source and target. Should we introduce another WCS in the fixtures?

ndcube/utils/wcs.py
DanRyanIrish: I came to this conversation late and I know @adwaitbhope has already made this change.  But I think the function name should say what is wants to do, rather than how it does it.  If not `validate_wcs`, then something like `assert_wcs_compatible`.

In the future we may want to do this comparison in a more sophisticated way, e.g. taking into account invariant axes, or comparing against the axes object types.  If the function name does not refer to its implementation but rrather the role it performs, changing how we validate should have minimal impact on the code outside this function.

ndcube/tests/test_utils_wcs.py
```diff
@@ -157,3 +157,11 @@ def test_array_indices_for_world_objects_2(wcs_4d_lt_t_l_ln):
     array_indices = utils.wcs.array_indices_for_world_objects(wcs_4d_lt_t_l_ln, ('lon', 'time'))
     assert len(array_indices) == 2
     assert array_indices == ((0, 3), (2,))
+
+
+def test_compare_wcs_physical_types(wcs_4d_t_l_lt_ln, wcs_3d_l_lt_ln):
+    wcs_compatibility = utils.wcs.compare_wcs_physical_types(wcs_4d_t_l_lt_ln, wcs_4d_t_l_lt_ln)
+    assert wcs_compatibility is True
```
DanRyanIrish, line 164:
```suggestion
    assert utils.wcs.compare_wcs_physical_types(wcs_4d_t_l_lt_ln, wcs_4d_t_l_lt_ln) is True
```

ndcube/tests/test_utils_wcs.py
```diff
@@ -157,3 +157,11 @@ def test_array_indices_for_world_objects_2(wcs_4d_lt_t_l_ln):
...
+
+    wcs_compatibility = utils.wcs.compare_wcs_physical_types(wcs_4d_t_l_lt_ln, wcs_3d_l_lt_ln)
+    assert wcs_compatibility is False
```
DanRyanIrish, line 167:
```suggestion
    utils.wcs.compare_wcs_physical_types(wcs_4d_t_l_lt_ln, wcs_3d_l_lt_ln) is False
```

DanRyanIrish:
> Should we introduce another WCS in the fixtures?

I don't think it's necessary at this stage.  The test as is adequately tests the current implementation.  If/when this function moves to a more complicated implementation, adapting the test should be part of the work.  Adding a new fixture would be part of that.

ndcube/tests/test_utils_wcs.py
adwaitbhope: Fixed

ndcube/tests/test_utils_wcs.py
adwaitbhope: Fixed

ndcube/utils/wcs.py
Cadair: I suspect that any further questions we are going to want to ask of the WCSes, i.e. invariant axes etc are going to want to be different functions. Also by convention functions which start `assert_` tend to raise assertion errors if the comparison fails.

Maybe the best thing to do here is change the first line of the docstring to make the scope of this function clear?

DanRyanIrish: Sounds fine to me.

ndcube/utils/wcs.py
```diff
@@ -429,3 +429,51 @@ def array_indices_for_world_objects(wcs, axes=None):
...
+        return wcs.low_level_wcs
+    elif isinstance(wcs, BaseLowLevelWCS):
+        return wcs
+    else:
+        raise(f'{name} must implement either BaseHighLevelWCS or BaseLowLevelWCS')
+
+
+def compare_wcs_physical_types(source_wcs, target_wcs):
+    """
+    Checks if two WCS objects are comptible with each other for reprojecting an NDCube on another.
```
Cadair, line 461:
```suggestion
    Checks to see if two WCS objects have the same physical types in the same order.
```

Cadair (approved):

PR #434, opened 2021-06-22
Reproject implementation

### LINKED ISSUE #429: Initial NDCube.resample Method
### Description

Create a method on `NDCube` to resample the data grid based on a new WCS.
* Should take a target APE-14-compliant WCS object from the user to which the data should be reprojected.  Initially this WCS should describe all axes of the cube.  Future versions should handle invariant axes, but not as part of this issue.
* Verify the target WCS is compatible with `NDCube`'s WCS.  Use the function outlined in #428
* Use `reproject.reproject_interp` to reproject `NDCube.data`.
* Return a new `NDCube` using the target WCS, the projected data array, the `NDCube.meta` and the `global_coords`.  Drop the `uncertainty`, `mask` and `extra_coords` information.  (Future versions can handle these cases.)

adwaitbhope:
### Description

Fixes #429.

For now, this is basically a wrapper around reproject, will add tests later.
Trying to figure out a way to compute `shape_out` from available data.

### DIFF
--- a/changelog/434.feature.rst
+++ b/changelog/434.feature.rst
@@ -1,1 +1,1 @@
-Adds functionality to reproject `NDCube` to use coordinates described by another WCS or FITS Header.
+Adds functionality to reproject `NDCube` to coordinates described by another WCS or FITS Header.
--- a/changelog/434.feature.rst
+++ b/changelog/434.feature.rst
@@ -1,1 +1,1 @@
-Adds functionality to reproject `NDCube` to coordinates described by another WCS or FITS Header.
+Adds functionality to reproject an `~.NDCube` object to coordinates described by another WCS or FITS Header by calling the new `~.NDCube.reproject_to` method.
--- a/changelog/434.feature.rst
+++ b/changelog/434.feature.rst
@@ -1,0 +1,1 @@
+Adds functionality to reproject `NDCube` to use coordinates described by another WCS or FITS Header.
--- a/ndcube/ndcube.py
+++ b/ndcube/ndcube.py
@@ -665,8 +665,10 @@ class NDCubeBase(NDCubeSlicingMixin, NDCubeABC):
             The WCS object to which the ``NDCube`` is to be reprojected.
 
         shape_out: `tuple`, optional
-            The shape of the output data. If not specified, `~astropy.wcs.wcsapi.BaseLowLevelWCS.pixel_shape` attribute (if available)
-            from the low level API of the ``target_wcs`` is used.
+            The shape of the output data array. The ordering of the dimensions must follow NumPy
+            ordering and not the WCS pixel shape.
+            If not specified, `~astropy.wcs.wcsapi.BaseLowLevelWCS.array_shape` attribute
+            (if available) from the low level API of the ``target_wcs`` is used.
 
         order: `int` or `str`
             The order of the interpolation. This can be any of: 'nearest-neighbor', 'bilinear',
--- a/ndcube/ndcube.py
+++ b/ndcube/ndcube.py
@@ -677,7 +677,7 @@ class NDCubeBase(NDCubeSlicingMixin, NDCubeABC):
             including a memory map, which may be helpful when dealing with extremely large files.
 
         return_footprint: `bool`
-            Whether to return the footprint in addition to the output array.
+            Whether to return the footprint in addition to the output NDCube.
 
         Returns
         -------
--- a/ndcube/ndcube.py
+++ b/ndcube/ndcube.py
@@ -688,8 +688,8 @@ class NDCubeBase(NDCubeSlicingMixin, NDCubeABC):
 
         Notes
         -----
-        This method doesn't support handling of `mask`, `extra_coords`, and `uncertainty` yet.
-        However, `meta` and `global_coords` are retained.
+        This method doesn't support handling of the ``mask``, ``extra_coords``, and ``uncertainty`` attributes yet.
+        However, ``meta`` and ``global_coords`` are copied to the output ``NDCube``.
         """
 
         if not utils.wcs.compare_wcs_physical_types(self.wcs, target_wcs):
--- a/ndcube/ndcube.py
+++ b/ndcube/ndcube.py
@@ -669,7 +669,7 @@ class NDCubeBase(NDCubeSlicingMixin, NDCubeABC):
             from the low level API of the ``target_wcs`` is used.
 
         order: `int` or `str`
-            The order of the interpolation. This can be any of: 'nearest-neighbour', 'bilinear',
+            The order of the interpolation. This can be any of: 'nearest-neighbor', 'bilinear',
             'biquadratic', or 'bicubic'.
 
         output_array: `numpy.ndarray`, optional
--- a/ndcube/ndcube.py
+++ b/ndcube/ndcube.py
@@ -711,7 +711,7 @@ class NDCubeBase(NDCubeSlicingMixin, NDCubeABC):
             if hasattr(low_level_target_wcs, 'pixel_shape') and low_level_target_wcs.pixel_shape is not None:
                 shape_out = low_level_target_wcs.pixel_shape
             else:
-                raise Exception("shape_out must be specified if target_wcs's low level API "
+                raise ValueError("shape_out must be specified if target_wcs's low level API "
                                 "does not have the pixel_shape attribute.")
 
         data = reproject_interp(self, output_projection=target_wcs,
--- a/ndcube/ndcube.py
+++ b/ndcube/ndcube.py
@@ -702,8 +702,8 @@ class NDCubeBase(NDCubeSlicingMixin, NDCubeABC):
             if hasattr(low_level_target_wcs, 'pixel_shape') and low_level_target_wcs.pixel_shape is not None:
                 shape_out = low_level_target_wcs.pixel_shape
             else:
-                raise Exception('shape_out must be specified if target_wcs\'s low level API '
-                                'does not have the pixel_shape attribute.')
+                raise Exception("shape_out must be specified if target_wcs's low level API "
+                                "does not have the pixel_shape attribute.")
 
         data = reproject_interp(self, output_projection=target_wcs,
                                 shape_out=shape_out, order=order,
--- a/ndcube/ndcube.py
+++ b/ndcube/ndcube.py
@@ -16,6 +16,7 @@ except ImportError:
     pass
 from astropy.wcs.wcsapi import BaseHighLevelWCS, HighLevelWCSWrapper
 from astropy.wcs.wcsapi.wrappers import SlicedLowLevelWCS
+from reproject import reproject_interp
 
 from ndcube import utils
 from ndcube.extra_coords import ExtraCoords
@@ -651,6 +652,51 @@ class NDCubeBase(NDCubeSlicingMixin, NDCubeABC):
             result_cubes.append(sliced_cube)
         # Creating a new NDCubeSequence with the result_cubes and common axis as axis
         return NDCubeSequence(result_cubes, meta=self.meta)
+
+    def resample(self, target_wcs, shape_out, order='bilinear', output_array=None, return_footprint=True):
+        """
+        Separates slices of NDCubes along a given axis into an NDCubeSequence of (N-1)DCubes.
+
+        Parameters
+        ----------
+        target_wcs : `astropy.wcs.wcsapi.BaseHighLevelWCS` or `astropy.wcs.wcsapi.BaseLowLevelWCS`
+            The WCS object on which the NDCube is to be reprojected.
+
+        shape_out: `tuple`
+            The shape of the output data.
+
+        order: `int or str`
+            The order of the interpolation. This can be any of: 'nearest-neighbour', 'bilinear',
+            'biquadratic', or 'bicubic'.
+
+        output_array: `None or numpy.ndarray`
+            An array in which to store the reprojected data. This can be any numpy array
+            including a memory map, which may be helpful when dealing with extremely large files.
+
+        return_footprint: `bool`
+            Whether to return the footprint in addition to the output array.
+
+        Returns
+        -------
+        resampled_cube : `ndcube.NDCube`
+            A new resultant NDCube object.
+
+        footprint: `numpy.ndarray`
+            Footprint of the input array in the output array. Values of 0 indicate no coverage or
+            valid values in the input image, while values of 1 indicate valid values.
+        """
+
+        if not utils.wcs.compare_wcs_physical_types(self.wcs, target_wcs):
+            raise('Given target_wcs is not compatible with this NDCube.')
+
+        resampled_data, footprint = reproject_interp(self, output_projection=target_wcs,
+                                          shape_out=shape_out, order=order,
+                                          output_array=output_array, return_footprint=return_footprint)
+
+        resampled_cube = NDCube(resampled_data, wcs=target_wcs, meta=deepcopy(self.meta))
+        resampled_cube._global_coords = deepcopy(self.global_coords)
+
+        return resampled_cube, footprint
 
 
 class NDCube(NDCubeBase, astropy.nddata.NDArithmeticMixin):
--- a/ndcube/ndcube.py
+++ b/ndcube/ndcube.py
@@ -693,7 +693,7 @@ class NDCubeBase(NDCubeSlicingMixin, NDCubeABC):
         """
 
         if not utils.wcs.compare_wcs_physical_types(self.wcs, target_wcs):
-            raise Exception('Given target_wcs is not compatible with this NDCube.')
+            raise Exception('Given target_wcs is not compatible with this NDCube, the physical types do not match.')
 
         low_level_target_wcs = utils.wcs.get_low_level_wcs(target_wcs, 'target_wcs')
 
--- a/ndcube/ndcube.py
+++ b/ndcube/ndcube.py
@@ -689,15 +689,24 @@ class NDCubeBase(NDCubeSlicingMixin, NDCubeABC):
         if not utils.wcs.compare_wcs_physical_types(self.wcs, target_wcs):
             raise('Given target_wcs is not compatible with this NDCube.')
 
-        resampled_data, footprint = reproject_interp(self, output_projection=target_wcs,
-                                                     shape_out=shape_out, order=order,
-                                                     output_array=output_array,
-                                                     return_footprint=return_footprint)
+        if return_footprint:
+            resampled_data, footprint = reproject_interp(self, output_projection=target_wcs,
+                                                         shape_out=shape_out, order=order,
+                                                         output_array=output_array,
+                                                         return_footprint=return_footprint)
+        else:
+            resampled_data = reproject_interp(self, output_projection=target_wcs,
+                                              shape_out=shape_out, order=order,
+                                              output_array=output_array,
+                                              return_footprint=return_footprint)
 
         resampled_cube = NDCube(resampled_data, wcs=target_wcs, meta=deepcopy(self.meta))
         resampled_cube._global_coords = deepcopy(self.global_coords)
 
-        return resampled_cube, footprint
+        if return_footprint:
+            return resampled_cube, footprint
+
+        return resampled_cube
 
 
 class NDCube(NDCubeBase, astropy.nddata.NDArithmeticMixin):
--- a/ndcube/ndcube.py
+++ b/ndcube/ndcube.py
@@ -662,7 +662,7 @@ class NDCubeBase(NDCubeSlicingMixin, NDCubeABC):
         ----------
         target_wcs : `astropy.wcs.wcsapi.BaseHighLevelWCS`, `astropy.wcs.wcsapi.BaseLowLevelWCS`,
             or `astropy.io.fits.Header`
-            The WCS object on which the ``NDCube`` is to be reprojected.
+            The WCS object to which the ``NDCube`` is to be reprojected.
 
         shape_out: `tuple`, optional
             The shape of the output data. If not specified, `~astropy.wcs.wcsapi.BaseLowLevelWCS.pixel_shape` attribute (if available)
--- a/ndcube/ndcube.py
+++ b/ndcube/ndcube.py
@@ -16,7 +16,6 @@ except ImportError:
     pass
 from astropy.wcs.wcsapi import BaseHighLevelWCS, HighLevelWCSWrapper
 from astropy.wcs.wcsapi.wrappers import SlicedLowLevelWCS
-from reproject import reproject_interp
 
 from ndcube import utils
 from ndcube.extra_coords import ExtraCoords
@@ -691,6 +690,10 @@ class NDCubeBase(NDCubeSlicingMixin, NDCubeABC):
         This method doesn't support handling of the ``mask``, ``extra_coords``, and ``uncertainty`` attributes yet.
         However, ``meta`` and ``global_coords`` are copied to the output ``NDCube``.
         """
+        try:
+            from reproject import reproject_interp
+        except ModuleNotFoundError:
+            raise ImportError("This method requires the optional package `reproject`.")
 
         if not utils.wcs.compare_wcs_physical_types(self.wcs, target_wcs):
             raise Exception('Given target_wcs is not compatible with this NDCube, the physical types do not match.')
--- a/ndcube/ndcube.py
+++ b/ndcube/ndcube.py
@@ -702,7 +702,7 @@ class NDCubeBase(NDCubeSlicingMixin, NDCubeABC):
         if return_footprint:
             data, footprint = data
 
-        resampled_cube = NDCube(resampled_data, wcs=target_wcs, meta=deepcopy(self.meta))
+        resampled_cube = NDCube(data, wcs=target_wcs, meta=deepcopy(self.meta))
         resampled_cube._global_coords = deepcopy(self.global_coords)
 
         if return_footprint:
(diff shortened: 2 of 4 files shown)


ndcube/ndcube.py
```diff
@@ -633,6 +634,52 @@ def explode_along_axis(self, axis):
...
+        if not utils.wcs.compare_wcs_physical_types(self.wcs, target_wcs):
+            raise('Given target_wcs is not compatible with this NDCube.')
+
+        resampled_data, footprint = reproject_interp(self, output_projection=target_wcs,
+                                                     shape_out=shape_out, order=order,
+                                                     output_array=output_array,
+                                                     return_footprint=return_footprint)
+
+        resampled_cube = NDCube(resampled_data, wcs=target_wcs, meta=deepcopy(self.meta))
+        resampled_cube._global_coords = deepcopy(self.global_coords)
```
DanRyanIrish, line 679: Does `meta` and `global_coords` have to be deepcopies here?

ndcube/ndcube.py
```diff
@@ -633,6 +634,52 @@ def explode_along_axis(self, axis):
...
+
+        return resampled_cube, footprint
```
DanRyanIrish, line 681: The check of `return_footprint` needs to be made here.  According to docstring, only when `return_footprint=True` will `footprint` be returned.

ndcube/ndcube.py
```diff
@@ -633,6 +634,52 @@ def explode_along_axis(self, axis):
...
+
+        footprint: `numpy.ndarray`
+            Footprint of the input array in the output array. Values of 0 indicate no coverage or
+            valid values in the input image, while values of 1 indicate valid values.
+        """
+
+        if not utils.wcs.compare_wcs_physical_types(self.wcs, target_wcs):
+            raise('Given target_wcs is not compatible with this NDCube.')
+
+        resampled_data, footprint = reproject_interp(self, output_projection=target_wcs,
```
DanRyanIrish, line 673: Handling of `return_footprints` needs to be done here.

DanRyanIrish:
I think this is a great start @adwaitbhope.  When you have a simple test confirming the ouput, we can talk about how next to proceed.

ndcube/ndcube.py
adwaitbhope: Right, missed that. In that case, should we handle this similar to reproject, or should we return `resampled_cube, None` if `return_footprint=False`?

ndcube/ndcube.py
adwaitbhope: Well, I thought it would be better to not have any references to the original cube's attributes.

ndcube/ndcube.py
adwaitbhope: I've added a commit that handles it like reproject for now.

ndcube/ndcube.py
Cadair: yeah probably wise.

ndcube/ndcube.py
```diff
@@ -633,6 +634,61 @@ def explode_along_axis(self, axis):
...
+        resampled_cube : `ndcube.NDCube`
+            A new resultant NDCube object.
+
+        footprint: `numpy.ndarray`
+            Footprint of the input array in the output array. Values of 0 indicate no coverage or
+            valid values in the input image, while values of 1 indicate valid values.
+        """
+
+        if not utils.wcs.compare_wcs_physical_types(self.wcs, target_wcs):
+            raise('Given target_wcs is not compatible with this NDCube.')
```
Cadair, line 671: You should add a test which raises this exception.

Cadair (changes requested):
Looking good @adwaitbhope just a few pretty minor comments :smile:

ndcube/ndcube.py
```diff
@@ -633,6 +634,61 @@ def explode_along_axis(self, axis):
         # Creating a new NDCubeSequence with the result_cubes and common axis as axis
         return NDCubeSequence(result_cubes, meta=self.meta)
 
+    def resample(self, target_wcs, shape_out, order='bilinear', output_array=None, return_footprint=True):
+        """
+        Separates slices of NDCubes along a given axis into an NDCubeSequence of (N-1)DCubes.
```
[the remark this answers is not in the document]
adwaitbhope: Oops, sorry. Copied parts of the docstring from elsewhere, fixed it now.

ndcube/ndcube.py
```diff
@@ -633,6 +634,61 @@ def explode_along_axis(self, axis):
...
+            Whether to return the footprint in addition to the output array.
+
+        Returns
+        -------
+        resampled_cube : `ndcube.NDCube`
+            A new resultant NDCube object.
+
+        footprint: `numpy.ndarray`
+            Footprint of the input array in the output array. Values of 0 indicate no coverage or
+            valid values in the input image, while values of 1 indicate valid values.
```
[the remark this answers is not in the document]
adwaitbhope: Added

ndcube/ndcube.py
```diff
@@ -633,6 +634,61 @@ def explode_along_axis(self, axis):
...
+        if return_footprint:
+            resampled_data, footprint = reproject_interp(self, output_projection=target_wcs,
+                                                         shape_out=shape_out, order=order,
+                                                         output_array=output_array,
+                                                         return_footprint=return_footprint)
+        else:
+            resampled_data = reproject_interp(self, output_projection=target_wcs,
+                                              shape_out=shape_out, order=order,
+                                              output_array=output_array,
+                                              return_footprint=return_footprint)
```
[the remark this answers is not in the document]
adwaitbhope: Yeah, this does look much cleaner. I'll make this change now, perhaps we could make a call about the footprint later?
Also, if we're going to calculate the footprint anyway, would it be better to return it as well?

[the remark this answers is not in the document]
Cadair: yeah the more I think about it exposing the option like this is probably the best way.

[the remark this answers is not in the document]
adwaitbhope: Maybe we could set the default argument to `False`?

ndcube/ndcube.py
```diff
@@ -633,6 +634,61 @@ def explode_along_axis(self, axis):
         # Creating a new NDCubeSequence with the result_cubes and common axis as axis
         return NDCubeSequence(result_cubes, meta=self.meta)
 
+    def resample(self, target_wcs, shape_out, order='bilinear', output_array=None, return_footprint=True):
```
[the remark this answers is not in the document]
adwaitbhope: In case both are available, should we treat `shape_out` as an override or go with `.pixel_shape` anyway?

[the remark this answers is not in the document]
Cadair: I would take `shape_out` to be an override. Explicit things should generally always override implicit things.

ndcube/ndcube.py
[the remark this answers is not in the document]
Cadair: sounds good.

ndcube/ndcube.py
```diff
@@ -629,6 +630,74 @@ def explode_along_axis(self, axis):
         # Creating a new NDCubeSequence with the result_cubes and common axis as axis
         return NDCubeSequence(result_cubes, meta=self.meta)
 
+    def reproject(self, target_wcs, shape_out=None, order='bilinear', output_array=None, return_footprint=False):
+        """
+        Reprojects an NDCube on another WCS object to upsample or downsample the pixel resolution.
```
Cadair, line 635:
```suggestion
        Reprojects this NDCube to the coordinates described by another WCS object.
```

ndcube/ndcube.py
```diff
@@ -629,6 +630,74 @@ def explode_along_axis(self, axis):
...
+
+        Parameters
+        ----------
+        target_wcs : `astropy.wcs.wcsapi.BaseHighLevelWCS` or `astropy.wcs.wcsapi.BaseLowLevelWCS`
+            The WCS object on which the NDCube is to be reprojected.
```
Cadair, line 640:
```suggestion
            The WCS object on which the ``NDCube`` is to be reprojected.
```

ndcube/ndcube.py
```diff
@@ -629,6 +630,74 @@ def explode_along_axis(self, axis):
...
+
+        shape_out: `tuple`, optional
+            The shape of the output data. If not specified, `pixel_shape` attribute (if available)
+            from the low level API of the `target_wcs` is used.
```
Cadair, line 644:
```suggestion
            The shape of the output data. If not specified, `~astropy.wcs.wcsapi.BaseLowLevelWCS.pixel_shape` attribute (if available)
            from the low level API of the ``target_wcs`` is used.
```

ndcube/ndcube.py
```diff
@@ -629,6 +630,74 @@ def explode_along_axis(self, axis):
...
+
+        order: `int or str`
```
Cadair, line 646:
```suggestion
        order: `int` or `str`
```

ndcube/ndcube.py
```diff
@@ -629,6 +630,74 @@ def explode_along_axis(self, axis):
...
+            The order of the interpolation. This can be any of: 'nearest-neighbour', 'bilinear',
+            'biquadratic', or 'bicubic'.
+
+        output_array: `None or numpy.ndarray`
```
Cadair, line 650:
```suggestion
        output_array: `numpy.ndarray`, optional
```

ndcube/ndcube.py
```diff
@@ -629,6 +630,74 @@ def explode_along_axis(self, axis):
...
+            An array in which to store the reprojected data. This can be any numpy array
+            including a memory map, which may be helpful when dealing with extremely large files.
+
+        return_footprint: `bool`
+            Whether to return the footprint in addition to the output array.
+
+        Returns
+        -------
+        resampled_cube : `ndcube.NDCube`
+            A new resultant NDCube object.
```
Cadair, line 660:
```suggestion
            A new resultant NDCube object, the supplied ``target_wcs`` will be the ``.wcs`` attribute of the output ``NDCube``.
```

ndcube/ndcube.py
```diff
@@ -629,6 +630,74 @@ def explode_along_axis(self, axis):
...
+
+        footprint: `numpy.ndarray`
+            Footprint of the input array in the output array. Values of 0 indicate no coverage or
+            valid values in the input image, while values of 1 indicate valid values.
+
+        Notes
+        -----
+        This method doesn't support handling of `mask`, `extra_coords`, and `uncertainty` yet.
+        However, `meta` and `global_coords` are retained.
```
Cadair, line 669:
```suggestion
        This method doesn't support handling of the ``mask``, ``extra_coords``, and ``uncertainty`` attributes yet.
        However, ``meta`` and ``global_coords`` are copied to the output ``NDCube``.
```

ndcube/ndcube.py
```diff
@@ -629,6 +630,74 @@ def explode_along_axis(self, axis):
...
+        """
+
+        if not utils.wcs.compare_wcs_physical_types(self.wcs, target_wcs):
+            raise Exception('Given target_wcs is not compatible with this NDCube.')
```
Cadair, line 673:
```suggestion
            raise Exception('Given target_wcs is not compatible with this NDCube, the physical types do not match.')
```

ndcube/ndcube.py
```diff
@@ -629,6 +630,74 @@ def explode_along_axis(self, axis):
...
+
+        low_level_target_wcs = utils.wcs.get_low_level_wcs(target_wcs, 'target_wcs')
+
+        # If shape_out is not specified explicity, try to extract it from the low level WCS
+        if not shape_out:
+            if hasattr(low_level_target_wcs, 'pixel_shape') and low_level_target_wcs.pixel_shape:
```
Cadair, line 679: You need to do the `is not None` any time you are comparing to a thing which could be an iterable.
```suggestion
            if hasattr(low_level_target_wcs, 'pixel_shape') and low_level_target_wcs.pixel_shape is not None:
```

ndcube/ndcube.py
```diff
@@ -629,6 +630,74 @@ def explode_along_axis(self, axis):
...
+                shape_out = low_level_target_wcs.pixel_shape
+            else:
+                raise Exception('shape_out must be specified if target_wcs\'s low level API '
+                                'does not have the pixel_shape attribute.')
```
Cadair, line 683:
```suggestion
                raise Exception("shape_out must be specified if target_wcs's low level API "
                                "does not have the pixel_shape attribute.")
```

Cadair (commented):
Mostly comments on the docstrings. In general in sphinx single backticks are references to things and double are plain verbatim text output. You should only use singles for things which resolve to Python objects.

I need to review the tests which hopefully I will do in a bit.

ndcube/tests/test_ndcube.py
```diff
@@ -520,3 +520,49 @@ def test_initialize_from_ndcube(ndcube_3d_l_ln_lt_ectime):
...
+    shape_out = (5, 20, 12, 8)
+
+    resampled_cube = ndcube_4d_ln_l_t_lt.reproject(target_wcs, shape_out)
+
+    assert ndcube_4d_ln_l_t_lt.data.shape == (5, 10, 12, 8)
+    assert resampled_cube.data.shape == (5, 20, 12, 8)
+
+    # Raises an exception if given target_wcs is not compatible with the NDCube
+    with pytest.raises(Exception):
+        _ = ndcube_4d_ln_l_t_lt.reproject(wcs_3d_lt_ln_l, shape_out)
```
DanRyanIrish, line 538: Could you move these lines to another test that just tests that the correct exception is raised when the WCS is incompatible?

DanRyanIrish (changes requested):
I think this PR will be good to go once these changes have been implemented.

Cadair (changes requested):
The tests look ok. I think the big part missing from this before we land is some narrative documentation. @DanRyanIrish What do yo think the best approach to that is?

ndcube/ndcube.py
adwaitbhope: Got it

ndcube/ndcube.py
```diff
@@ -629,6 +630,74 @@ def explode_along_axis(self, axis):
         # Creating a new NDCubeSequence with the result_cubes and common axis as axis
         return NDCubeSequence(result_cubes, meta=self.meta)
 
+    def reproject(self, target_wcs, shape_out=None, order='bilinear', output_array=None, return_footprint=False):
+        """
+        Reprojects this NDCube to the coordinates described by another WCS object.
+
+        Parameters
+        ----------
+        target_wcs : `astropy.wcs.wcsapi.BaseHighLevelWCS` or `astropy.wcs.wcsapi.BaseLowLevelWCS`
```
Cadair, line 639: I think we should also allow a dict / `fits.Header` instance here as well and turn it into a WCS?

ndcube/ndcube.py
```diff
@@ -629,6 +630,74 @@ def explode_along_axis(self, axis):
...
+
+        data = reproject_interp(self, output_projection=target_wcs,
+                                shape_out=shape_out, order=order,
+                                output_array=output_array,
+                                return_footprint=return_footprint)
+
+        if return_footprint:
+            data, footprint = data
+
+        resampled_cube = NDCube(data, wcs=target_wcs, meta=deepcopy(self.meta))
```
Cadair, line 693: We should return the type of ourselves so that when people subclass us they get the class back they are expecting.
```suggestion
        resampled_cube = type(self)(data, wcs=target_wcs, meta=deepcopy(self.meta))
```

ndcube/ndcube.py
```diff
@@ -629,6 +630,74 @@ def explode_along_axis(self, axis):
         # Creating a new NDCubeSequence with the result_cubes and common axis as axis
         return NDCubeSequence(result_cubes, meta=self.meta)
 
+    def reproject(self, target_wcs, shape_out=None, order='bilinear', output_array=None, return_footprint=False):
```
Cadair, line 633: From sunpy/sunpy#5470 Albert called this `reproject_to` I kind of like that name, thoughts on changing it?

ndcube/ndcube.py
Cadair:
```
isinstance(target_wcs, collections.abc.Mapping):
    target_wcs = astropy.wcs.WCS(target_wcs)
```

ndcube/ndcube.py
```diff
@@ -16,6 +16,7 @@
     pass
 from astropy.wcs.wcsapi import HighLevelWCSWrapper
 from astropy.wcs.wcsapi.wrappers import SlicedLowLevelWCS
+from reproject import reproject_interp
```
Cadair, line 19: Please move this into the reproject method.

setup.cfg
```diff
@@ -36,6 +36,7 @@ setup_requires =
 install_requires =
   astropy>=4.1
   gwcs>=0.15
+  reproject>=0.7.1
```
Cadair, line 39: move this into a new `reproject` extra in the block below.

ndcube/ndcube.py
```diff
@@ -652,6 +654,85 @@ def explode_along_axis(self, axis):
...
+
+        low_level_target_wcs = utils.wcs.get_low_level_wcs(target_wcs, 'target_wcs')
+
+        # If shape_out is not specified explicity, try to extract it from the low level WCS
+        if not shape_out:
+            if hasattr(low_level_target_wcs, 'pixel_shape') and low_level_target_wcs.pixel_shape is not None:
+                shape_out = low_level_target_wcs.pixel_shape
+            else:
+                raise Exception("shape_out must be specified if target_wcs's low level API "
+                                "does not have the pixel_shape attribute.")
```
ayshih, line 718: I believe these should all be `array_shape` instead of `pixel_shape`

ndcube/ndcube.py
```diff
@@ -652,6 +654,85 @@ def explode_along_axis(self, axis):
...
+        target_wcs : `astropy.wcs.wcsapi.BaseHighLevelWCS`, `astropy.wcs.wcsapi.BaseLowLevelWCS`,
+            or `astropy.io.fits.Header`
+            The WCS object on which the ``NDCube`` is to be reprojected.
+
+        shape_out: `tuple`, optional
+            The shape of the output data. If not specified, `~astropy.wcs.wcsapi.BaseLowLevelWCS.pixel_shape` attribute (if available)
+            from the low level API of the ``target_wcs`` is used.
+
+        order: `int` or `str`
+            The order of the interpolation. This can be any of: 'nearest-neighbour', 'bilinear',
```
ayshih, line 672:
```suggestion
            The order of the interpolation. This can be any of: 'nearest-neighbor', 'bilinear',
```
Presumably?

ndcube/ndcube.py
```diff
@@ -652,6 +654,85 @@ def explode_along_axis(self, axis):
...
+            'biquadratic', or 'bicubic'.
+
+        output_array: `numpy.ndarray`, optional
+            An array in which to store the reprojected data. This can be any numpy array
+            including a memory map, which may be helpful when dealing with extremely large files.
+
+        return_footprint: `bool`
+            Whether to return the footprint in addition to the output array.
```
ayshih, line 680:
```suggestion
            Whether to return the footprint in addition to the output NDCube.
```

ndcube/ndcube.py
```diff
@@ -652,6 +654,85 @@ def explode_along_axis(self, axis):
...
+
+        Parameters
+        ----------
+        target_wcs : `astropy.wcs.wcsapi.BaseHighLevelWCS`, `astropy.wcs.wcsapi.BaseLowLevelWCS`,
+            or `astropy.io.fits.Header`
+            The WCS object on which the ``NDCube`` is to be reprojected.
+
+        shape_out: `tuple`, optional
+            The shape of the output data. If not specified, `~astropy.wcs.wcsapi.BaseLowLevelWCS.pixel_shape` attribute (if available)
+            from the low level API of the ``target_wcs`` is used.
```
ayshih, line 669: You should specify the ordering of dimensions needs to be NumPy ordering

ndcube/ndcube.py
```diff
@@ -652,6 +654,85 @@ def explode_along_axis(self, axis):
...
 
+    def reproject_to(self, target_wcs, shape_out=None, order='bilinear', output_array=None, return_footprint=False):
+        """
+        Reprojects this NDCube to the coordinates described by another WCS object.
+
+        Parameters
+        ----------
+        target_wcs : `astropy.wcs.wcsapi.BaseHighLevelWCS`, `astropy.wcs.wcsapi.BaseLowLevelWCS`,
+            or `astropy.io.fits.Header`
+            The WCS object on which the ``NDCube`` is to be reprojected.
```
ayshih, line 665:
```suggestion
            The WCS object to which the ``NDCube`` is to be reprojected.
```

ayshih (commented):
Also, don't raise the generic `Exception`, but instead raise `ValueError` or the like.

DanRyanIrish:
Thanks for these comments @ayshih.  @adwaitbhope, we should definitely address these before merging.

changelog/434.feature.rst
```diff
@@ -0,0 +1 @@
+Adds functionality to reproject `NDCube` to use coordinates described by another WCS or FITS Header.
```
DanRyanIrish, line 1:
```suggestion
Adds functionality to reproject `NDCube` to coordinates described by another WCS or FITS Header.
```

ndcube/ndcube.py
```diff
@@ -652,6 +654,85 @@ def explode_along_axis(self, axis):
...
+        try:
+            from reproject import reproject_interp
+        except ModuleNotFoundError:
+            raise ImportError("This method requires the optional package `reproject`.")
+
+        if isinstance(target_wcs, Mapping):
+            try:
+                target_wcs = WCS(header=target_wcs)
+            except Exception:
+                raise Exception('Invalid header supplied; unable to create a WCS object.')
```
DanRyanIrish, line 705:
```suggestion
            target_wcs = WCS(header=target_wcs)
```
I don't think a `try...except``` is needed here.  If the header is invalid, `WCS` will raise an error with a more informative traceback.

ndcube/ndcube.py
```diff
@@ -652,6 +654,85 @@ def explode_along_axis(self, axis):
...
+
+        if not utils.wcs.compare_wcs_physical_types(self.wcs, target_wcs):
+            raise Exception('Given target_wcs is not compatible with this NDCube, the physical types do not match.')
```
DanRyanIrish, line 708:
```suggestion
            raise ValueError('Given target_wcs is not compatible with this NDCube, the physical types do not match.')
```

ndcube/ndcube.py
```diff
@@ -652,6 +654,85 @@ def explode_along_axis(self, axis):
...
+
+        low_level_target_wcs = utils.wcs.get_low_level_wcs(target_wcs, 'target_wcs')
+
+        # If shape_out is not specified explicity, try to extract it from the low level WCS
+        if not shape_out:
+            if hasattr(low_level_target_wcs, 'pixel_shape') and low_level_target_wcs.pixel_shape is not None:
+                shape_out = low_level_target_wcs.pixel_shape
+            else:
+                raise Exception("shape_out must be specified if target_wcs's low level API "
```
DanRyanIrish, line 717:
```suggestion
                raise ValueError("shape_out must be specified if target_wcs's low level API "
```

DanRyanIrish (changes requested):

ndcube/ndcube.py
adwaitbhope: Yes, looking back at it, I think it should be `array_shape`. Will change it.

ndcube/ndcube.py
adwaitbhope: Sure

ndcube/ndcube.py
```diff
@@ -652,6 +654,82 @@ def explode_along_axis(self, axis):
         # Creating a new NDCubeSequence with the result_cubes and common axis as axis
         return NDCubeSequence(result_cubes, meta=self.meta)
 
+    def reproject_to(self, target_wcs, shape_out=None, order='bilinear', output_array=None, return_footprint=False):
+        """
+        Reprojects this NDCube to the coordinates described by another WCS object.
+
+        Parameters
+        ----------
+        target_wcs : `astropy.wcs.wcsapi.BaseHighLevelWCS`, `astropy.wcs.wcsapi.BaseLowLevelWCS`,
```
adwaitbhope, line 663: Should we use a better name for `target_wcs` now that we're accepting `fits.Header`? Maybe `target_projection` or something.

DanRyanIrish: I think it's fine.

DanRyanIrish (approved):

DanRyanIrish:
@Cadair, do think this is ready to merge now?

ndcube/ndcube.py
```diff
@@ -652,6 +654,84 @@ def explode_along_axis(self, axis):
...
+
+        low_level_target_wcs = utils.wcs.get_low_level_wcs(target_wcs, 'target_wcs')
+
+        # If shape_out is not specified explicity, try to extract it from the low level WCS
+        if not shape_out:
+            if hasattr(low_level_target_wcs, 'array_shape') and low_level_target_wcs.array_shape is not None:
+                shape_out = low_level_target_wcs.array_shape
+            else:
+                raise ValueError("shape_out must be specified if target_wcs's low level API "
+                                 "does not have the array_shape attribute.")
```
Cadair, line 717:
```suggestion
                raise ValueError("shape_out must be specified if target_wcs does not have the array_shape attribute.")
```

Cadair (approved):

PR #436, opened 2021-07-04
Combine cubes from NDCubeSequence using reproject

### LINKED ISSUE #430: Reproject NDCubeSequence to NDCube
### Description

Create a method on `NDCubeSequence` that:
* Uses `NDCube.resample` (#429) reprojects each `NDCube` in the sequence to the WCS of the first `NDCube`;
* Stacks the data arrays along an additional 0th array dimension
* Generates a new WCS that adds an initial array axis to the WCS of the first cube.  Let this new axis simply be in pixel units.  A later issue can generate physical coordinates for this axis based on user input of cube global_coords.
* Returns a new (n+1)-D `NDCube` using the just the reprojected and stacked data array and the newly generated WCS object.  Drop all other information to start with.

adwaitbhope:
### Description

Fixes #430.

I'm currently only assuming one sequence axis, that part needs to be handled better.
I'm also exposing an `axis` parameter for stacking data which will affect the `mapping` passed on to the `CompoundLowLevelWCS`.

### DIFF
--- a/changelog/436.feature.rst
+++ b/changelog/436.feature.rst
@@ -1,0 +1,1 @@
+Adds functionality to stack cubes of an `~.NDCubeSequence` object together by reprojecting them to a common WCS and adding an extra dimension corresponding to the sequence axis.
--- a/changelog/436.feature.rst
+++ b/changelog/436.feature.rst
@@ -1,1 +1,1 @@
-Adds functionality to stack cubes of an `~.NDCubeSequence` object together by reprojecting them to a common WCS and adding an extra dimension corresponding to the sequence axis.
+Adds functionality to stack cubes of an `~.NDCubeSequence` object together into a single `.NDCube` object, by reprojecting them to a common WCS and adding an extra dimension corresponding to the sequence axis.
--- a/ndcube/ndcube_sequence.py
+++ b/ndcube/ndcube_sequence.py
@@ -393,6 +393,11 @@ class NDCubeSequenceBase:
                 table_coords.append(SkyCoordTableCoordinate(SkyCoord(axis_coords),
                                                             names=axis_name))
 
+            else:
+                raise Exception(f"Unable to convert global_coords '{axis_name}' into an axis. "
+                                 "global_coords must be either `astropy.units.Quantity`, "
+                                 "`astropy.time.Time`, or `astropy.coordinates.SkyCoord`.")
+
         return MultipleTableCoordinate(*table_coords).wcs
 
     def combine_cubes(self, common_wcs_index=0):
--- a/ndcube/ndcube_sequence.py
+++ b/ndcube/ndcube_sequence.py
@@ -408,9 +408,6 @@ class NDCubeSequenceBase:
 
         Parameters
         ----------
-        axis: `int`, optional
-            The axis along which the data is to be stacked.
-
         common_wcs_index: `int`, optional
             The index of the `~ndcube.NDCube` whose WCS is to be used as the base on which to
             reproject all other `~ndcube.NDCube`s in the `~ndcube.NDCubeSequence`.
--- a/ndcube/ndcube_sequence.py
+++ b/ndcube/ndcube_sequence.py
@@ -423,9 +423,9 @@ class NDCubeSequenceBase:
         # Stack data of all cubes together
         combined_data = np.stack([cube.data for cube in reprojected_cubes], axis=0)
 
-        sequence_axes_wcs = self.__get_sequence_axes_wcs()
+        sequence_axes_wcs = self._get_sequence_axes_wcs()
 
-        mapping =list(range(target_wcs.pixel_n_dim))
+        mapping = list(range(target_wcs.pixel_n_dim))
 
         # Sequence axis can map to multiple world axes
         mapping.extend([target_wcs.pixel_n_dim] * sequence_axes_wcs.pixel_n_dim)
--- a/ndcube/ndcube_sequence.py
+++ b/ndcube/ndcube_sequence.py
@@ -9,7 +9,7 @@ from astropy.time import Time
 
 from ndcube import utils
 from ndcube.visualization import PlotterDescriptor
-from ndcube.extra_coords.table_coord import (QuantityTableCoordinate,
+from ndcube.extra_coords.table_coord import (MultipleTableCoordinate, QuantityTableCoordinate,
                                              SkyCoordTableCoordinate, TimeTableCoordinate)
 from ndcube.wcs.wrappers.compound_wcs import CompoundLowLevelWCS
 
@@ -374,28 +374,19 @@ class NDCubeSequenceBase:
             [slice(0, n_cubes)] + [slice(start, stop) for start, stop in zip(starts, stops)])
 
     def __get_sequence_axes_wcs(self):
-        combined_table_coord = None
+        table_coords = []
 
         for axis_name, axis_coords in self.sequence_axis_coords.items():
-            table_coord = None
-
             if isinstance(axis_coords[0], u.Quantity):
-                table_coord = QuantityTableCoordinate(u.Quantity(axis_coords))
+                table_coords.append(QuantityTableCoordinate(u.Quantity(axis_coords)))
 
             elif isinstance(axis_coords[0], Time):
-                table_coord = TimeTableCoordinate(Time(axis_coords))
+                table_coords.append(TimeTableCoordinate(Time(axis_coords)))
 
             elif isinstance(axis_coords[0], SkyCoordTableCoordinate):
-                table_coord = SkyCoordTableCoordinate(SkyCoord(axis_coords))
+                table_coords.append(SkyCoordTableCoordinate(SkyCoord(axis_coords)))
 
-            if table_coord:
-                if not combined_table_coord:
-                    combined_table_coord = table_coord
-                else:
-                    combined_table_coord = combined_table_coord & table_coord
-
-        if combined_table_coord:
-            return combined_table_coord.wcs
+        return MultipleTableCoordinate(*table_coords).wcs
 
     def combine_cubes(self, common_wcs_index=0):
         """
--- a/ndcube/ndcube_sequence.py
+++ b/ndcube/ndcube_sequence.py
@@ -382,17 +382,22 @@ class NDCubeSequenceBase:
         table_coords = []
 
         for axis_name, axis_coords in self.sequence_axis_coords.items():
+            physical_type = self[0].global_coords._internal_coords[axis_name][0]
+
             if isinstance(axis_coords[0], (u.Quantity, numbers.Number)):
                 table_coords.append(QuantityTableCoordinate(u.Quantity(axis_coords),
-                                                            names=axis_name))
+                                                            names=axis_name,
+                                                            physical_types=physical_type))
 
             elif isinstance(axis_coords[0], Time):
                 table_coords.append(TimeTableCoordinate(Time(axis_coords),
-                                                        names=axis_name))
+                                                        names=axis_name,
+                                                        physical_types=physical_type))
 
             elif isinstance(axis_coords[0], SkyCoordTableCoordinate):
                 table_coords.append(SkyCoordTableCoordinate(SkyCoord(axis_coords),
-                                                            names=axis_name))
+                                                            names=axis_name,
+                                                            physical_types=physical_type))
 
             else:
                 warnings.warn(f"Skipping '{axis_name}'. Unable to convert global_coords "
@@ -404,7 +409,8 @@ class NDCubeSequenceBase:
         # If no global coords were found, use the cube index to form an axis.
         if not table_coords:
             table_coords.append(QuantityTableCoordinate(u.Quantity(np.arange(len(self.data))),
-                                                        names='sequence'))
+                                                        names='sequence',
+                                                        physical_types='meta.obs.sequence'))
 
         return MultipleTableCoordinate(*table_coords).wcs
 
--- a/ndcube/ndcube_sequence.py
+++ b/ndcube/ndcube_sequence.py
@@ -1,6 +1,7 @@
 import copy
 import numbers
 import textwrap
+import warnings
 
 import astropy.units as u
 import numpy as np
@@ -394,9 +395,15 @@ class NDCubeSequenceBase:
                                                             names=axis_name))
 
             else:
-                raise Exception(f"Unable to convert global_coords '{axis_name}' into an axis. "
-                                "global_coords must be either `astropy.units.Quantity`, "
-                                "`astropy.time.Time`, or `astropy.coordinates.SkyCoord`.")
+                warnings.warn(f"Skipping '{axis_name}'. Unable to convert global_coords "
+                              f"'{axis_name}' into an axis. global_coords must be either "
+                              "`astropy.units.Quantity`, `astropy.time.Time`, or "
+                              "`astropy.coordinates.SkyCoord`.",
+                              UserWarning)
+
+        if not table_coords:
+            table_coords.append(QuantityTableCoordinate(u.Quantity(np.arange(len(self.data))),
+                                                        names='sequence'))
 
         return MultipleTableCoordinate(*table_coords).wcs
 
--- a/ndcube/ndcube_sequence.py
+++ b/ndcube/ndcube_sequence.py
@@ -422,8 +422,8 @@ class NDCubeSequenceBase:
         # Reproject all cubes to a common WCS
         target_wcs = self[common_wcs_index].wcs
         shape_out = self[common_wcs_index].data.shape
-        reprojected_cubes = [cube.reproject(target_wcs, shape_out=shape_out,
-                                            return_footprint=False) for cube in self]
+        reprojected_cubes = [cube.reproject_to(target_wcs, shape_out=shape_out,
+                                               return_footprint=False) for cube in self]
 
         # Stack data of all cubes together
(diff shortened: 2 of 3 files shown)


adwaitbhope:
Will add tests soon.

ndcube/ndcube_sequence.py
```diff
@@ -239,6 +241,46 @@ def explode_along_axis(self, axis):
...
+        target_wcs = self[common_wcs_index].wcs
+        shape_out = self[common_wcs_index].data.shape
+        reprojected_cubes = [cube.reproject(target_wcs, shape_out=shape_out,
+                                            return_footprint=False) for cube in self]
+
+        # Stack data of all cubes together
+        combined_data = np.stack([cube.data for cube in reprojected_cubes], axis=axis)
+
+        sequence_axis = list(self.sequence_axis_coords.values())[0]
+        sequence_axis_wcs = QuantityTableCoordinate(u.Quantity(sequence_axis)).wcs
```
Cadair, line 276: I think you want to dispatch on type here. Depending on what type the `sequence_axis_coords` are will depend on what type of table coordinate you want to make.

ndcube/ndcube_sequence.py
```diff
@@ -239,6 +241,46 @@ def explode_along_axis(self, axis):
...
+
+        # Create a Compound WCS object
+        compound_wcs = CompoundLowLevelWCS(target_wcs, sequence_axis_wcs,
+                                           mapping=[i for i in range(target_wcs.pixel_n_dim + 1)])
```
Cadair, line 280: I would expect that you would need to involve `axis` in this computation. If you haven't put the new pixel axis first then the new WCS axis doesn't want to be last?

ndcube/ndcube_sequence.py
```diff
@@ -239,6 +241,46 @@ def explode_along_axis(self, axis):
...
+        # Reproject all cubes to a common WCS
+        target_wcs = self[common_wcs_index].wcs
+        shape_out = self[common_wcs_index].data.shape
+        reprojected_cubes = [cube.reproject(target_wcs, shape_out=shape_out,
+                                            return_footprint=False) for cube in self]
+
+        # Stack data of all cubes together
+        combined_data = np.stack([cube.data for cube in reprojected_cubes], axis=axis)
+
+        sequence_axis = list(self.sequence_axis_coords.values())[0]
```
Cadair, line 275: I don't really see any reason why we should only use the first world axis along the sequence here? What about if there is two coords along the sequence axis?

@DanRyanIrish is this a thing which can even happen?

Cadair, line 275: The fact you drop to `values()` and then rebuild a quantity here is not good. I think the fix is to dispatch on type of table coordinate like I mention on the line below.

ndcube/ndcube_sequence.py
```diff
@@ -239,6 +241,46 @@ def explode_along_axis(self, axis):
...
+    def combine_cubes(self, axis=0, common_wcs_index=0):
+        """
+        Reprojects all `~ndcube.NDCube` objects to a common WCS, and stacks the data together
+        to return a single (N+1)-dimensional `~ndcube.NDCube` with an associated Compound WCS
+        containing extra axes corresponding to the sequence axis.
+
+        Parameters
+        ----------
+        axis: `int`, optional
+            The axis along which the data is to be stacked.
```
Cadair, line 253: This is a little confusing. What this is is the *array* axis we are stacking along.

I do wonder if we even want this to be a configurable parameter. If we are being purists about it then the slowest varying coordinate should go last (which it is a safe assumption is the sequence dimension). I would propose maybe just dropping this parameter all together. I wonder if @DanRyanIrish or @wtbarnes have differing opinions on that though.

ndcube/ndcube_sequence.py
```diff
@@ -239,6 +241,46 @@ def explode_along_axis(self, axis):
...
+
+        common_wcs_index: `int`, optional
```
Cadair, line 255: I would put this parameter first, it's definitely the most important one.

Cadair (commented):
This looks like a great start. I have only looked at the main stack function at the moment.

By biggest high-level comment would be that we probably want to be a little smarter about how we generate the new combined WCS. I have left a couple of comments inline with specific suggestions. It also might make sense to have a function which generates that new WCS which we can then swap out if/when we add support for other ways of adding WCS dimensions.

ndcube/ndcube_sequence.py
adwaitbhope: Yeah, this computation needs to be improved. It would also depend on whether we want to drop the axis parameter entirely as you mentioned in the other comment.

ndcube/ndcube_sequence.py
adwaitbhope: I've used the first world axis just to simplify things for getting a working prototype.

Since `sequence_axis_coords` extracts coordinates from the `global_coords` of every cube, I think it's possible to have more than one world axis. The construction of a compound WCS would be a little more complicated in that case.

ndcube/ndcube_sequence.py
adwaitbhope: Yup, will be doing that.

ndcube/ndcube_sequence.py
adwaitbhope: `values()` actually still returns a list of astropy objects like `Time` or `Quantity`. This list is wrapped in one single `Quantity` object (where type checking needs to be done).

adwaitbhope: I've made some updates that can handle multiple coords along the sequence axis.

ndcube/ndcube_sequence.py
adwaitbhope: For now, I've dropped the axis parameter to make things a little simpler while computing the pixel to world axis mapping.

ndcube/ndcube_sequence.py
```diff
@@ -239,6 +244,67 @@ def explode_along_axis(self, axis):
...
+        shape_out = self[common_wcs_index].data.shape
+        reprojected_cubes = [cube.reproject(target_wcs, shape_out=shape_out,
+                                            return_footprint=False) for cube in self]
+
+        # Stack data of all cubes together
+        combined_data = np.stack([cube.data for cube in reprojected_cubes], axis=0)
+
+        sequence_axes_wcs = self.__get_sequence_axes_wcs()
+
+        mapping = [i for i in range(target_wcs.pixel_n_dim)]
```
DanRyanIrish, line 299:
```suggestion
        mapping =list(range(target_wcs.pixel_n_dim))
```

DanRyanIrish (commented):
I think this is looking good.

ndcube/ndcube_sequence.py
```diff
@@ -239,6 +244,67 @@ def explode_along_axis(self, axis):
         # creating a new sequence with the result_cubes keeping the meta and common axis as axis
         return self._new_instance(result_cubes, common_axis=new_common_axis, meta=self.meta)
 
+    def __get_sequence_axes_wcs(self):
```
Cadair, line 247: Don't think the double underscore is really needed.

```suggestion
    def _get_sequence_axes_wcs(self):
```

ndcube/ndcube_sequence.py
```diff
@@ -239,6 +244,67 @@ def explode_along_axis(self, axis):
...
+                table_coords.append(QuantityTableCoordinate(u.Quantity(axis_coords),
+                                                            names=axis_name))
+
+            elif isinstance(axis_coords[0], Time):
+                table_coords.append(TimeTableCoordinate(Time(axis_coords),
+                                                        names=axis_name))
+
+            elif isinstance(axis_coords[0], SkyCoordTableCoordinate):
+                table_coords.append(SkyCoordTableCoordinate(SkyCoord(axis_coords),
+                                                            names=axis_name))
```
Cadair, line 265: This probably needs an `else: get angry` clause.

Cadair (commented):
This looks like it's heading in the right direction, I would like to test it out myself tomorrow.

ndcube/ndcube_sequence.py
```diff
@@ -239,6 +244,72 @@ def explode_along_axis(self, axis):
...
+
+            else:
+                raise Exception(f"Unable to convert global_coords '{axis_name}' into an axis. "
```
Cadair, line 268: convert this to a warning.

ndcube/ndcube_sequence.py
```diff
@@ -239,6 +244,72 @@ def explode_along_axis(self, axis):
...
+                                "global_coords must be either `astropy.units.Quantity`, "
+                                "`astropy.time.Time`, or `astropy.coordinates.SkyCoord`.")
+
```
Cadair, line 271:
```
if not table_coords:
    table_coords.append(QuantityTableCoordinate(u.Quantity(np.arange(length_of_sequence)),
                                                names=axis_name))

File: ndcube/ndcube_sequence.py [hunk 11]
```diff
@@ -239,6 +244,72 @@ def explode_along_axis(self, axis):
...
 
+    def _get_sequence_axes_wcs(self):
+        """
+        Returns a WCS object corresponding to the sequence axis of this NDCubeSequence.
+        It can contain multiple pixel axes that map to the same world axis.
+        """
+        table_coords = []
+
+        for axis_name, axis_coords in self.sequence_axis_coords.items():
+            if isinstance(axis_coords[0], u.Quantity):
```
Cadair, line 255:
```suggestion
            if isinstance(axis_coords[0], (u.Quantity, numbers.Number)):
```

ndcube/ndcube_sequence.py
```diff
@@ -4,8 +4,13 @@
 
 import astropy.units as u
 import numpy as np
+from astropy.coordinates.sky_coordinate import SkyCoord
```
Cadair, line 7:
```suggestion
from astropy.coordinates import SkyCoord
```

ndcube/ndcube_sequence.py
```diff
@@ -239,6 +244,72 @@ def explode_along_axis(self, axis):
...
+            The index of the `~ndcube.NDCube` whose WCS is to be used as the base on which to
+            reproject all other `~ndcube.NDCube`s in the `~ndcube.NDCubeSequence`.
+
+        Returns
+        -------
+        `ndcube.NDCube`
+            A combined (N+1)D cube with stacked data of all cubes.
+        """
+        from ndcube import NDCube  # adding this here to prevent a circular import
+
```
DanRyanIrish, line 292: Insert a check here to ensure all cubes have the same shape.  For now, if cubes have different shapes, raise a `ValueError`.

(`NDCubeSequence` supports some limited cases were cubes are of different shapes.)

ndcube/ndcube_sequence.py
```diff
@@ -239,6 +245,85 @@ def explode_along_axis(self, axis):
...
+                warnings.warn(f"Skipping '{axis_name}'. Unable to convert global_coords "
+                              f"'{axis_name}' into an axis. global_coords must be either "
+                              "`astropy.units.Quantity`, `astropy.time.Time`, or "
+                              "`astropy.coordinates.SkyCoord`.",
+                              UserWarning)
+
+        # If no global coords were found, use the cube index to form an axis.
+        if not table_coords:
+            table_coords.append(QuantityTableCoordinate(u.Quantity(np.arange(len(self.data))),
+                                                        names='sequence'))
```
DanRyanIrish, line 278: Set `physical_type='meta.obs.sequence'`

ndcube/ndcube_sequence.py
```diff
@@ -239,6 +245,85 @@ def explode_along_axis(self, axis):
...
+        """
+        Returns a WCS object corresponding to the sequence axis of this NDCubeSequence.
+        It can contain multiple pixel axes that map to the same world axis.
+        """
+        table_coords = []
+
+        for axis_name, axis_coords in self.sequence_axis_coords.items():
+            if isinstance(axis_coords[0], (u.Quantity, numbers.Number)):
+                table_coords.append(QuantityTableCoordinate(u.Quantity(axis_coords),
+                                                            names=axis_name))
```
DanRyanIrish, line 258: In each case, try to pull the `physical_type` of each coord from the first cube in the sequence and use it to set the `physical_type` of each `TableCoordinate`.

DanRyanIrish (commented):
I think this looks good.  I just have a request regarding extracting the `physical_types`.  See in-line comments.

DanRyanIrish:
This also needs a changelog.

ndcube/ndcube_sequence.py
DanRyanIrish: @adwaitbhope Any progress on this?

adwaitbhope: I have now. I'm extracting the physical type from `global_coords`. But since we're iterating through `.items()` of the `sequence_axis_coords` dictionary, their order doesn't seem to be the same at every run.

I'm sorting the `wcs.world_axis_physical_types` in the tests due to this, is that alright?

changelog/436.feature.rst
```diff
@@ -0,0 +1 @@
+Adds functionality to stack cubes of an `~.NDCubeSequence` object together by reprojecting them to a common WCS and adding an extra dimension corresponding to the sequence axis.
```
Cadair, line 1:
```suggestion
Adds functionality to stack cubes of an `~.NDCubeSequence` object together into a single `.NDCube` object, by reprojecting them to a common WCS and adding an extra dimension corresponding to the sequence axis.
```

ndcube/ndcube_sequence.py
```diff
@@ -239,6 +245,91 @@ def explode_along_axis(self, axis):
         # creating a new sequence with the result_cubes keeping the meta and common axis as axis
         return self._new_instance(result_cubes, common_axis=new_common_axis, meta=self.meta)
 
+    def _get_sequence_axes_wcs(self):
+        """
+        Returns a WCS object corresponding to the sequence axis of this NDCubeSequence.
+        It can contain multiple pixel axes that map to the same world axis.
```
Cadair, line 251: Isn't this backwards? We have one pixel axis (the sequence axis) but can have many world coordinates (the sequence axis)

ndcube/ndcube_sequence.py
```diff
@@ -239,6 +245,91 @@ def explode_along_axis(self, axis):
...
+        """
+        table_coords = []
+
+        for axis_name, axis_coords in self.sequence_axis_coords.items():
+            physical_type = self[0].global_coords._internal_coords[axis_name][0]
```
Cadair, line 256: Not all global coords are in `_internal_coords`, i.e. ones sliced out of a WCS. I don't think this is the best way to do this.

Wouldn't just `self[0].global_coords.physical_types[axis_name]` work?

Cadair (commented):
I think this looks good. I wonder if @wtbarnes has had a chance to test it :thinking:

ndcube/ndcube_sequence.py
adwaitbhope: Right, yes (I'm gonna go check if there are more such instances!)

ndcube/ndcube_sequence.py
adwaitbhope: It does - I think I missed this property while going through the code. Updated it now.

DanRyanIrish:
@wtbarnes have you or are you planning to test this out?

wtbarnes:
> @wtbarnes have you or are you planning to test this out?

Sorry I probably won't have time in the next day or two.

ndcube/ndcube_sequence.py
```diff
@@ -239,6 +245,91 @@ def explode_along_axis(self, axis):
...
+        if not table_coords:
+            table_coords.append(QuantityTableCoordinate(u.Quantity(np.arange(len(self.data))),
+                                                        names='sequence',
+                                                        physical_types='meta.obs.sequence'))
+
+        return MultipleTableCoordinate(*table_coords).wcs
+
+    def combine_cubes(self, common_wcs_index=0):
+        """
+        Reprojects all `~ndcube.NDCube` objects to a common WCS, and stacks the data together
```
hayesla, line 290:
```suggestion
        Reprojects all `~ndcube.NDCube` objects in the `~ndcube.NDCubeSequence` to a common WCS, and stacks the data together
```

ndcube/ndcube_sequence.py
```diff
@@ -239,6 +245,91 @@ def explode_along_axis(self, axis):
...
+        # Verify if all cubes have the same shape.
+        cube_shape = self[0].data.shape
+        for cube in self:
+            if cube.data.shape != cube_shape:
+                raise ValueError('All cubes must have the same shape.')
+
+        # Reproject all cubes to a common WCS
+        target_wcs = self[common_wcs_index].wcs
+        shape_out = self[common_wcs_index].data.shape
+        reprojected_cubes = [cube.reproject_to(target_wcs, shape_out=shape_out,
```
hayesla, line 316: maybe a warning or something should be flagged here if this reproject is going to take a long time. Imagine a user has a sequence of a hundred 4Kx4K AIA images and then tried to run this, it will take a looonng time

hayesla (commented):
This is great - just needs to be rebased with main.

My only concern is (I'm wearing a map hat here) is that this can take a long time for 2D images. This is of course down to `reproject_to` here, but even with like 16 AIA images at 256x256 took 7 mins on my local machine. In this way, I think a user warning or something could be added to warn of this?

but other than taking a long time, its working as expected!

PR #438, opened 2021-07-16
Modify crop so all corners of the ROI are used

### LINKED ISSUE #418: Robustly ensure the behaviour of `crop` is correct.
SunPy's Map has had unlimited trouble with `submap` we need to make sure our semantics are correct before release: https://github.com/sunpy/sunpy-SEP/pull/51#discussion_r533597518

Cadair:
This vendors in https://github.com/astropy/astropy/pull/11950 until we can rely on importing it.

Fixes #418

### DIFF
--- a/ndcube/ndcube.py
+++ b/ndcube/ndcube.py
@@ -1,6 +1,7 @@
 import abc
 import textwrap
 import warnings
+import itertools
 from copy import deepcopy
 from collections import namedtuple
 
@@ -618,7 +619,7 @@ class NDCubeBase(NDCubeSlicingMixin, NDCubeABC):
         """
         Convert two corners of a bounding box to the points of all corners.
         """
-        return lower_corner_values, upper_corner_values
+        return tuple(itertools.product(*zip(lower_corner_values, upper_corner_values)))
 
     def _crop_from_points(self, *world_points_values, wcs):
         if isinstance(wcs, BaseHighLevelWCS):
--- a/ndcube/ndcube.py
+++ b/ndcube/ndcube.py
@@ -622,6 +622,24 @@ class NDCubeBase(NDCubeSlicingMixin, NDCubeABC):
         return tuple(itertools.product(*zip(lower_corner_values, upper_corner_values)))
 
     def _crop_from_points(self, *world_points_values, wcs):
+        """
+        Crop the cube such that all the specified world points fall within it.
+
+        Parameters
+        ----------
+        world_points_values
+            The world coordinates in wcsapi "values" form (i.e. arrays /
+            floats), for however many world points should be contained in the
+            output cube.
+
+        wcs : `~astropy.wcs.wcsapi.BaseHighLevelWCS`, `~astropy.wcs.wcsapi.BaseLowLevelWCS`
+            The WCS to use to convert the world coordinates to array indices.
+
+        Returns
+        -------
+        new_cube
+            A new ``NDCube`` instance.
+        """
         if isinstance(wcs, BaseHighLevelWCS):
             wcs = wcs.low_level_wcs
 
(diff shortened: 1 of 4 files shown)


Cadair:
This could probably do with a more advanced 3D regression test, but I have to go start my long weekend.

ndcube/ndcube.py
```diff
@@ -587,26 +612,39 @@ def _crop(self, lower_corner, upper_corner, wcs, crop_by_values):
...
+
+        return lower_corner, upper_corner
+
+    def _bounding_box_to_points(self, lower_corner_values, upper_corner_values, wcs):
+        """
+        Convert two corners of a bounding box to the points of all corners.
+        """
+        return tuple(itertools.product(*zip(lower_corner_values, upper_corner_values)))
+
+    def _crop_from_points(self, *world_points_values, wcs):
```
DanRyanIrish, line 624: I think this method, even though it's private, deserves a docstring since it's the core of the `crop` methods and could in the future be exposed.

DanRyanIrish (approved):
This looks good.  I only request a docstring for `crop_by_points` that explains what it does and it's inputs and outputs.  Although it's "private", it's the core of the algorithm and I think a docstring would be valuable.

ndcube/ndcube.py
```diff
@@ -587,26 +612,57 @@ def _crop(self, lower_corner, upper_corner, wcs, crop_by_values):
...
+        """
+        Crop the cube such that all the specified world points fall within it.
```
DanRyanIrish, line 626:
```suggestion
        Crop to the minimum cube in pixel-space that contains all the specified world points.
```

The aim of this suggestion is to clarify that:
1. The new cube is defined in pixel-space
2. It is the minimum region that contains all the points, given the constraint of 1.

ndcube/ndcube.py
```diff
@@ -587,26 +612,57 @@ def _crop(self, lower_corner, upper_corner, wcs, crop_by_values):
...
+
+        Parameters
+        ----------
+        world_points_values
+            The world coordinates in wcsapi "values" form (i.e. arrays /
+            floats), for however many world points should be contained in the
+            output cube.
```
DanRyanIrish, line 633:
```suggestion
            output cube.  The number of objects is dictated by the number of world axes,
            while the length of the objects is dictated by the number of points.
```
The aim of this suggestion is to make it easy for readers to understand how to format the input.

ndcube/ndcube.py
```diff
@@ -587,26 +612,57 @@ def _crop(self, lower_corner, upper_corner, wcs, crop_by_values):
...
+
+    def _crop_from_points(self, *world_points_values, wcs):
+        """
+        Crop the cube such that all the specified world points fall within it.
+
+        Parameters
+        ----------
+        world_points_values
+            The world coordinates in wcsapi "values" form (i.e. arrays /
+            floats), for however many world points should be contained in the
```
DanRyanIrish, line 632: Is values form really arrays/floats in the wcsapi?  Not Quantities?  Is it just the values form in ndcube is quantities?

DanRyanIrish (changes requested):

ndcube/ndcube.py
Cadair: That's not strictly true. I will clarify.

Cadair, line 633:
```suggestion
            The world coordinates in wcsapi "values" form (i.e. arrays /
            floats), for however many world points should be contained in the
            output cube. Each argument should be a tuple with number of 
            coordinates equal to the number of world axes.
```

ndcube/ndcube.py
Cadair: yeah, it can also be quantities, but I think they are just treated as arrays.

DanRyanIrish (approved):

PR #443, opened 2021-07-23
Remove test for invalid header

Cadair:
This test was really asserting a warning (which was being promoted to an error). I don't think we really need it we can just delegate this behaviour to the WCS constructor.


Cadair:
(I can only test the dist builds after merge)

PR #449, opened 2021-07-29
Resampled wcs

DanRyanIrish:
### Description

This PR updates `ResampledWCS` to account for the fact that pixel indices represent the centres of pixels and therefore a shift must be applied when converting between pixel grids.

Fixes #<Issue Number>

### DIFF
--- a/ndcube/wcs/wrappers/resampled_wcs.py
+++ b/ndcube/wcs/wrappers/resampled_wcs.py
@@ -21,8 +21,8 @@ class ResampledLowLevelWCS(BaseWCSWrapper):
     def __init__(self, wcs, factor):
         self._wcs = wcs
         if np.isscalar(factor):
-            factor = np.array([factor] * self.pixel_n_dim)
-        self._factor = factor
+            factor = [factor] * self.pixel_n_dim
+        self._factor = np.array(factor)
 
     def _top_to_underlying_pixels(self, top_pixels):
         # Convert user-facing pixel indices to the pixel grid of underlying WCS.
--- a/ndcube/wcs/wrappers/resampled_wcs.py
+++ b/ndcube/wcs/wrappers/resampled_wcs.py
@@ -65,7 +65,7 @@ class ResampledLowLevelWCS(BaseWCSWrapper):
 
     @property
     def pixel_shape(self):
-        return tuple(np.asarray(self._wcs.pixel_shape) / self._factor)
+        return tuple((np.asarray(self._wcs.pixel_shape) / self._factor).astype(int))
 
     @property
     def pixel_bounds(self):
--- a/ndcube/wcs/wrappers/resampled_wcs.py
+++ b/ndcube/wcs/wrappers/resampled_wcs.py
@@ -14,6 +14,7 @@ class ResampledLowLevelWCS(BaseWCSWrapper):
     ----------
     wcs : `~astropy.wcs.wcsapi.BaseLowLevelWCS`
         The original WCS for which to reorder axes
+
     factor : int or float or iterable
         The factor by which to increase the pixel size for each pixel
         axis. If a scalar, the same factor is used for all axes.
@@ -36,7 +37,6 @@ class ResampledLowLevelWCS(BaseWCSWrapper):
         # Subtractive factor makes sure the correct sub-pixel location is returned.
         factor_shape = list(self._factor.shape) + [1] * (underlying_pixels.ndim - 1)
         factor = self._factor.reshape(factor_shape)
-        #factor = self._factor
         return (underlying_pixels - (factor - 1) / 2) / factor
 
     def pixel_to_world_values(self, *pixel_arrays):
--- a/ndcube/wcs/wrappers/resampled_wcs.py
+++ b/ndcube/wcs/wrappers/resampled_wcs.py
@@ -43,8 +43,8 @@ class ResampledLowLevelWCS(BaseWCSWrapper):
 
     def _underlying_to_top_pixels(self, underlying_pixels):
         # Convert pixel indices of underlying pixel grid to user-facing grid.
-        factor = self._pad_dims(self._factor, top_pixels.ndim)
-        offset = self._pad_dims(self._offset, top_pixels.ndim)
+        factor = self._pad_dims(self._factor, underlying_pixels.ndim)
+        offset = self._pad_dims(self._offset, underlying_pixels.ndim)
         return (underlying_pixels - offset) / factor
 
     def _pad_dims(self, arr, ndim):
@@ -65,7 +65,12 @@ class ResampledLowLevelWCS(BaseWCSWrapper):
 
     @property
     def pixel_shape(self):
-        return tuple((np.asarray(self._wcs.pixel_shape) / self._factor).astype(int))
+        # Return pixel shape of resampled grid.
+        # Where shape is an integer, return an int type as its required for some uses.
+        underlying_shape = np.asarray(self._wcs.pixel_shape)
+        int_elements = np.mod(underlying_shape, self._factor) == 0
+        pixel_shape = underlying_shape / self._factor
+        return tuple(int(i) if is_int else i for i, is_int in zip(pixel_shape, int_elements))
 
     @property
     def pixel_bounds(self):
--- a/ndcube/wcs/wrappers/resampled_wcs.py
+++ b/ndcube/wcs/wrappers/resampled_wcs.py
@@ -53,7 +53,6 @@ class ResampledLowLevelWCS(BaseWCSWrapper):
 
     @property
     def pixel_bounds(self):
-        if self._wcs.pixel_bounds is None:
-            return self._wcs.pixel_bounds
-        top_level_bounds = self._underlying_to_top_pixels(np.asarray(self._wcs.pixel_bounds))
-        return [tuple(bounds) for bounds in top_level_bounds]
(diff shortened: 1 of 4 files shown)


DanRyanIrish:
@Cadair, here are my changes to `ResampledWCS`.  Tests still need fixing.  But before I invest that effort, any comments on what I'm doing?

ndcube/wcs/wrappers/resampled_wcs.py
```diff
@@ -13,34 +14,47 @@ class ResampledLowLevelWCS(BaseWCSWrapper):
...
     factor : int or float or iterable
         The factor by which to increase the pixel size for each pixel
         axis. If a scalar, the same factor is used for all axes.
     """
     def __init__(self, wcs, factor):
         self._wcs = wcs
         if np.isscalar(factor):
-            factor = [factor] * self.pixel_n_dim
+            factor = np.array([factor] * self.pixel_n_dim)
         self._factor = factor
```
Cadair, line 26:
```suggestion
            factor = [factor] * self.pixel_n_dim
        self._factor = np.array(factor)
```

Cadair (commented):
I haven't sat down and worked through the maths of the translation (do you want me to?), but looks good otherwise.

DanRyanIrish:
> I haven't sat down and worked through the maths of the translation (do you want me to?), but looks good otherwise.

A check on the maths would be welcome but not expected.

DanRyanIrish:
In going through the tests, I've come across an important point regarding `pixel_bounds`.  My understanding is that `pixel_bounds` refer to the center of the pixels at either end of the limit of the axis.  Currently, `ResampledWCS` just translates the centre of the underlying `pixel_bounds` to the new pixel grid.  However, this does not represent that limit of the new pixel grid for which the transformations are defined.

Consider a 1-D pixel grid that we upsample by a factor of 1/3, i.e. there are 3 new pixels for every old/underlying pixel.  Say the pixel bounds were `(-1, 2)`.  The location of the centre of these pixels in the new pixel grid is `(-3, 6)`.  However, the full extent of the defined range includes the edges of these pixels.  Defined with pixel edges, the `pixel_bounds` `(-1.5, 2.5)`.  In the new pixel grid, this corresponds to `(-4.5, 7.5)`.  This means that in the new pixel grid, pixels `-4 -- 7` (inclusive) are fully within the pixel bounds of the original pixel grid.  However, `pixel_bounds` currently returns `(-3, 6)`, implying that the transformations aren't defined for new pixels `-4` or `7`.

In the above example I chose numbers that worked out nicely in whole pixels.  But this might not always be the case.  We must therefore clarify what `pixel_bounds` represents.  Is it:
* the centre of the last whole pixel included in the valid range?  (If this is the case then resampling can cause the valid range in world coords to shrink a little if a non-integer number of new pixels fit into the pixel bounds.)
* the centre the the last pixel that has at least some of its area in the valid range?  (If this is the case then resampling can cause the valid range in world coords to increase a little for the same reason as above.)
* the non-integer pixel value representing the edge of the valid range.  If this is the case, the pixel bounds can always be kept the same in world coords between the new and old pixel grids.

DanRyanIrish:
According to @ayshih the answer to the above is option 3.

DanRyanIrish:
It was realised that the additive/subtractive factors introduced by this PR shold not be included.  They were adding because symmetry around pixel 0 was overlooked.  As this was the main point of this PR, it has been closed.

ndcube/wcs/wrappers/resampled_wcs.py
```diff
@@ -13,34 +13,68 @@ class ResampledLowLevelWCS(BaseWCSWrapper):
...
+        top_pixel_arrays = self._underlying_to_top_pixels(np.asarray(underlying_pixel_arrays))
+        return tuple(array for array in top_pixel_arrays)
 
     @property
     def pixel_shape(self):
-        return tuple(self._wcs.pixel_shape[i] / self._factor[i]
-                     for i in range(self.pixel_n_dim))
+        # Return pixel shape of resampled grid.
+        # Where shape is an integer, return an int type as its required for some uses.
+        underlying_shape = np.asarray(self._wcs.pixel_shape)
```
Cadair, line 70: The underlying pixel shape can be None, so this needs a conditional like `pixel_bounds`

Cadair (changes requested):
Looks good, and I think I understand the new algorithm. I think this at least needs new tests for `offset=`? also does it not need tests which demonstrate that the new algorithm is required?

DanRyanIrish:
Any further comments before merging this @Cadair?  (The failing test will be resolved once #451 is merged.)

ndcube/wcs/wrappers/resampled_wcs.py
```diff
@@ -13,34 +13,70 @@ class ResampledLowLevelWCS(BaseWCSWrapper):
...
-        return tuple(self._wcs.pixel_shape[i] / self._factor[i]
-                     for i in range(self.pixel_n_dim))
+        # Return pixel shape of resampled grid.
+        # Where shape is an integer, return an int type as its required for some uses.
+        if self._wcs.pixel_shape is None:
+            return self._wcs.pixel_shape
+        underlying_shape = np.asarray(self._wcs.pixel_shape)
+        int_elements = np.mod(underlying_shape, self._factor) == 0
+        pixel_shape = underlying_shape / self._factor
+        return tuple(int(i) if is_int else i for i, is_int in zip(pixel_shape, int_elements))
```
dhomeier, line 75:
```suggestion
        int_elements = np.isclose(np.mod(underlying_shape, self._factor), 0, atol=np.finfo(float).resolution)
        pixel_shape = underlying_shape / self._factor
        return tuple(int(np.rint(i)) if is_int else i for i, is_int in zip(pixel_shape, int_elements))
```
Less a suggestion than a very nitpicky idea – for factors not representable as exact floats, like ⅓, the first check will fail, so e.g. `factor=[1/2, 1/3]` turns a pixel shape of 1 into `(2, 3.0)`. Very cosmetic issue probably.
It will also still create floats from certain numbers like 1/11, 1/13, which turn out to be slightly larger in numpy than the exact value.

dhomeier (changes requested):
Some mainly technical remarks; if my comments on the lengths of `factor` and `offset` are correct, those checks could perhaps use a test as well.

ndcube/wcs/wrappers/resampled_wcs.py
```diff
@@ -13,34 +13,75 @@ class ResampledLowLevelWCS(BaseWCSWrapper):
...
+    def __init__(self, wcs, factor, offset=0):
         self._wcs = wcs
         if np.isscalar(factor):
             factor = [factor] * self.pixel_n_dim
-        self._factor = factor
+        self._factor = np.array(factor)
+        if len(self._factor) !=  self.pixel_n_dim:
+            raise ValueError(f"Length of factor must equal number of dimensions {self.pixel_n_dim}.")
+        if offset is None:
+            offset = 0
```
DanRyanIrish, line 34:
```suggestion
```

ndcube/wcs/wrappers/resampled_wcs.py
```diff
@@ -13,34 +13,75 @@ class ResampledLowLevelWCS(BaseWCSWrapper):
...
+        # Return pixel shape of resampled grid.
+        # Where shape is an integer, return an int type as its required for some uses.
+        if self._wcs.pixel_shape is None:
+            return self._wcs.pixel_shape
+        underlying_shape = np.asarray(self._wcs.pixel_shape)
+        int_elements = np.isclose(np.mod(underlying_shape, self._factor), 0,
+                                  atol=np.finfo(float).resolution)
+        pixel_shape = underlying_shape / self._factor
+        return tuple(int(np.rint(i)) if is_int else i for i, is_int
+                     in zip(pixel_shape, int_elements))
```
DanRyanIrish, line 80:
```suggestion
        return tuple(int(np.rint(i)) if is_int else i
                     for i, is_int in zip(pixel_shape, int_elements))
```

DanRyanIrish:
Hi @dhomeier.  Thanks so much for your comments.  I've taken them into account and will add an extra test as you suggest.

ndcube/wcs/wrappers/resampled_wcs.py
```diff
@@ -13,34 +13,73 @@ class ResampledLowLevelWCS(BaseWCSWrapper):
...
     def pixel_shape(self):
-        return tuple(self._wcs.pixel_shape[i] / self._factor[i]
-                     for i in range(self.pixel_n_dim))
+        # Return pixel shape of resampled grid.
+        # Where shape is an integer, return an int type as its required for some uses.
+        if self._wcs.pixel_shape is None:
+            return self._wcs.pixel_shape
+        underlying_shape = np.asarray(self._wcs.pixel_shape)
+        int_elements = np.isclose(np.mod(underlying_shape, self._factor), 0,
+                                  atol=np.finfo(float).resolution)
```
Cadair, line 75: :exploding_head:

Cadair (approved):

ndcube/wcs/wrappers/resampled_wcs.py
dhomeier: Actually
```python
        int_elements = np.isclose(underlying_shape / self._factor, np.rint(underlying_shape / self._factor),
                                  atol=np.finfo(float).resolution)
```
might be the most robust way to check this.

Issue #444, opened 2021-07-26
ExtraCoords.add docstring should make clear how to set physical type.


DanRyanIrish:
### Description

This is a specific example of #413

dhomeier:
I am having a look at this as I am familiarising myself more with the docs.
I thought #413 was more specifically directed at the `__str__` and `__repr__` formatting and content, but this is concerning the docstring of `ExtraCoords.add()` directly, right?

For the actual 'howto set physical type', did you think about documenting how one can (or should) pass `physical_types` in `**kwargs` with any `lookup_table` that is a subclass of `BaseTableCoordinate`?
I think to begin with both here and in `from_lookup_table` `lookup_table` itself should be more precisely documented than just (tuple of) `object`, as effectively only `BaseTableCoordinate` and certain of its subclasses are accepted. Maybe to x-ref their docs would be a convenient start – currently I cannot find them anywhere in the API docs.

dhomeier:
As for setting them, apart from the init syntax of `BaseTableCoordinate` this would require information on _what_ can be set as a physical type, e.g. a reference to https://docs.sunpy.org/projects/ndcube/en/v2.0.0rc1/data_classes.html?highlight=physical%20type#dimensions-and-physical-types which seems to be the most comprehensive (if not the only) explanation of the role of physical types. I think this could still use additional clarification especially what distinguishes them from the `unit.physical_type` already brought in by `Quantity` objects setting the `TableCoordinate`s.

dhomeier:
Not directly related, in the `from_lookup_table` docstring I noted an `array_shape : ``tuple`` of ``int``, optional` as parameter, which is neither declared nor used anywhere in the function.

Cadair:
> I am having a look at this as I am familiarising myself more with the docs.
> I thought #413 was more specifically directed at the `__str__` and `__repr__` formatting and content, but this is concerning the docstring of `ExtraCoords.add()` directly, right?

yes, I don't really think they are that related either.

> For the actual 'howto set physical type', did you think about documenting how one can (or should) pass `physical_types` in `**kwargs` with any `lookup_table` that is a subclass of `BaseTableCoordinate`?

Yeah, I think this needs clarifying a lot, and the physical_type argument probably needs moving out of the `**kwargs` to make it clearer as well.

[...]

Issue #416, opened 2021-03-29
Support >1D tables in TableCoordinate and ExtraCoords


Cadair:
Currently (as of #414) we explicitly disable constructing lookup tables (through `ExtraCoords` or directly) because the inverse transform [in the tabular model](https://github.com/astropy/astropy/blob/master/astropy/modeling/tabular.py#L250) doesn't support converting from value (world) to pixel. The reason for this is that it is possible for that value -> index (world to pixel) transform can be multi-valued. If there is two values in the table then you can't get a single index out, so the transform is invalid.

There are a few different ways to approach fixing this, the ones that come to mind at the moment are:

* Implement the inverse transform in `_Tabular` as long as it is singular valued. This would make it work as long as you pass a well behaved table in. The catch to note here is that each coordinate would have to be singular valued (i.e. lat and lon, or x, y and z) not the union of all three. This is because the inverse is called for each coordinate individually. This limitation would be pretty damming for the most common of our use cases.
* Do the above *and* also refactor `QuantityTableCoordinate` to only use one N+1 dimension lookup table. This would probably require some special handling if the units on the table are different, and might have some other fun issues which I can't forsee.
* Make `BaseTableCoordinate` implement both the Low and High level APE 14 interfaces itself. Then we can work around the limitation of the first option by writing our own special cased multi-table world to pixel transform. We could lean on the gWCS coordinate frame machinery to help us out a lot with this, it would really only come down to implementing `world_to_pixel` and `pixel_to_world` as all the other stuff (and the values variants of those methods) we could lean of the gWCS coordinate frames for. This would give us the other advantage of being able to pass a `BaseTableCoordinate` instance right into `NDCube` without having to do irreversibly convert it to a gWCS first.

The third option is definitely a lot of work, but the first two aren't exactly easy either. I think the third one also gives us some nice secondary advantages, so would be the way I would be tempted to go.

PR #451, opened 2021-08-02
Improve documentation for `ExtraCoords` methods and `physical_types`

dhomeier:
### Description
This is addressing some of the open documentation issues, specifically #444 and other `TableCoordinate`-related interfaces discussed there.

First, this adds `ndcube.extra_coords.table_coord` to the API reference, as a proxy for `BaseTableCoordinate` and to provide links to its subclasses. Also trying to fix some other broken links in `tabular_coordinates.rst`.

`physical_types` is explicitly declared and documented as kwarg in `ExtraCoords.add` and `ExtraCoords.from_lookup_tables`; as a by-product the latter now also passes `physical_types` on.

Some of the x-refs with labels look a bit unsightly in the inline docstrings, but should render decently in the built docs.

I was trying to also label the intersphinx link for gwcs somehow like `gwcs <~gwcs>`, but that is not recognised as external reference – if anyone has an idea how to do this, or what the intersphinx reference to the toplevel gwcs docs would be, pointers welcome. But since it's only in a single place, putting in the explicit URL is probably OK.

### DIFF
--- a/ndcube/extra_coords/extra_coords.py
+++ b/ndcube/extra_coords/extra_coords.py
@@ -53,8 +53,8 @@ class ExtraCoordsABC(abc.ABC):
         array_dimension : `int` or `tuple` of `int`
             The pixel dimension(s), in the array, to which this lookup table corresponds.
         lookup_table : `object` or sequence of `object`
-            The lookup table. A `BaseTableCoordinate <.table_coord>` or anything that can
-            instantiate one of its subclasses, i.e. currently a `~astropy.time.Time`,
+            The lookup table. A `BaseTableCoordinate <.table_coord>` subclass or anything
+            that can instantiate one, i.e. currently a `~astropy.time.Time`,
             `~astropy.coordinates.SkyCoord`, or a (sequence of) `~astropy.units.Quantity`.
             Note, if this table is multi-dimensional it must (currently) be specified with
             its axes in world order, so transposed with respect to the data array.
@@ -157,13 +157,13 @@ class ExtraCoords(ExtraCoordsABC):
         pixel_dimensions : `tuple` of `int`
             The pixel dimensions (in the array) to which the ``lookup_tables``
             apply. Must be the same length as ``lookup_tables``.
-        lookup_tables : iterable of `object`, optional
+        lookup_tables : iterable of `object`
             The lookup tables which specify the world coordinates for the ``pixel_dimensions``.
-            Must be `BaseTableCoordinate <.table_coord>` or objects from which to
-            instantiate one of its subclasses.
-        physical_types: sequence of `str` or of sequences of `str`
+            Must be `BaseTableCoordinate <.table_coord>` subclass instances or objects from
+            which to instantiate them (see `.ExtraCoords.add`).
+        physical_types: sequence of `str` or of sequences of `str`, optional
             Descriptors of the `physical types <../data_classes.html#dimensions-and-physical-types>`_
-            associated with each axis in the tables. Must be the same length as ``lookup_tables``
+            associated with each axis in the tables. Must be the same length as ``lookup_tables``;
             and length of each element must match the number of dimensions in corresponding
             ``lookup_tables[i]``.
 
--- a/ndcube/extra_coords/extra_coords.py
+++ b/ndcube/extra_coords/extra_coords.py
@@ -56,8 +56,6 @@ class ExtraCoordsABC(abc.ABC):
             The lookup table. A `BaseTableCoordinate <.table_coord>` subclass or anything
             that can instantiate one, i.e. currently a `~astropy.time.Time`,
             `~astropy.coordinates.SkyCoord`, or a (sequence of) `~astropy.units.Quantity`.
-            Note, if this table is multi-dimensional it must (currently) be specified with
-            its axes in world order, so transposed with respect to the data array.
         physical_types: `str` or iterable of `str`, optional
             Descriptor(s) of the `physical type <../data_classes.html#dimensions-and-physical-types>`_
             associated with each axis; length must match the number of dimensions in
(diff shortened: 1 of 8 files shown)


dhomeier:
astropy needed a new `devdeps` version since 4.3x has been released. :)

ndcube/extra_coords/extra_coords.py
```diff
@@ -38,23 +38,30 @@ class ExtraCoordsABC(abc.ABC):
...
-        lookup_table
-            The lookup table. Note, if this table is multi-dimensional it must
-            (currently) be specified with its axes in world order, so
-            transposed with respect to the data array.
+        lookup_table : `object` or sequence of `object`
+            The lookup table. A `BaseTableCoordinate <.table_coord>` subclass or anything
+            that can instantiate one, i.e. currently a `~astropy.time.Time`,
+            `~astropy.coordinates.SkyCoord`, or a (sequence of) `~astropy.units.Quantity`.
+            Note, if this table is multi-dimensional it must (currently) be specified with
+            its axes in world order, so transposed with respect to the data array.
```
DanRyanIrish, line 60: I don't think this is true any more.  I think the axis is now given in array order.

dhomeier:
Is there a `451.documentation.rst` category for the changelog? Though technically with the added support for `physical_types` in `from_lookup_tables` this might also go under `feature`.

DanRyanIrish (commented):
Hi @dhomeler.  This looks like a great help.  I have one clarification about the ordering of `axis` in `ExtraCoords.add`.

ndcube/extra_coords/extra_coords.py
dhomeier: So just cut the last sentence, or better explicitly state that it has changed from 1.x?

DanRyanIrish:
> Is there a `451.documentation.rst` category for the changelog? Though technically with the added support for `physical_types` in `from_lookup_tables` this might also go under `feature`.

You should be able to use `451.doc.rst`.  I think a `doc` type changelog is fine here since the main thrust of the PR is making how to use `physical_types` clearer.

ndcube/extra_coords/extra_coords.py
dhomeier: In #416 it sounds as if at this point there is no support for multi-dimensional tables at all.

DanRyanIrish: Ah, I slightly misread this sentence initially which has caused a little confusion.  There is support for multi-coordinate tables.  But each individual coordinate must be 1D and can only correspond to one array axis.

On a separate point that caused my initial confusion, the ordering in which the single array axis must be specified is in array order, not pixel order.  For example, say you have a 3-D cube and want to add a 1-D extra coord to the 0th array axis.  In pixel order you would give this as axis 2 because pixel and array ordering are reversed.  However, the API for `ExtraCoords.add` requires you use array ordering to specify the single axis, i.e. 0.  I now realise you were not referring to this part of the API, but my misunderstanding prompted my response.

So to summarise, just cut the last sentence.

DanRyanIrish:
> > Is there a `451.documentation.rst` category for the changelog? Though technically with the added support for `physical_types` in `from_lookup_tables` this might also go under `feature`.
> 
> You should be able to use `451.doc.rst`. I think a `doc` type changelog is fine here since the main thrust of the PR is making how to use `physical_types` clearer.

Alternatively, you can add two changelogs, `451.doc.rst` and `451.feature.rst` and break up the descriptions of the two different types of changes.  This might actually be the best way.

Cadair:
> I was trying to also label the intersphinx link for gwcs somehow like gwcs <~gwcs>, but that is not recognised as external reference – if anyone has an idea how to do this, or what the intersphinx reference to the toplevel gwcs docs would be, pointers welcome.

I think I have run into this before, I don't think there is a high level reference. (Perhaps we should PR one?!)

Cadair:
Looks good to me :+1:

ndcube/extra_coords/extra_coords.py
dhomeier: I have tried not to touch the actual content of the existing description of lookup tables, so did not check if any of that had changed recently. But if that bit is obsolete now I'll cut it.

dhomeier:
I wasn't sure if the gwcs documentation is still under rapid development and correspondingly fluid...
We could probably address this on our side with ReST substitutions as astropy has them in `common_links.txt`; don't know if that's worth it just for gwcs and perhaps BaseTableCoordinate at this point.

dhomeier:
I should add that in the test I added I haven't spent much thought on whether this is a sensible application of `from_lookup_tables`, as I initially expected `ExtraCoords.from_lookup_tables(["time", "exposure_time"], ...)` would be strictly equivalent to the successive `ec.add('time', ...); ec.add('exposure time', ...)` in the previous test before realising that the former is a class method.
So in the new test example checking mapping will raise
```python
>>> ec.mapping
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/derek/git/ndcube/ndcube/extra_coords/extra_coords.py", line 250, in mapping
    converter = partial(convert_between_array_and_pixel_axes, naxes=len(self._ndcube.dimensions))
AttributeError: 'NoneType' object has no attribute 'dimensions'
```
Frankly I didn't fully understand (quite literally) the Magic going on when initialising `ExtraCoords(cube)` in the first test, or how to correctly reproduce this using `from_lookup_tables`. The NDCube can be associated with it by `ec._ndcube = cube`, making this work in a somewhat after the fact manner; maybe the more straightforward use would be for an existing
```python
my_cube._extra_coords = ExtraCoords.from_lookup_tables(['time', 'exp_time'], (0, 0), [input_time, exp_time], physical_types=['custom:time:creation', 'custom:time:duration'])
```
but I haven't found a way to do this on initialisation or using public methods.

For checking the physical types the test works so far, but since there is only one other test for `from_lookup_tables`, and that's testing an exception, it could perhaps use some more coverage and documentation how to use it in real code.

dhomeier:
I've put the `mapping` check back in a `pytest.raises` so it's not forgotten; otherwise I think the physical types are now addressed and the general `from_lookup_tables` usage can go into a separate issue, if necessary.

DanRyanIrish (approved):

DanRyanIrish:
Hi @dhomeier.  This looks great.  I think it's ready to merge.  Does anyone else have any final requests?  @Cadair?

PR #452, opened 2021-08-04
Crop by ExtraCoords

### LINKED ISSUE #345: Does crop actually work with just extra_coords?
Do we have tests for it? how does it handle extra_coords having less pixel axes than the array? @DanRyanIrish

DanRyanIrish:
### Description

This PR adds support to `NDCube.crop` and `NDCube.crop_by_values` for cropping using an `ExtraCoords` object.

It also fixes a bug in `CompoundLowLevelWcs,world_to_pixel_values` where it checks whether the input world values lead to compatible pixels values when pixel dimensions are shared and mapping is not trivial.

Fixes #345


ndcube/ndcube.py
```diff
@@ -593,7 +589,10 @@ def _fill_in_crop_nones(self, lower_corner, upper_corner, wcs, crop_by_values):
             else:
                 array_index_to_world = wcs.array_index_to_world_values
         else:
-            array_index_to_world = wcs.array_index_to_world
+            if isinstance(wcs, BaseHighLevelWCS):
+                array_index_to_world = wcs.array_index_to_world
+            else:
+                array_index_to_world = HighLevelWCSWrapper(wcs).array_index_to_world
```
Cadair, line 595: This check feels redundant with the one above.

DanRyanIrish:
@Cadair: This is ready for review.

Cadair (changes requested):
I feel like this could use some more test cases, i.e. an EC for all array axes, more than one EC per array axis etc.

ndcube/utils/misc.py
```diff
@@ -57,6 +61,46 @@ def wcs_wrapper(*args, **kwargs):
...
+        dummy_axes = cube_axes - ec_axes
+        if dummy_axes:
+            n_dummy_axes = len(dummy_axes)
+            dummy_wcs = WCS(naxis=n_dummy_axes)
+            dummy_wcs.wcs.crpix = [1] * n_dummy_axes
+            dummy_wcs.wcs.cdelt = [1] * n_dummy_axes
+            dummy_wcs.wcs.crval = [0] * n_dummy_axes
+            dummy_wcs.wcs.ctype = ["PIXEL"] * n_dummy_axes
+            mapping = list(wcs.mapping) + list(dummy_axes)
+            wcs = CompoundLowLevelWCS(wcs.wcs, dummy_wcs, mapping=mapping)
```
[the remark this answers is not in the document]
DanRyanIrish: I agree.  I'll put into into a new property named something like `cube_wcs`.

DanRyanIrish:
@Cadair.  The changes you requested have been made and this PR is ready for another review.

ndcube/ndcube.py
```diff
@@ -526,8 +524,14 @@ def crop(self, lower_corner, upper_corner, wcs=None):
...
-        lower_corner_values = [v << u.Unit(unit) for v, unit in zip(lower_corner_values, wcs.world_axis_units)]
-        upper_corner_values = [v << u.Unit(unit) for v, unit in zip(upper_corner_values, wcs.world_axis_units)]
+        # lower_corner_values = [v << u.Unit(unit)
+        #                       for v, unit in zip(lower_corner_values, wcs.world_axis_units)]
+        # upper_corner_values = [v << u.Unit(unit)
+        #                       for v, unit in zip(upper_corner_values, wcs.world_axis_units)]
+        lower_corner_values = [u.Quantity(v, unit=u.Unit(unit))
+                               for v, unit in zip(lower_corner_values, wcs.world_axis_units)]
+        upper_corner_values = [u.Quantity(v, unit=u.Unit(unit))
+                               for v, unit in zip(upper_corner_values, wcs.world_axis_units)]
```
DanRyanIrish, line 534: Insert `copy=False` to Quantity constructor.

Delete commented out code.

ndcube/ndcube.py
```diff
@@ -615,7 +620,33 @@ def _bounding_box_to_points(self, lower_corner_values, upper_corner_values, wcs)
         """
         Convert two corners of a bounding box to the points of all corners.
         """
-        return tuple(itertools.product(*zip(lower_corner_values, upper_corner_values)))
+        corners = np.array(tuple(itertools.product(*zip(lower_corner_values, upper_corner_values))),
+                           dtype=object)
+        if hasattr(wcs, "mapping"):
```
DanRyanIrish, line 625: This code is to catch circumstances where world axes share pixel axes.  The first line of this method would create 2^n corners where `n` is the number of world axes.  However, only 2 of these corners are valid as they all share the same axis.  This will cause `world_to_pixel` to error.

This solution however, is specific to a WCS implementation.  It should be refactored to detect the situation based on the `axis_correlation_matrix`.  Exactly how to do this requires more thought.

DanRyanIrish:
@Cadair I've come up with a more general implementation of handle dependent axes when calculating the bounding box.  This PR is ready for another review.

ndcube/ndcube.py
```diff
@@ -607,11 +609,105 @@ def _fill_in_crop_nones(self, lower_corner, upper_corner, wcs, crop_by_values):
...
+                                axis_correlation_matrix):
         """
         Convert two corners of a bounding box to the points of all corners.
         """
-        return tuple(itertools.product(*zip(lower_corner_values, upper_corner_values)))
+        # Calculate which world axes share multiple pixel axes.
+        world_n_dim, pixel_n_dim = axis_correlation_matrix.shape
+        world_axes = np.arange(world_n_dim)
+        world_n_dep_dim = axis_correlation_matrix.sum(axis=1)
+        # If all world coordinates are 1D, bounding box is simple.
```
Cadair, line 621: The bounding box is also simple if n_world == n_pixel

DanRyanIrish: Good point.  I'll add a check for this and call the original implementation if `True`

ndcube/ndcube.py
```diff
@@ -607,11 +610,108 @@ def _fill_in_crop_nones(self, lower_corner, upper_corner, wcs, crop_by_values):
...
+                                "must be of same type.")
+            if isinstance(lcv, u.Quantity):
+                corner_units[i] = lcv.unit
+                lower_corner_values[i] = lower_corner_values[i].value
+                upper_corner_values[i] = ucv.to_value(corner_units[i])
+        lower_corner_values = np.asarray(lower_corner_values)
+        upper_corner_values = np.asarray(upper_corner_values)
+        corners = np.stack([lower_corner_values, upper_corner_values])
+
+        # Next, calculate the sets of pixel axes upon which each world axis depends.
```
DanRyanIrish, line 647:
```suggestion

        # Otherwise we need calculate the corners more carefully.
        # This must be done based on correlation matrix as not all
        # world axes are independent.
        # Start by generating array of world corner values assuming
        # all coordinates are 1D. Strip units and add them back on later.
        corner_units = [None] * len(lower_corner_values)
        for i, (lcv, ucv) in enumerate(zip(lower_corner_values, upper_corner_values)):
            if not isinstance(ucv, type(lcv)):
                raise TypeError("Corresponding entries in lower and upper corner values "
                                "must be of same type.")
            if isinstance(lcv, u.Quantity):
                corner_units[i] = lcv.unit
                lower_corner_values[i] = lower_corner_values[i].value
                upper_corner_values[i] = ucv.to_value(corner_units[i])
        lower_corner_values = np.asarray(lower_corner_values)
        upper_corner_values = np.asarray(upper_corner_values)
        corners = np.stack([lower_corner_values, upper_corner_values])

        # Next, calculate the sets of pixel axes upon which each world axis depends.
        world_n_dim, pixel_n_dim = axis_correlation_matrix.shape
```

ndcube/ndcube.py
```diff
@@ -607,11 +610,108 @@ def _fill_in_crop_nones(self, lower_corner, upper_corner, wcs, crop_by_values):
 
         return lower_corner, upper_corner
 
-    def _bounding_box_to_points(self, lower_corner_values, upper_corner_values, wcs):
+    def _bounding_box_to_points(self, lower_corner_values, upper_corner_values,
+                                axis_correlation_matrix):
```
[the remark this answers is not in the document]
DanRyanIrish: Good question.  The `axis_correlation_matrix` is the only thing from the `wcs` that's needed.  So in the interests of only requiring what's needed, and the fact this is a hidden method, I think passing the matrix is better.  It also makes it easier to test and deploy in other circumstances as a matrix is easier to build than a WCS.

ndcube/ndcube.py
```diff
@@ -607,11 +610,108 @@ def _fill_in_crop_nones(self, lower_corner, upper_corner, wcs, crop_by_values):
...
+        dep_pix_axes = np.array([set(np.arange(pixel_n_dim)[axis_correlation_matrix[j]])
+                                 for j in world_axes], dtype=object)
+
+        # Iterate through number of shared pixel axes, i, and create additional corners.
+        # Iterate from highest to lowest number of shared axes to reduce duplication.
+        # Do not include world axes sharing only 1 pixel axis, as we have already
+        # captured those in our initial definition of corners, above.
+        for i in range(max_dep_dim, 1, -1):
+            # Extract the world axes whose number of shared pixel axes is i.
+            world_axes_idx = np.where(world_n_dep_dim == i)[0]
```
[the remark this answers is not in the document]
DanRyanIrish: This case is still handled.  The above line of code (and the one below it) gets the world axis (axes) which correspond to `i` number of pixel axes.  At this point, the number of pixel axes is checked, but the specific set of pixel axes can be different or the same.

[The next line](https://github.com/sunpy/ndcube/pull/452/files#diff-929794bb532be1486907b360285433b432f8505ab7e7ff63eca3eab657f7d00dR657) then retrieves the set (sets) of pixel axes corresponding to each above world axis and stores them in `dep_pix_axes_i`.  At this stage, if there are two world axes that both share `i` pixel axes, but not the same set of pixel axes, the entries in `dep_pix_axes_i` will be different.  If both world axes do correspond to the same set of pixel axes, the entries in `dep_pix_axes_i` will be the same.

[The while loop](https://github.com/sunpy/ndcube/pull/452/files#diff-929794bb532be1486907b360285433b432f8505ab7e7ff63eca3eab657f7d00dR660) loops through each world axis and corresponding set of dependent pixel axes and builds the required corners.  But that loop also checks the sets of remaining dependent pixel axes in `dep_pix_axes_i`.  If any are the same as the current set, remaining sets are deleted from `dep_pix_axes_i` along with the corresponding world axis in `world_axes_i`.

This means corners are not duplicated by subsequent iterations of the `while` loop.  But it also means that if the set of dependent pixel axes of another world axis has the same length, but ISN'T the same set, it will not be deleted.  Hence its corners will still be calculated by the `while` loop.

ndcube/utils/cube.py
```diff
@@ -0,0 +1,351 @@
...
+            raise TypeError("wcs argument must be a High Level WCS or an ExtraCoords object.")
+
+        params.arguments['wcs'] = wcs
+
+        return func(*params.args, **params.kwargs)
+
+    return wcs_wrapper
+
+
+def get_crop_item(lower_corner, upper_corner, wcs, data_shape):
```
DanRyanIrish, line 54: Move this back onto `NDCube` as a method.

ndcube/utils/cube.py
```diff
@@ -0,0 +1,351 @@
...
+                           for v, unit in zip(lower_corner_values, wcs.world_axis_units)]
+    upper_corner_values = [u.Quantity(v, unit=u.Unit(unit), copy=False)
+                           for v, unit in zip(upper_corner_values, wcs.world_axis_units)]
+
+    points = bounding_box_to_corners(lower_corner_values, upper_corner_values,
+                                     wcs.axis_correlation_matrix)
+    return get_crop_item_from_points(*points, wcs=wcs, data_shape=data_shape)
+
+
+def get_crop_by_values_item(lower_corner, upper_corner, wcs, data_shape, units=None):
```
DanRyanIrish, line 79: Move this back onto `NDCube` as a method.

ndcube/utils/cube.py
```diff
@@ -0,0 +1,351 @@
...
+            corners = corners.reshape((np.prod(corners.shape[:3]), corners.shape[-1]))
+
+    # Reattach units if were present in input.
+    corners = tuple(tuple(corner * corner_unit if corner_unit else corner
+                          for corner, corner_unit in zip(corners[i], corner_units))
+                    for i in range(corners.shape[0]))
+    return corners
+
+
+def get_crop_item_from_points(*world_points_values, wcs, data_shape):
```
DanRyanIrish, line 305: Consider moving this back onto `NDCube` once `get_crop_item` and `get_crop_item_values` are moved back.

PR #455, opened 2021-08-14
Sliceable MetaData Class

DanRyanIrish:
### Description

This PR introduces a sliceable `Meta` class based on `dict`, which allows metadata to be associated with data axes and the whole class to be sliced using the normal Python slicing API.  Metadata not associated with any axes is not touched by the slicing mechanism.

This PR introduces two types of axis-aware metadata: axis-aligned and grid-aligned.  Axis-aligned metadata gives one value for each of its associated axes, and is scalar or a string if only associated with one axis.  This type of metadata is only changed if the axis/axes within which it is associated is dropped by slicing.  In this case, the values corresponding to the dropped axes are also dropped.

Grid-aligned metadata provides a value for each data array element in its associated axis/axes.  This metadata cannot be scalar.  Since this metadata mirrors at least part of the data array, it is sliced as if it were data, and its shape is kept consistent with its associated data axes.  If all its axes are sliced away, the scalar value at the row/column location at which it was sliced is kept, and its axis-awareness is dropped, i.e. it becomes a piece of normal non-axis-aware scalar metadata.

### API Summary
```python
>>> import astropy.units as u
>>> from ndcube.meta import Meta

# Define a Meta object associated with data of shape (3, 4, 5)
>>> header = {"salutation": "hello",
              "name": "world",
              "exposure_time": ([2] * 3) * u.s,  # 1-dimension grid-aligned (pixel-wise) metadata
              "instrument_axes": ["slit step", "slit", "dispersion"],  # axis-aligned (scalar/string-per-axis) metadata
              "pixel_efficiency": np.random.ones(4, 5)  # multi-dimension grid-aligned (pixel-wise) metadata}
>>> comments = {"salutation": "This is polite."}
>>> axes = {"exposure_time": 0,
              "instrument_axes": (0, 1, 2),
              "pixel_efficiency": (1, 2)}
>>> meta = Meta(header, axes=axes, comments=comments, data_shape=(3, 4, 5))

# Use slice item that you would apply to the cube on the Meta object.
>>> sliced_meta = meta[0:2, 2]  # Reduce length of 0th axis and drop 1st axis.
>>> sliced_meta.shape  # Shape has been updated.
(2, 5)
>>> sliced_meta["exposure_time"]  # Exposure time now length of new 0th axis
<Quantity [2., 2.] s >
>>> sliced_meta["instrument_axes"]  # instrument axes has dropped slit entry
["slit step", "dispersion"]
>>> sliced_meta.axes["instrument_axes"]
(0, 1)
>>> sliced_meta["pixel_efficiency"]  # pixel efficiency is now 1-D of length 5
array([1., 1., 1., 1., 1.])
>>> sliced_meta.axes["pixel_efficiency"]
(1,)

# Rebin the Meta object
>>> rebinned_meta = meta.rebin((1, 2, 1))
>>> rebinned_meta["exposure_time"]  # grid-aligned metadata associated with non-rebinned axes remains unchanged...
<Quantity [2., 2., 2.] s >
>>> rebinned_meta.axes["exposure_time"]  # ...and still associated with its axis/axes
(0,)
>>> rebinned_meta["pixel_efficiency"]  # grid-aligned metadata associated with rebinned axes remains...
np.random.ones(4, 5)
>>> rebinned_meta.axes["pixel_efficiency"]  # ...but grid-awareness is removed
None
>>> rebinned_meta["instrument_axes"]  # axis-aligned (scalar/string-per-axis) metadata unchanged...
["slit step", "slit", "dispersion"]
>>> rebinned_meta.axes["instrument_axes"]  # ...and still associated with axes as they haven't been dropped
(0, 1, 2)
```
### To Do
- [x] Implement `__delitem__` and remove `remove` method.
- [x] Implement setter for `NDMeta.data_shape` which verifies that it's compatible with pre-existing axis- and grid-aligned metadata.
- [x] Improve narrative docs


DanRyanIrish:
@nabobalis Could you give this PR a review?  It should be very similar if not the same as what's in your sunraster PR.

DanRyanIrish:
@Cadair this PR is ready for review.  To pre-empt possible concerns regarding DKIST tools, the introduction of this new class does not require that metadata be sliceable.  Instead it allows for it if desired.

In order for the `Meta` object to be sliceable, `Meta.shape` must be set via the `data_shape` kwarg during intialization.  This must be set to the shape of the data array with which the metadata object is associated.  If `Meta.shape` is `None`, the object is not sliceable.

Not all entries in the `Meta` object have to be sliceable.  Only those who have a corresponding entry in the `Meta.axes` `dict` will be.  This gives the axes with which the metadata is associated and is therefore used to work out how to slice it.  If `Meta.shape` is set, but there are no entries in `Meta.axes`, then slicing has no effect on the `Meta` object.

Therefore, as far as I understand, this new class should not threaten your assumption in the DKIST tools that `.meta` never changes.

ndcube/meta.py
```diff
@@ -0,0 +1,248 @@
...
+    data_shape: iterable of `int` (optional)
+        The shape of the data with which this metadata is associated.
+        Must be set if axes input is set.
+    """
+    def __init__(self, header=None, comments=None, axes=None, data_shape=None):
+        self.original_header = header
+
+        # Sanitize metadata values and instantiate class.
+        if header is None:
+            header = {}
```
nabobalis, line 43: Code cov is saying quite a few places aren't covered. Unsure if you want to address this in this PR or in a future one.

ndcube/meta.py
```diff
@@ -0,0 +1,248 @@
...
+
+    def add(self, name, value, comment, axis, overwrite=False):
+        """Add a new piece of metadata to instance.
+
+        Parameters
+        ----------
+        name: `str`
+            The name/label of the metadata.
+
+        value:
```
nabobalis, line 133: Can it be any type for the value?

nabobalis (approved):

ndcube/meta.py
DanRyanIrish: Yes.

ndcube/meta.py
DanRyanIrish: Almost all those places are errors being raised and over 90% of the diff is being hit.  I think that's fine for this PR.

nabobalis: Works for me.

DanRyanIrish:
@Cadair, this is ready to merge as far as I can see.  Do you still want more time to look at it given your DKIST concerns?  As I above, I don't think this causes an issues as the new functionalities are optional and don't have to change previous behaviour.

DanRyanIrish:
@Cadair, went looking for this functionality for another package looking to use it and realised we never actually merged it and the stale bot took it.  Any objections to merging it now, pending a couple minor changes to make the tests pass?  As the previous comments pointed out, the use of this new Metadata class is optional and so doesn't disrupt any current users' pipelines.

DanRyanIrish:
Anyone know what's causing [this warning](https://github.com/sunpy/ndcube/actions/runs/8986257004/job/24682053039?pr=455#step:10:461) in the doc build?  The traceback is not helpful.

ndcube/__init__.py
```diff
@@ -7,11 +7,29 @@
 * Homepage: https://github.com/sunpy/ndcube
 * Documentation: https://docs.sunpy.org/projects/ndcube/
 """
+import sys
```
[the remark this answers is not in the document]
nabobalis: Sure, two separate problems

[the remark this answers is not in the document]
DanRyanIrish: Ah right.  So what's the issue behind the docs warnings?

ndcube/meta.py
```diff
@@ -0,0 +1,249 @@
...
+__all__ = ["Meta"]
+
+
+class Meta(dict):
+    """
+    A sliceable object for storing metadata.
+
+    Metadata can be linked to a data axis which causes it to be sliced when the
+    standard Python numeric slicing API is applied to the object.
+    Specific pieces of metadata can be obtain using the dict-like str slicing API.
```
nabobalis, line 16:
```suggestion
    Specific pieces of metadata can be obtain using the dict-like string slicing API.
```

ndcube/meta.py
```diff
@@ -0,0 +1,249 @@
...
+    Metadata associated with an axis/axes must have the same length/shape as those axes.
+
+    Parameters
+    ----------
+    header: dict-like
+        The names and values of metadata.
+
```
nabobalis, line 23: Does ndcube docs use new lines between parameters?

ndcube/meta.py
```diff
@@ -0,0 +1,249 @@
...
+    comments: dict-like (optional)
```
nabobalis, line 24:
```suggestion
    comments: dict-like, optional
```

ndcube/meta.py
```diff
@@ -0,0 +1,249 @@
...
+        Comments associated with any of the above pieces of metadata.
+
+    axes: dict-like (optional)
```
nabobalis, line 27:
```suggestion
    axes: dict-like, optional
```

ndcube/meta.py
```diff
@@ -0,0 +1,249 @@
...
+        The axis/axes associated with the above metadata values.
+        Each axis value must be None (for no axis association), and `int`
+        or an iterable of `int` if the metadata is associated with multiple axes.
+        Metadata in header without a corresponding entry here are assumed to not
+        be associated with an axis.
+
+    data_shape: iterable of `int` (optional)
```
nabobalis, line 34:
```suggestion
    data_shape: `iterable` of `int`, optional
```
Not sure if this links.

ndcube/meta.py
```diff
@@ -0,0 +1,249 @@
...
+        if comments is None:
+            self._comments = dict()
+        else:
+            comments = dict(comments)
+            if not set(comments.keys()).issubset(set(header_keys)):
+                raise ValueError(
+                    "All comments must correspond to a value in header under the same key.")
+            self._comments = comments
+
+        # Define data shape.
```
nabobalis, line 60:
```suggestion
```

ndcube/meta.py
```diff
@@ -0,0 +1,249 @@
...
+
+        # Sanitize metadata values and instantiate class.
+        if header is None:
+            header = {}
+        else:
+            header = dict(header)
+        super().__init__(header.items())
+        header_keys = header.keys()
+
+        # Generate dictionary for comments.
```
nabobalis, line 50:
```suggestion
```

ndcube/meta.py
```diff
@@ -0,0 +1,249 @@
...
+
+    data_shape: iterable of `int` (optional)
+        The shape of the data with which this metadata is associated.
+        Must be set if axes input is set.
+    """
+    def __init__(self, header=None, comments=None, axes=None, data_shape=None):
+        self.__ndcube_can_slice__ = True
+        self.original_header = header
+
+        # Sanitize metadata values and instantiate class.
```
nabobalis, line 42:
```suggestion
```

ndcube/meta.py
```diff
@@ -0,0 +1,249 @@
...
+                    "All comments must correspond to a value in header under the same key.")
+            self._comments = comments
+
+        # Define data shape.
+        if data_shape is None:
+            self._data_shape = data_shape
+        else:
+            self._data_shape = np.asarray(data_shape, dtype=int)
+
+        # Generate dictionary for axes.
```
nabobalis, line 66: ```suggestion```

ndcube/meta.py
```diff
@@ -0,0 +1,249 @@
...
+            self._axes = dict([(key, self._sanitize_axis_value(axis, header[key], key))
+                               for key, axis in axes.items()])
+
+    def _sanitize_axis_value(self, axis, value, key):
+        if axis is None:
+            return None
+        if self.shape is None:
+            raise TypeError("Meta instance does not have a shape so new metadata "
+                            "cannot be assigned to an axis.")
+        # Verify each entry in axes is an iterable of ints.
```
nabobalis, line 88:
```suggestion
```

ndcube/meta.py
```diff
@@ -0,0 +1,249 @@
...
+        if isinstance(axis, numbers.Integral):
+            axis = (axis,)
+        if not (isinstance(axis, collections.abc.Iterable) and all([isinstance(i, numbers.Integral)
+                                                                    for i in axis])):
+            raise TypeError("Values in axes must be an int or tuple of ints giving "
```
nabobalis, line 93:
```suggestion
            raise TypeError("Values in axes must be an integer or iterable of integer giving "
```

ndcube/meta.py
```diff
@@ -0,0 +1,249 @@
...
+
+    def add(self, name, value, comment, axis, overwrite=False):
+        """Add a new piece of metadata to instance.
+
+        Parameters
+        ----------
+        name: `str`
+            The name/label of the metadata.
+
+        value:
```
nabobalis, line 134:
```suggestion
        value: Any
```

ndcube/meta.py
```diff
@@ -0,0 +1,249 @@
...
+            The value of the metadata.  If axes input is not None, this must have the
+            same length/shape as those axes as defined by self.shape.
```
nabobalis, line 136:
```suggestion
            The value of the metadata. If axes input is not None, this must have the
            same length/shape as those axes as defined by ``self.shape``.
```

ndcube/meta.py
```diff
@@ -0,0 +1,249 @@
...
+
+        comment: `str` or `None`
+            Any comment associated with this metadata. Set to None if no comment desired.
+
+        axis: `int`, iterable of `int`, or `None`
+            The axis/axes with which the metadata is linked.  If not associated with any
```
nabobalis, line 142:
```suggestion
            The axis/axes with which the metadata is linked. If not associated with any
```

ndcube/meta.py
```diff
@@ -0,0 +1,249 @@
...
+            axis, set this to None.
+
+        overwrite: `bool` (optional)
```
nabobalis, line 145:
```suggestion
        overwrite: `bool`, optional
```

ndcube/meta.py
```diff
@@ -0,0 +1,249 @@
...
+            raise KeyError(f"'{name}' already exists. "
+                           "To update an existing metadata entry set overwrite=True.")
+        if comment is not None:
+            self._comments[name] = comment
+        if axis is not None:
+            axis = self._sanitize_axis_value(axis, value, name)
+            self._axes[name] = axis
+        elif name in self._axes:
+            del self._axes[name]
+        self.__setitem__(name, value)  # This must be done after updating self._axes otherwise it may error.
```
nabobalis, line 158:
```suggestion
        # This must be done after updating self._axes otherwise it may error.
        self.__setitem__(name, value)
```

ndcube/meta.py
```diff
@@ -0,0 +1,249 @@
...
+        if isinstance(item, str):
+            return super().__getitem__(item)
+
+        # Else, the item is assumed to be a typical slicing item.
+        elif self.shape is None:
+            raise TypeError("Meta object does not have a shape and so cannot be sliced.")
+
+        else:
+            new_meta = copy.deepcopy(self)
+            # Convert item to array of ints and slices for consistent behaviour.
```
nabobalis, line 199:
```suggestion
```

ndcube/tests/helpers.py
```diff
@@ -90,10 +91,38 @@ def assert_extra_coords_equal(test_input, extra_coords):
...
+            "input and expected are of different type. "
+            f"input: {type(test_input)}; expected: {type(expected_output)}")
+    multi_element_msg = "more than one element is ambiguous"
+    if isinstance(test_input, Meta) and isinstance(expected_output, Meta):
+        # Check keys are the same.
         assert test_input.keys() == expected_output.keys()
-        for key in list(test_input.keys()):
-            assert test_input[key] == expected_output[key]
+
+        # Check shapes are the same.
```
nabobalis, line 103:
```suggestion
```

ndcube/tests/helpers.py
```diff
@@ -90,10 +91,38 @@ def assert_extra_coords_equal(test_input, extra_coords):
...
+        if test_input.shape is None or expected_output.shape is None:
+            assert test_input.shape == expected_output.shape
+        else:
+            assert np.allclose(test_input.shape, expected_output.shape)
+
+        # Check values and axes are the same.
```
nabobalis, line 109:
```suggestion
```

ndcube/tests/helpers.py
```diff
@@ -90,10 +91,38 @@ def assert_extra_coords_equal(test_input, extra_coords):
...
+        for test_value, expected_value in zip(test_input.values(), expected_output.values()):
+            try:
+                assert test_value == expected_value
+            except ValueError as err:
+                if multi_element_msg in err.args[0]:
+                    assert np.allclose(test_value, expected_value)
+        # Check axes are the same.
```
nabobalis, line 116:
```suggestion
```

nabobalis:
So my current thinking is that the doc build is failing due to this:

1. Meta subclass dict, inheriting its methods. 
2. The methods pop and update (which we don't override here) have code links in them that break our doc build (due to nitpicky). 
3. We have to just ignore these in this file: https://github.com/sunpy/ndcube/blob/main/docs/nitpick-exceptions

ndcube/meta.py
```diff
@@ -0,0 +1,250 @@
...
+        Comments associated with any of the above pieces of metadata.
+
+    axes: dict-like, optional
+        The axis/axes associated with the above metadata values.
+        Each axis value must be None (for no axis association), and `int`
+        or an iterable of `int` if the metadata is associated with multiple axes.
+        Metadata in header without a corresponding entry here are assumed to not
+        be associated with an axis.
+
+    data_shape: `iterable` of `int`, optional
```
nabobalis, line 34:
```suggestion
    data_shape: `Iterator` of `int`, optional
```

docs/nitpick-exceptions
```diff
@@ -16,3 +16,5 @@ py:class a set-like object providing a view on D's keys
 py:class an object providing a view on D's values
 py:class None.  Remove all items from D.
 py:class a shallow copy of D
+py:class v, remove specified key and return the corresponding value.
+py:class None.  Update D from dict/iterable E and F.
```
samaloney, line 20: maybe?
```suggestion
py:class None.  Update D from dict/iterable E and F.
py:obj reference target not found: iterator
```

nabobalis: iterator is one I introduced. We should patch the doc string to not use it.

ndcube/meta.py
```diff
@@ -0,0 +1,250 @@
...
+        Comments associated with any of the above pieces of metadata.
+
+    axes: dict-like, optional
+        The axis/axes associated with the above metadata values.
+        Each axis value must be None (for no axis association), and `int`
+        or an iterable of `int` if the metadata is associated with multiple axes.
+        Metadata in header without a corresponding entry here are assumed to not
+        be associated with an axis.
+
+    data_shape: `iterator` of `int`, optional
```
nabobalis, line 34:
```suggestion
    data_shape: iterator of `int`, optional
```

ndcube/meta.py
```diff
@@ -0,0 +1,244 @@
...
+                raise ValueError(
+                    "All comments must correspond to a value in header under the same key.")
+            self._comments = comments
+
+        if data_shape is None:
+            self._data_shape = data_shape
+        else:
+            self._data_shape = np.asarray(data_shape, dtype=int)
+
+        # Generate dictionary for axes.
```
nabobalis, line 63:
```suggestion
```

ndcube/meta.py
```diff
@@ -0,0 +1,244 @@
...
+        if axes is None:
+            self._axes = dict()
+        else:
+            # Verify data_shape is set if axes is set.
```
nabobalis, line 67:
```suggestion
```

ndcube/meta.py
```diff
@@ -0,0 +1,244 @@
...
+        if isinstance(axis, numbers.Integral):
+            axis = (axis,)
+        if not (isinstance(axis, collections.abc.Iterable) and all([isinstance(i, numbers.Integral)
+                                                                    for i in axis])):
+            raise TypeError("Values in axes must be an integer or iterable of integers giving "
+                            "the data axis/axes associated with the metadata.")
+        axis = np.asarray(axis)
+
+        # Confirm each axis-associated piece of metadata has the same shape
+        # as its associated axes.
```
nabobalis, line 94:
```suggestion
```

ndcube/meta.py
```diff
@@ -0,0 +1,244 @@
...
+                                    f"{recommendation}")
+        super().__setitem__(key, val)
+
+    def __getitem__(self, item):
+        # There are two ways to slice:
+        # by key, or
+        # by typical python numeric slicing API,
+        # i.e. slice the each piece of metadata associated with an axes.
+
+        # If item is single string, slicing is simple.
```
nabobalis, line 186:
```suggestion
```

ndcube/meta.py
```diff
@@ -0,0 +1,244 @@
...
+        if isinstance(item, str):
+            return super().__getitem__(item)
+
+        # Else, the item is assumed to be a typical slicing item.
```
nabobalis, line 190:
```suggestion
```

ndcube/tests/test_meta.py
```diff
@@ -0,0 +1,162 @@
...
+import numpy as np
+import pytest
+
+from ndcube.meta import Meta
+from .helpers import assert_metas_equal
+
+# Fixtures
+
+
+@pytest.fixture
```
nabobalis, line 12:
```suggestion


@pytest.fixture
```

ndcube/tests/test_meta.py
```diff
@@ -0,0 +1,162 @@
...
+    values["b"] = values["b"][0]
+    comments = meta.comments
+    axes = dict([(key, axis) for key, axis in meta.axes.items()])
+    del axes["b"]
+    axes["c"] -= 1
+    axes["d"] -= 1
+    shape = meta.shape[1:]
+    print(values, comments, axes, shape)
+    expected = Meta(values, comments, axes, shape)
+    # Compare output and expected.
```
nabobalis, line 82:
```suggestion
```

ndcube/tests/test_meta.py
```diff
@@ -0,0 +1,162 @@
...
+    # Build expected result.
+    values = dict([(key, value) for key, value in meta.items()])
+    values["b"] = values["b"][0]
+    comments = meta.comments
+    axes = dict([(key, axis) for key, axis in meta.axes.items()])
+    del axes["b"]
+    axes["c"] -= 1
+    axes["d"] -= 1
+    shape = meta.shape[1:]
+    print(values, comments, axes, shape)
```
nabobalis, line 80:
```suggestion
```

ndcube/tests/test_meta.py
```diff
@@ -0,0 +1,162 @@
...
+    meta = basic_meta
+    output = meta[:, :, :, 0]
+    expected = copy.deepcopy(meta)
+    expected._data_shape = meta._data_shape[:-1]
+    assert_metas_equal(output, expected)
+
+
+def test_slice_away_independent_axis(basic_meta):
+    meta = basic_meta
+    # Get output
```
nabobalis, line 68:
```suggestion
```

ndcube/tests/test_meta.py
```diff
@@ -0,0 +1,162 @@
...
+    item = 0
+    output = meta[item]
+    # Build expected result.
```
nabobalis, line 71:
```suggestion
```

ndcube/tests/test_meta.py
```diff
@@ -0,0 +1,162 @@
...
+    shape = meta.shape[1:]
+    print(values, comments, axes, shape)
+    expected = Meta(values, comments, axes, shape)
+    # Compare output and expected.
+    assert_metas_equal(output, expected)
+
+
+def test_slice_dependent_axes(basic_meta):
+    meta = basic_meta
+    print(meta["a"])
```
nabobalis, line 88:
```suggestion
```

ndcube/tests/test_meta.py
```diff
@@ -0,0 +1,162 @@
...
+    # Get output
```
nabobalis, line 89:
```suggestion
```

ndcube/tests/test_meta.py
```diff
@@ -0,0 +1,162 @@
...
+    output = meta[:, 1:3, 1]
+    print(meta["a"])
```
nabobalis, line 91:
```suggestion
```

ndcube/tests/test_meta.py
```diff
@@ -0,0 +1,162 @@
...
+    # Build expected result.
```
nabobalis, line 92:
```suggestion
```

ndcube/tests/test_meta.py
```diff
@@ -0,0 +1,162 @@
...
+    values = dict([(key, value) for key, value in meta.items()])
+    values["c"] = values["c"][1:3, 1]
+    values["d"] = values["d"][1]
+    comments = meta.comments
+    axes = dict([(key, axis) for key, axis in meta.axes.items()])
+    axes["c"] = 1
+    del axes["d"]
+    shape = np.array([2, 2, 5])
+    expected = Meta(values, comments, axes, shape)
+    # Compare output and expected.
```
nabobalis, line 102:
```suggestion
```

ndcube/meta.py
```diff
@@ -0,0 +1,238 @@
...
+        Metadata in header without a corresponding entry here are assumed to not
+        be associated with an axis.
+
+    data_shape: iterator of `int`, optional
+        The shape of the data with which this metadata is associated.
+        Must be set if axes input is set.
+    """
+    def __init__(self, header=None, comments=None, axes=None, data_shape=None):
+        self.__ndcube_can_slice__ = True
+        self.original_header = header
```
DanRyanIrish, line 40: I used `header` as the nomenclature with `sunpy.map.Map` in mind.  But this tends to imply a FITS header, which is over-specific.  Perhaps this should be changed to `meta` and `original_meta`.  Thoughts @ayshih @Cadair?

DanRyanIrish:
This PR should probably also make `NDCube.rebin` check if the `.meta` is an instance of this metadata class and if so, just drop the axis-awareness, without removing the metadata itself.

DanRyanIrish:
> This PR should probably also make `NDCube.rebin` check if the `.meta` is an instance of this metadata class and if so, just drop the axis-awareness, without removing the metadata itself.

Alternatively, how to handle rebinning could to moved onto the `Meta` class itself via `Meta.rebin`, and then `NDCube.rebin` could check if the `.meta` attribute has a `.__rebinnable_by_ndcube__` attribute, which, if `True`, calls `Meta.rebin`.  This is the same strategy used for slicing the `Meta` class.  This would enable subclasses of `Meta` to implement rebinning differently, say, if it applied to well constrained metadata for which rebinning has a clear meaning.

ndcube/meta.py
nabobalis: Just curious.

DanRyanIrish:
Documentation of what axis-aware metadata this can handle needs updating.

DanRyanIrish:
@nabobalis The formalisation of "axis-aligned" metadata introduced in this PR, may remove much of the need for `sunraster`'s data classes.  `instrument_axes` could simply become a piece of axis-aligned metadata on the `Meta` object.

nabobalis:
> @nabobalis The formalisation of "axis-aligned" metadata introduced in this PR, may remove much of the need for `sunraster`'s data classes. `instrument_axes` could simply become a piece of axis-aligned metadata on the `Meta` object.

That sounds great. 

After we merge this in and we get a release out, should we meet to discuss what to do with sunraster?

DanRyanIrish:
> After we merge this in and we get a release out, should we meet to discuss what to do with sunraster?

I'd be happy to have that discussion.  We could inventory what `sunraster` now provides in light of this PR other recent-ish functionalities.  My sense is the following:
* spice module can be moved to `sunkit-instruments`
* `exposure_time_correction` can be removed, as users can now directly divide the cube by the exposure time correction thanks to arithmetic operations.
* `instrument_axes` can be put onto the `.meta` as an axis-aligned piece of metadata.

This leaves the
* solar slit spectrograph metadata API
  * I think this really belongs in sunpy core, but should be part of a wider metadata API policy.
* The `RasterSequence`, whose main power is its interchangeable 4-D/3-D with an instrumentally meaningful API
  * I don't foresee this living anywhere other than on a slit-spectrograph-specific subclass of `NDCube`.
* Coordinate properties, e.g. `spectral_axis`.
  * Again, I don't see this living anywhere other than on a slit-spectrograph-specific subclass of `NDCube`.  Unless something clever can be done with a `Coords` class that has been mentioned before.  Even so, I don't see this in the foreseeable future.
  * Alternatively, the same result can be achieved slightly less conveniently through `axis_world_coords`.

nabobalis:
> > After we merge this in and we get a release out, should we meet to discuss what to do with sunraster?
> 
> I'd be happy to have that discussion. We could inventory what `sunraster` now provides in light of this PR other recent-ish functionalities. My sense is the following:
> 
> * spice module can be moved to `sunkit-instruments`

We should ask if the spice team want it in their spice package.

> * `exposure_time_correction` can be removed, as users can now directly divide the cube by the exposure time correction thanks to arithmetic operations.

Agreed.

> * `instrument_axes` can be put onto the `.meta` as an axis-aligned piece of metadata.

Agreed.

> 
> This leaves the
> 
> * solar slit spectrograph metadata API
>   
>   * I think this really belongs in sunpy core, but should be part of a wider metadata API policy.

This is tricky, do we think we have any time in this year to try and work on this and get a wider metadata API in core/somewhere?

> * The `RasterSequence`, whose main power is its interchangeable 4-D/3-D with an instrumentally meaningful API
>   
>   * I don't foresee this living anywhere other than on a slit-spectrograph-specific subclass of `NDCube`.
> * Coordinate properties, e.g. `spectral_axis`.
>   
>   * Again, I don't see this living anywhere other than on a slit-spectrograph-specific subclass of `NDCube`.  Unless something clever can be done with a `Coords` class that has been mentioned before.  Even so, I don't see this in the foreseeable future.
>   * Alternatively, the same result can be achieved slightly less conveniently through `axis_world_coords`.

If sunraster became just these two items, I would be ok with it but it feels like the package would be on its last legs. I would be ok with that, it would not require much maintenance to keep it going with those items.

DanRyanIrish:
> > * spice module can be moved to `sunkit-instruments`
> 
> We should ask if the spice team want it in their spice package.

That would be even better.

> >   * I think this really belongs in sunpy core, but should be part of a wider metadata API policy.
> 
> This is tricky, do we think we have any time in this year to try and work on this and get a wider metadata API in core/somewhere?

I don't know, but I think it should be a priority for `sunpy`.  I banged the drum on this topic a good while ago, but didn't let the ball drop.  I think we need a renewed effort, but should give some thought about how to get more people interested.

> > * The `RasterSequence`, whose main power is its interchangeable 4-D/3-D with an instrumentally meaningful API
> >   
> >   * I don't foresee this living anywhere other than on a slit-spectrograph-specific subclass of `NDCube`.
> > * Coordinate properties, e.g. `spectral_axis`.
> >   
> >   * Again, I don't see this living anywhere other than on a slit-spectrograph-specific subclass of `NDCube`.  Unless something clever can be done with a `Coords` class that has been mentioned before.  Even so, I don't see this in the foreseeable future.
> >   * Alternatively, the same result can be achieved slightly less conveniently through `axis_world_coords`.
> 
> If sunraster became just these two items, I would be ok with it but it feels like the package would be on its last legs. I would be ok with that, it would not require much maintenance to keep it going with those items.

I think you're right that a package for 2 features would feel overkill.  But those two features are useful.  I think we'd need to think more about the future of `sunraster` at that point.

DanRyanIrish:
Hi @ebuchlin.  This is a confusing place to have this conversation, which is my fault.  Apologies.  I did not mean that functionality should be moved from `ndcube` to `sospice`.  I was suggesting passing the following to `sospice` from `sunraster`.
* the `spice` module, including the spice FITS reader and `SPICEMeta` class.

This would leave two main functionalities in `sunraster`:
* The generic metadata API definitions in the `meta.py` module on which the `SPICEMeta` class depends;
* The `SpectrogramCube`, `SpectrogramSequence`, and `RasterSequence` data classes.

ndcube/meta.py
```diff
@@ -0,0 +1,373 @@
...
+
+    Note that because rebinning does not drop axes, axis-aligned metadata is unaltered
+    by rebinning. By contrast, grid-aligned metadata must necessarily by affected by
+    rebinning. However, how it is affected depends on the nature of the metadata and
+    there is no generalized solution. Therefore, this class does not alter the shape
+    or values of grid-aligned metadata during rebinning, but simply removes its
+    axis-awareness.  If specific pieces of metadata have a known way to behave during
+    rebinning, this can be handled by subclasses or mixins.
+    """
+    def __init__(self, header=None, comments=None, axes=None, data_shape=None):
```
DanRyanIrish, line 75:
```suggestion
    def __init__(self, meta=None, comments=None, axes=None, data_shape=None):
```

ndcube/meta.py
```diff
@@ -0,0 +1,373 @@
...
+        self.__ndcube_can_slice__ = True
+        self.__ndcube_can_rebin__ = True
+        self.original_header = header
```
DanRyanIrish, line 78:
```suggestion
        self.original_meta = meta
```

ndcube/meta.py
```diff
@@ -0,0 +1,373 @@
...
+
+        if header is None:
+            header = {}
+        else:
+            header = dict(header)
+        super().__init__(header.items())
+        header_keys = header.keys()
```
DanRyanIrish, line 85:
```suggestion
        if meta is None:
            meta = {}
        else:
            meta = dict(header)
        super().__init__(meta.items())
        meta_keys = meta.keys()
```

ndcube/meta.py
```diff
@@ -0,0 +1,373 @@
...
+
+        if comments is None:
+            self._comments = dict()
+        else:
+            comments = dict(comments)
+            if not set(comments.keys()).issubset(set(header_keys)):
```
DanRyanIrish, line 91:
```suggestion
            if not set(comments.keys()).issubset(set(meta_keys)):
```

ndcube/meta.py
```diff
@@ -0,0 +1,373 @@
...
+
+        if axes is None:
+            self._axes = dict()
+        else:
+            if not (isinstance(data_shape, collections.abc.Iterable) and
+                    all([isinstance(i, numbers.Integral) for i in data_shape])):
+                raise TypeError("If axes is set, data_shape must be an iterable giving "
+                                "the length of each axis of the associated cube.")
+            axes = dict(axes)
+            if not set(axes.keys()).issubset(set(header_keys)):
```
DanRyanIrish, line 109:
```suggestion
            if not set(axes.keys()).issubset(set(meta_keys)):
```

ndcube/meta.py
```diff
@@ -0,0 +1,373 @@
...
+                raise ValueError(
+                    "All axes must correspond to a value in header under the same key.")
+            self._axes = dict([(key, self._sanitize_axis_value(axis, header[key], key))
```
DanRyanIrish, line 112:
```suggestion
            self._axes = dict([(key, self._sanitize_axis_value(axis, meta[key], key))
```

ndcube/meta.py
```diff
@@ -0,0 +1,373 @@
...
+    A sliceable object for storing metadata.
+
+    Metadata can be linked to a data axis which causes it to be sliced when the
+    standard Python numeric slicing API is applied to the object.
+    Specific pieces of metadata can be obtain using the dict-like string slicing API.
+    Metadata associated with an axis/axes must have the same length/shape as those axes.
+
+    Parameters
+    ----------
+    header: dict-like
```
DanRyanIrish, line 21:
```suggestion
    meta: dict-like
```

ndcube/meta.py
```diff
@@ -0,0 +1,373 @@
...
+
+    comments: dict-like, optional
+        Comments associated with any of the above pieces of metadata.
+
+    axes: dict-like, optional
+        The axis/axes associated with the metadata denoted by the keys.
+        Metadata not included are considered not to be associated with any axis.
+        Each axis value must be an iterable of `int`. An `int` itself is also
+        acceptable if the metadata is associated with a single axis.
+        The value of axis-assigned metadata in header must be same length as
```
DanRyanIrish, line 32:
```suggestion
        The value of axis-assigned metadata in meta must be same length as
```

ndcube/meta.py
```diff
@@ -0,0 +1,373 @@
...
+        super().__init__(header.items())
+        header_keys = header.keys()
+
+        if comments is None:
+            self._comments = dict()
+        else:
+            comments = dict(comments)
+            if not set(comments.keys()).issubset(set(header_keys)):
+                raise ValueError(
+                    "All comments must correspond to a value in header under the same key.")
```
DanRyanIrish, line 93:
```suggestion
                    "All comments must correspond to a value in meta under the same key.")
```

ndcube/meta.py
```diff
@@ -0,0 +1,373 @@
...
+            self._axes = dict()
+        else:
+            if not (isinstance(data_shape, collections.abc.Iterable) and
+                    all([isinstance(i, numbers.Integral) for i in data_shape])):
+                raise TypeError("If axes is set, data_shape must be an iterable giving "
+                                "the length of each axis of the associated cube.")
+            axes = dict(axes)
+            if not set(axes.keys()).issubset(set(header_keys)):
+                raise ValueError(
+                    "All axes must correspond to a value in header under the same key.")
```
DanRyanIrish, line 111:
```suggestion
                    "All axes must correspond to a value in meta under the same key.")
```

DanRyanIrish:
@Cadair, as requested, an explainer on the API/behaviour has been added to the PR description.

ndcube/meta.py
```diff
@@ -0,0 +1,373 @@
...
+    by rebinning. By contrast, grid-aligned metadata must necessarily by affected by
+    rebinning. However, how it is affected depends on the nature of the metadata and
+    there is no generalized solution. Therefore, this class does not alter the shape
+    or values of grid-aligned metadata during rebinning, but simply removes its
+    axis-awareness.  If specific pieces of metadata have a known way to behave during
+    rebinning, this can be handled by subclasses or mixins.
+    """
+    def __init__(self, header=None, comments=None, axes=None, data_shape=None):
+        self.__ndcube_can_slice__ = True
+        self.__ndcube_can_rebin__ = True
```
Cadair, line 77: These should be on the class i.e.
```python
class Meta(dict):
    __ndcube_can_slice__ = True
   ...
   def __init__...
```

ndcube/meta.py
```diff
@@ -0,0 +1,373 @@
...
+        super().__setitem__(key, val)
+
+    def __getitem__(self, item):
+        # There are two ways to slice:
+        # by key, or
+        # by typical python numeric slicing API,
+        # i.e. slice the each piece of metadata associated with an axes.
+
+        if isinstance(item, str):
+            return super().__getitem__(item)
```
Cadair, line 215: I would suggest we split out the  cube-like slicing from `__getitem__` into a property so you do something like `mymeta.slice[10:20]` if you want to do cube-like slicing.

ndcube/meta.py
Cadair: I am in favour of `meta` rather than `header`

ndcube/meta.py
```diff
@@ -0,0 +1,373 @@
...
+    @property
+    def axes(self):
+        return self._axes
+
+    @property
+    def shape(self):
+        return self._data_shape
+
+    def add(self, name, value, comment=None, axis=None, overwrite=False):
+        """Add a new piece of metadata to instance.
```
Cadair, line 153:
```suggestion
        """
        Add a new piece of metadata to instance.
```

ndcube/meta.py
Cadair, line 75: I am not a big fan of having `data_shape` as a kwarg. It generally feels clunky when you are making this and a cube together. What I would suggest is making it so that the items axes shapes are checked that they are consistent with each other, and then this is used to build up a internal partial shape to validate against.

Then inside `NDCube` we should have a way of telling the meta (in the `.meta` setter maybe) what shape the cube is i.e. doing `mymeta.data_shape = (10, 20, 30)` will validate that any axes which have axis or grid metadata associated with them match the shape passed in.

ndcube/meta.py
```diff
@@ -0,0 +1,374 @@
...
+        else:
+            self._data_shape = np.asarray(data_shape, dtype=int)
+
+        if axes is None:
+            self._axes = dict()
+        else:
+            if not (isinstance(data_shape, collections.abc.Iterable) and
+                    all([isinstance(i, numbers.Integral) for i in data_shape])):
+                raise TypeError("If axes is set, data_shape must be an iterable giving "
+                                "the length of each axis of the associated cube.")
```
Cadair, line 107: I would move this to the block above about data_shape?

ndcube/meta.py
```diff
@@ -0,0 +1,374 @@
...
+    @property
+    def comments(self):
+        return self._comments
+
+    @property
+    def axes(self):
+        return self._axes
+
+    @property
+    def shape(self):
```
Cadair, line 149: We should be consistent in how we name this, I think `data_shape` is better:
```suggestion
    def data_shape(self):
```

ndcube/meta.py
```diff
@@ -0,0 +1,374 @@
...
+            self._comments[name] = comment
+        if axis is not None:
+            axis = self._sanitize_axis_value(axis, value, name)
+            self._axes[name] = axis
+        elif name in self._axes:
+            del self._axes[name]
+        # This must be done after updating self._axes otherwise it may error.
+        self.__setitem__(name, value)
+
+    def remove(self, name):
```
Cadair, line 188: I think you want to maybe make this `__delitem__` so `del mymeta[name]` works. I don't see any particular reason why we need `.remove` as well.

ndcube/meta.py
```diff
@@ -0,0 +1,374 @@
...
+                    raise TypeError("Unrecognized slice type. "
+                                    "Must be an int, slice and tuple of the same.")
+            kept_axes = np.invert(dropped_axes)
+            new_meta._data_shape = new_shape[kept_axes]
+
+            # Slice all metadata associated with axes.
+            for key, value in self.items():
+                axis = self.axes.get(key, None)
+                drop_key = False
+                if axis is not None:
```
Cadair, line 257: I think it would be neater if this was:
```suggestion
                if axis is None:
                    continue
```

ndcube/meta.py
```diff
@@ -0,0 +1,374 @@
...
+                    else:
+                        value_is_axis_aligned = _is_axis_aligned(value, axis_shape)
+                        if value_is_axis_aligned:
+                            new_item = kept_axes[axis]
+                        else:
+                            new_item = tuple(item[axis])
+                        # Slice metadata value.
+                        try:
+                            new_value = value[new_item]
+                        except:
```
Cadair, line 284: this needs an exception type. No Pokemon exceptions.

ndcube/ndcube.py
```diff
@@ -1208,13 +1208,21 @@ def my_propagate(uncertainty, data, mask, **kwargs):
         # Resample WCS
         new_wcs = ResampledLowLevelWCS(self.wcs.low_level_wcs, bin_shape[::-1])
 
+        # If meta is axis-aware, drop axis-awareness for metadata associated with rebinned axes.
+        if hasattr(self.meta, "__ndcube_can_rebin__") and self.meta.__ndcube_can_rebin__:
+            new_shape = (np.asarray(self.shape) / np.asarray(bin_shape)).astype(int)
+            rebinned_axes = set(np.where(bin_shape != 1)[0])
+            new_meta = self.meta.rebin(rebinned_axes, new_shape)
+        else:
+            new_meta = self.meta
```
Cadair, line 1217: should this have always copied the meta? :thinking:

ndcube/conftest.py
```diff
@@ -334,6 +334,24 @@ def ndcube_4d_ln_lt_l_t(wcs_4d_t_l_lt_ln):
...
+    meta = NDMeta({"a": "scalar",
+                 "slit position": np.arange(shape[0], dtype=int),
+                 "pixel label": np.arange(np.prod(shape[:2])).reshape(shape[:2]),
+                 "line": ["Si IV"] * shape[2],
+                 "exposure time": ([2] * shape[-1]) * u.s},
+                data_shape=shape,
+                axes={"slit position": 0,
+                      "pixel label": (0, 1),
+                      "line": (2,),
+                      "exposure time": 3})
```
Cadair, line 351: this indentation is upsetting :stuck_out_tongue:

Cadair (commented):
This also needs some narrative docs to teach people how and why they should use it.

ndcube/meta.py
DanRyanIrish: Leave it up to the user to not apply a slice item that doesn't also work for the object they've entered as grid-aligned metadata.  So replace the `try...except...` block with L284.  (Confirm I understand the implications of this correcty.)

DanRyanIrish: Actually, this is needed for setting a tuple, list, etc. as axis-aligned metadata.  Otherwise, dropping multiple axes during slicing will not be supported.

ndcube/ndcube.py
DanRyanIrish: Not sure what you mean here.

DanRyanIrish:
@Cadair, except for narrative docs, I think this PR now addresses all but one minor comment of yours.  Unless you'd like to look at it again, I will merge it once I've added narrative docs.

ndcube/ndcube.py
samaloney: I assume `self.meta.copy()`

ndcube/meta.py
```diff
@@ -0,0 +1,466 @@
...
+                # Slice metadata value.
+                try:
+                    new_value = value[new_item]
+                except:
+                    # If value cannot be sliced by fancy slicing, convert it
+                    # it to an array, slice it, and then if necessary, convert
+                    # it back to its original type.
+                    new_value = (np.asanyarray(value)[new_item])
+                    if hasattr(new_value, "__len__"):
+                        new_value = type(value)(new_value)
```
Cadair, line 422: Did we say we were going to nuke this bit?

ndcube/ndcube.py
Cadair: I mean should rebin have always made a copy of meta, i.e so the rebinned meta isn't the same object as the parent?

ndcube/ndcube.py
```diff
@@ -399,6 +400,10 @@ def __init__(self, data, wcs=None, uncertainty=None, mask=None, meta=None,
                 global_coords = deepcopy(global_coords)
             self._global_coords = global_coords
 
+        # If meta is axis-aware, make it to have same shape as cube.
+        if isinstance(self.meta, NDMetaABC):
+            self.meta.data_shape = self.shape
```
Cadair, line 405: Should this be in the setter for `.meta`? Do we have a setter for `.meta`?

ndcube/meta.py
```diff
@@ -0,0 +1,466 @@
...
+        """
+
+
+class NDMeta(dict, NDMetaABC):
+    # Docstring in ABC
+    __ndcube_can_slice__ = True
+    __ndcube_can_rebin__ = True
+
+    def __init__(self, meta=None, comments=None, axes=None):
+        self.original_meta = meta
```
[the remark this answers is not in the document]
DanRyanIrish: Sounds reasonable.  Should original meta be part of the ABC too?

ndcube/meta.py
```diff
@@ -0,0 +1,466 @@
...
+        self._data_shape = np.array([], dtype=int)
+
+        if meta is None:
+            meta = {}
+        else:
+            meta = dict(meta)
+        super().__init__(meta.items())
+        meta_keys = meta.keys()
+
+        if comments is None:
```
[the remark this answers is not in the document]
DanRyanIrish: I have no objections to this.  I'll make the change.

ndcube/meta.py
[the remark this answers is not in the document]
Cadair: I don't think it's a hard requirement? so no?

ndcube/meta.py
```diff
@@ -0,0 +1,466 @@
...
+            self._comments = dict()
+        else:
+            comments = dict(comments)
+            if not set(comments.keys()).issubset(set(meta_keys)):
+                raise ValueError(
+                    "All comments must correspond to a value in meta under the same key.")
```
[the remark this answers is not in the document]
DanRyanIrish: You mean have a
```python
@key_comments.setter
def key_comments(self, name, value):
    if name not in self:
        raise ValueError()
```

ndcube/meta.py
```diff
@@ -0,0 +1,466 @@
...
+        new_meta = copy.deepcopy(self.meta)
+        naxes = len(data_shape)
+        if isinstance(item, (numbers.Integral, slice)):
+            item = [item]
+        if len(item) < naxes:
+            item = np.array(list(item) + [slice(None)] * (naxes - len(item)), dtype=object)
+        elif len(item) > naxes:
+            # If item applies to more axes than have been defined in NDMeta,
+            # ignore items applying to those additional axes.
+            item = np.array(item[:naxes])
```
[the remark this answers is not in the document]
DanRyanIrish: This could happen now because `data_shape` is determined only from the metadata in `NDMeta` unless explicitly set by the user.  So imagine the user does not explicitly give the `data_shape`, and the `NDMeta` is applied to 4-D, but there is only metadata that applies to the 2nd axis.  In that case, the `data_shape` will be `(0, n)` where `n` is the length of the 2nd axis.  But a valid `item` could be a `tuple` of 4 `int`/`slice`, because the data has 4 axes.  To make things simple, we want the `item` to be the same length as the `data_shape`.  So we discard the trailing elements `item` as they will not be used.

ndcube/meta.py
```diff
@@ -0,0 +1,466 @@
...
+                    start = data_shape[i] - start
+                stop = axis_item.stop
+                if stop is None:
+                    stop = data_shape[i]
+                if stop < 0:
+                    stop = data_shape[i] - stop
+                new_shape[i] = stop - start
+            else:
+                raise TypeError("Unrecognized slice type. "
+                                "Must be an int, slice and tuple of the same.")
```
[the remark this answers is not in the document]
DanRyanIrish: As in:
```suggestion
        new_shape = np.broadcast_to(1, new_meta.data_shape)[item].shape
        dropped_axes = np.array([True if isinstance(axis_item, numbers.Integral) else False for axis_item in item], dtype=bool)
```

ndcube/meta.py
DanRyanIrish: I left a comment about this somewhere after our talk.  I realised that without this, `NDMeta` could not support dropping axes as part of slicing if it contained lists or tuples as axis-aligned metadata.  So as ugly as it is, I think it's better to keep it.

ndcube/ndcube.py
DanRyanIrish: We don't.  We outsource that to `NDData`.

ndcube/ndcube.py
DanRyanIrish: Ah.  I guess yes.  I'll make this line include copy.

ndcube/meta.py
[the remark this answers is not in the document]
DanRyanIrish: Alternatively, we could allow slicing to adjust `data_shape`.  So if `data_shape` starts as `(0, 10)`, a slice item of `(slice(0, 3), slice(0, 6), slice(None), slice(0, 4))` would lead to a new data shape of `(3, 6, 0, 4)`.

ndcube/meta.py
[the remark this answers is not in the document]
DanRyanIrish: @Cadair Can you give a clarifying answer here?

ndcube/meta.py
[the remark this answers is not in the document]
DanRyanIrish: @Cadair Are we happy with this as is?

DanRyanIrish:
ping @Cadair.  A friendly reminder for a review on this PR :)

docs/explaining_ndcube/metadata.rst
```diff
@@ -0,0 +1,268 @@
...
+Exposure time is therefore metadata, not a coordinate.
+
+One reason why it is important to distinguish between coordinates and axis-aware metadata is `ndcube`'s dependence on WCS.
+Most WCS implementations require that there be a unique invertible mapping between pixel and world coordinates, i.e., there is only one pixel value that corresponds to a specific real world value (or combination of such if the coordinate is multi-dimensionsal), and vice versa.
+Therefore, while there may be exceptions for rare and exotic WCS implementations, a good rule of thumb for deciding whether something is a coordinate is:
+coordinates are numeric and strictly monotonic.  Otherwise you have metadata.
+
+The keen-eyed reader may have realised of the above framework that, while not all axis-aligned metadata can be treated as coordinates, all coordinates can be treated like axis-aware metadata.
+This raises the question of why not dispense with coordinates altogether and only have axis-aligned metadata?
+The reason is that the stricter requirements on coordinates have led to a host of powerful coordinate infrastructures that are not valid for generalised axis-aware metadata.
```
Cadair, line 46:
```suggestion
The reason is that the stricter requirements on coordinates have led to a host of powerful coordinate infrastructure that are not valid for generalised axis-aware metadata.
```

docs/explaining_ndcube/metadata.rst
```diff
@@ -0,0 +1,268 @@
...
+
+.. code-block:: python
+
+  >>> meta.add("row temperature", [-10, -11, -12] * u.deg_C, axes=1)
+  >>> meta.data_shape
+  array([5, 3, 2])
+
+.. _assigning_axis_aligned_metadata:
+
+Assigning Axis-aligned Metadata
```
Cadair, line 249: "Axis aligned metadata" wasn't a phrase which immediately described what this was to me. My initial thought at the difference between "grid aligned" and "axis aligned" was that grid is aligned to >1 axis and axis aligned was 1.

I am not sure I have a better word, but it's weird.

docs/explaining_ndcube/slicing.rst
```diff
@@ -416,3 +416,110 @@ Let's say we axis order of the ``linewidths`` cube was reversed.
   array([2, 1], dtype=object)
 
 The same result is obtained.
+
+.. _meta_slicing:
+
+Slicing NDMeta
+--------------
```
Cadair, line 423:
```suggestion
==============
```

docs/explaining_ndcube/slicing.rst
```diff
@@ -416,3 +416,110 @@ Let's say we axis order of the ``linewidths`` cube was reversed.
...
+.. code-block:: python
+
+  >>> sliced_meta["salutation"]
+  'hello'
+  >>> sliced_meta["name"]
+  'world'
+
+
+Automatically Slicing NDMeta Attached to Other ND Objects
+*********************************************************
```
Cadair, line 482:
```suggestion
---------------------------------------------------------
```

docs/explaining_ndcube/slicing.rst
```diff
@@ -416,3 +416,110 @@ Let's say we axis order of the ``linewidths`` cube was reversed.
...
+
+`~ndcube.NDMeta` has a dunder property called ``__ndcube_can_slice__``, which, by default, is set to ``True``.
+When attached to `~ndcube.NDCube`, `~ndcube.NDCubeSequence`, and `~ndcube.NDCollection`, this property lets those classes know that that the metadata object is sliceable.
+Hence, when their ``__getitem__`` methods are called (i.e. when they are sliced), they apply the slice item to ``self.meta.slice`` and attach the result as the meta attribute of the newly sliced ND object.
```
Cadair, line 486: This is very developer focused. I think we should skip how this works here and just say "when attached to cube, slicing happens".

ndcube/mixins/ndslicing.py
```diff
@@ -19,10 +19,25 @@ def __getitem__(self, item):
...
             raise IndexError("None indices not supported")
 
+        # If cube has a sliceable metadata, remove it and handle it separately.
+        # This is to prevent the shapes of the data and metadata getting out of
+        # sync part way through the slicing process.
+        meta_is_sliceable = False
+        if hasattr(self.meta, "__ndcube_can_slice__") and self.meta.__ndcube_can_slice__:
+            meta_is_sliceable = True
+            meta = self.meta
+            self.meta = None
```
Cadair, line 29: Does this not mean that the original cube you are slicing now has no metadata after slicing it?

ndcube/tests/test_ndcube.py
```diff
@@ -168,6 +169,32 @@ def test_slicing_removed_world_coords(ndcube_3d_ln_lt_l):
...
+    wcs.wcs.crval = 10, 0.5, 1
+    cube = NDCube(data, wcs=wcs, meta=meta)
+    sliced_cube = cube[0, 1:3]
+    sliced_meta = sliced_cube.meta
+    assert sliced_meta.keys() == meta.keys()
+    assert tuple(sliced_meta.axes.keys()) == ("pixel response",)
+    assert sliced_meta["salutation"] == meta["salutation"]
+    assert (sliced_meta["pixel response"] == meta["pixel response"][1:3]).all()
+    assert sliced_meta["exposure time"] == 2 * u.s
+
```
Cadair, line 196:
```suggestion
    assert cube.meta is meta
```

ndcube/meta.py
```diff
@@ -0,0 +1,489 @@
...
+
+    Note that because rebinning does not drop axes, axis-aligned metadata is unaltered
+    by rebinning. By contrast, grid-aligned metadata must necessarily by affected by
+    rebinning. However, how it is affected depends on the nature of the metadata and
+    there is no generalized solution. Therefore, this class does not alter the shape
+    or values of grid-aligned metadata during rebinning, but simply removes its
+    axis-awareness.  If specific pieces of metadata have a known way to behave during
+    rebinning, this can be handled by subclasses or mixins.
+    """
+
```
Cadair, line 73: The ABC wont enforce this (I don't think there's a way) but we should add them here for completeness
```suggestion

    __ndcube_can_slice__: bool
    __ndcube_can_rebin__: bool
    
```

ndcube/meta.py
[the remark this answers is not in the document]
Cadair: That's the idea, but that wouldn't work. We would need to be validating that when we add an element to `key_comments` not when we are setting `NDMeta.key_comments`. I am not exactly sure how we would do that without making `key_comments` be a special dict subclass.

The point stands though that we have no runtime validation of this, is that a problem?

ndcube/meta.py
```diff
@@ -0,0 +1,489 @@
...
+            meta = {}
+        super().__init__(meta.items())
+        meta_keys = meta.keys()
+
+        if key_comments is None:
+            self._key_comments = {}
+        else:
+            if not set(key_comments.keys()).issubset(set(meta_keys)):
+                raise ValueError(
+                    "All comments must correspond to a value in meta under the same key.")
```
Cadair, line 164:
```suggestion
                raise ValueError(
                    "All comments must correspond to a value in meta under the same key."
                )
```

ndcube/meta.py
```diff
@@ -0,0 +1,489 @@
...
+            axes = dict(axes)
+            if not set(axes.keys()).issubset(set(meta_keys)):
+                raise ValueError(
+                    "All axes must correspond to a value in meta under the same key.")
+            self._axes = {key:self._sanitize_axis_value(axis, meta[key], key)
+                          for key, axis in axes.items()}
+
+    def _sanitize_axis_value(self, axis, value, key):
+        axis_err_msg = ("Values in axes must be an integer or iterable of integers giving "
+                        f"the data axis/axes associated with the metadata.  axis = {axis}.")
```
Cadair, line 179:
```suggestion
                        f"the data axis/axes associated with the metadata. axis = {axis}.")
```

ndcube/meta.py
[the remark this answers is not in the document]
Cadair: yeah makes sense.

ndcube/meta.py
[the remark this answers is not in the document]
Cadair: yes?

ndcube/meta.py
```diff
@@ -0,0 +1,489 @@
...
+def _not_scalar(value):
+    return (
+        (
+         hasattr(value, "shape")
+         or hasattr(value, "__len__")
+        )
+        and not
+        (
+         isinstance(value, str)
+        ))
```
Cadair, line 468: This indentation is wild.

Cadair (approved):
Can't see any major issues. :shipit:

docs/explaining_ndcube/metadata.rst
DanRyanIrish: Yeh, I take the point.  I don't have a better phrase either though.

docs/explaining_ndcube/slicing.rst
DanRyanIrish:
```suggestion
`~ndcube.NDMeta` has a dunder property called ``__ndcube_can_slice__``, which, by default, is set to ``True``.
When attached to `~ndcube.NDCube`, `~ndcube.NDCubeSequence`, and `~ndcube.NDCollection`, this property lets those classes know to automatically slice the `~ndcube.NDMeta` along with its other sliceable attributes.
```

ndcube/mixins/ndslicing.py
```diff
@@ -19,10 +19,26 @@ def __getitem__(self, item):
...
+
+        # Slice cube.
         item = tuple(sanitize_slices(item, len(self.shape)))
         sliced_cube = super().__getitem__(item)
+        self.meta = meta  # Add unsliced meta back onto  unsliced cube.
```
DanRyanIrish, line 34:
```suggestion
        if meta_is_sliceable:
            self.meta = meta  # Add unsliced meta back onto unsliced cube.
```

ndcube/mixins/ndslicing.py
DanRyanIrish: Yes.  It's fixed now.

DanRyanIrish:
@nabobalis Can you interpret [this doc error](https://github.com/sunpy/ndcube/actions/runs/12324330739/job/34401759004?pr=455) for me?

samaloney:
I think you need to add the warning to the nitpick-exceptions file so something like this but no 100% sure of the syntax.

```
py:class Update D from mapping/iterable E and F
```

docs/nitpick-exceptions
nabobalis, line 20:
```suggestion
py:class Update D from mapping/iterable E and F
```
From Shane

DanRyanIrish: Thanks @nabobalis and @samaloney Doesn't seem to have fixed it though...  Any other ideas?

docs/nitpick-exceptions
```diff
@@ -16,3 +16,5 @@ py:class a set-like object providing a view on D's keys
 py:class an object providing a view on D's values
 py:class None.  Remove all items from D.
 py:class a shallow copy of D
+py:class v, remove specified key and return the corresponding value.
+py:class Update D from mapping/iterable E and F
```
nabobalis, line 20:
```suggestion
py:class None.  Update D from mapping/iterable E and F.
```
Let us try this

samaloney: Best guess is this seems like has to match the exact text
```suggestion
py:class None.  Update D from dict/iterable E and F.
py:class None.  Update D from mapping/iterable E and F.
```

PR #456, opened 2021-08-14
Sequence plotter

### LINKED ISSUE #315: Should ndcube 2.x support NDCubeSequence plotting?
**TL;DR it has been decided to drop `NDCubeSequence` plotting for 2.0.  The rationale is explained below.  If you feel this should be re-instated, please let us know by leaving a comment below telling us about your use case.  Better still, if you would like to help build the tools needed to make this functionality possible again, please see the sections below about the Ideal Solution and let us know in a comment below.**

The changes made for ndcube 2.x are widespread and breaking.  They have direct impacts on the `NDCubeSequence` plotting infrastructure.  Although in principle, `NDCubeSequence` plotting is clearly desirable, its use cases are thought to be much fewer than its `NDCube` counterpart.  Whether the significant effort required to update and maintain this functionality is worthwhile is the subject of this issue.

### Background: Comparison to `NDCube` plotting
In ndcube 2.x, `NDCube` moves its plotting at an entirely `astropy` `WCSAxes` based environment.  This is made possible because `NDCube`'s `extra_coords` can now be represented as an APE 14 WCS and can be combined into a single WCS with the `NDCube`'s primary WCS (located at `NDCube.wcs`).  Moreover, this transition has shown the value of `WCSAxes`'s ability to represent multiple world coordinates along single array axes, especially in the case of dependent axes, e.g. lat, lon.  This is not possible with the ndcube 1.x `NDCubeSequence` plotting infrastructure, nor is it possible unless the `NDCubeSequence` coordinates can be represented in a single WCS. However, because `NDCubeSequence` has a separate WCS for each `NDCube` within it, it is not trivial to build the tools required to convert them into a single sequence WCS.  However, if this was possible, it would open up new possibilities.  **If you would like to work on this, see the sections on the "Preferred Solution" and leave a comment.**

### Impacts on Specific `NDCubeSequence` Plotting Methods
We define two broad types of consequence on the `NDCubeSequence` plotting methods: **outmoded** and **broken**.  **Outmoded** means that, with effort, the 1.x behavior can be recreated, but that this behavior is not strictly correct and does not match the `NDCube` counterpart unless a sequence WCS can be created.  For example, dependent coordinates cannot be correctly represented and multiple coordinates cannot be easily shown on the same axis.  **Broken** means that the current behavior cannot be recreated, nor can the `NDCube` counterpart behavior be mimicked unless a sequence WCS can be created.
* `plot_1D_sequence`: **Outmoded**.
* `plot_2D_sequence_as_1Dline`: **Outmoded**.
* `plot_2D_sequence`: **Outmoded**.
* `plot_3D_sequence_as_2Dimage`: **Outmoded**.
* `NDCubeSequenceAnimator`: **Broken**.  Works by appending a dummy sequence axis to the WCS object of the first cube in the sequence.  This is not possible with APE 14.

#### Ideal, if Effort Resources Abundant
The ideal solution would be to build a tool that can convert a sequence of `NDCube` WCSs into a single sequence WCS.  Then the sequence's data could be stacked into a single array and the `NDCube` plotting infrastructure could be used for sequence.  Separate `NDCubeSequence` infrastructure would then be minimal.  It would only be required to create the sequence WCS and stack the data.  This is beyond the effort levels available for 2.0.  **If you would like to contribute, let us know.**

#### Easiest for Developers
Do not support sequence plotting in 2.0.  Raise a `NotImplementedError` with a message asking users to comment on this issue to let us know whether this is a feature we need to support.  A recipe guide of how to manually create visualizations of `NDCubeSequence`s could also be added to the docs.  The error could also point users to that guide.

#### Replicate as Best as Possible
Re-jig API to allow users to enter a coordinate name from the `wcs`/`extra_coords`/`combined wcs` to show on the axis.  The values will be 1-D, ignoring dependency on other coordinates.  Values on plotted axes will represent the values along the 0th row/column of the cube and slider axes values will represent the mean value taken along that axis.  This should be easier to adapt from the existing code although will still take some work.  However, it will also produce an approximate but sometimes incorrect representation of the data.

#### Compromise: Limited Support
Implementing plotting where there is *no regridding of the data at all*, is worthwhile and not that complex. This would result in `NDCubeSequence.plot` always returning an animation, where there will always be a slider for the sequence axis. This is a significant downgrade in functionality from 1.x, but less drastic than dropping support altogether.  And we feel that it aligns better with the abstraction for which `NDCubeSequence` exists, i.e. cubes with different pixel spaces.

The summary of the proposed implementation is that a new animator subclass will be created, which delegates to `NDCube.plot()` for all actual plotting, with a slider being generated for the sequence axis and then every non-axes coordinate in each subcube. This is similar to how [`MapSequenceAnimator`](https://github.com/sunpy/sunpy/blob/master/sunpy/map/mapsequence.py#L253) works in sunpy core currently.

### Decision for 2.0
The preferred solution is unfortunately beyond the effort resources available to us before the 2.0 release.  While the "Replicate as Best as Possible" method should be less effort, it is still significant and will produce a sub-standard product.  Moreover, it would require some small breaking changes to the API (e.g. only allowing `str` `axes_coordinates` and allow them to refer to the `wcs` coordinates, not just the `extra_coords`).  Dropping support altogether is punitive, especially since there are several use cases in sunpy alone where users would want to animate a sequence-like object. We have therefore decided to adopt the "Limited Support" option for 2.0.  **Please let us know if you rely on this functionality, would like greater plotting support, or would like to work on it.**

### Wider Consequences of Ideal Solution
If it was possible to create an overarching sequence WCS which combined the various `NDCube` WCS's, it would likely render `NDCubeSequence` class obsolete and enable the data to be representing in a single `NDCube`.  The initial use case for `NDCubeSequence` was to combine data with difference WCSs, especially since at the time, FITSWCS could not handle discontinuities in the coordinate transformations.  This can be addressed with the more recent `gWCS` package.

### Initial Roadmap for Implementing Ideal Solution
The best solution is to create a gWCS object which is a composite of all the APE 14 WCSes of all the cubes in the sequence. This requires writing new models for `astropy.modeling`, one which wraps a `BaseLowLevelWCS` object as an astropy model and one which is lookup table-like model which takes a series of `BaseLowLevelWCS` objects and a lookup table for the third axis and exposes this as a singular N+1 D model. These models can then be passed to gWCS as the forward transform along with appropriate coordinate frame arguments. With this done, it would be possible to represent the current use cases for `NDCubeSequence` as a `NDCube` object as you would have a singular contiguous array with a single WCS.

DanRyanIrish:
### Description

This PR creates a plotter for `NDCubeSequence`  It delegates the plotting to `NDCube.plot`, but assumes that method uses a matplotlib-based animator.  The sequence axis is always a slider, i.e. is cannot be a plot axis, and plotting a sequence 1-D cubes is not yet supported.

Fixes #315


ndcube/visualization/mpl_plotter.py
```diff
@@ -251,3 +231,30 @@ def _as_mpl_axes(self):
         if n_dim > 2:
             kwargs['slices'] = ['x', 'y'] + [None] * (ndim - 2)
         return WCSAxes, kwargs
+
+    def _prep_animate_args(self, wcs, plot_axes, axes_units, data_unit):
+        # If data_unit set, convert data to that unit
+        if data_unit is None:
+            data = self._ndcube.data
+        else:
+            data = u.Quantity(self._ndcube.data, unit=self._ndcube.unit).to_value(data_unit)
```
Cadair, line 240:
```suggestion
            data = u.Quantity(self._ndcube.data, unit=self._ndcube.unit, copy=False).to_value(data_unit)
```

ndcube/visualization/sequence_mpl_plotter.py
```diff
@@ -0,0 +1,120 @@
...
+        # Calculate data and wcs for initial animation state and instantiate Animator.
+        data, wcs, plot_axes, coord_params = self._cubes[0].plotter._prep_animate_args(
+            self._cubes[0].wcs, self._plot_axes, self._axes_units, self._data_unit)
+        super().__init__(data, wcs, plot_axes, coord_params=coord_params, **base_kwargs)
+
+    def _sequence_slider_function(self, val, im, slider):
+        self._sequence_idx = int(val)
+        self.data, self.wcs, plot_axes, coord_params = self._cubes[self._sequence_idx].plotter._prep_animate_args(
+            self._cubes[self._sequence_idx].wcs, self._plot_axes, self._axes_units, self._data_unit)
+        return super().plot_start_image(self.axes)
```
Cadair, line 120: I am not convinced (from a quick look) that this call is correctly resetting the WCSAxes instance for the WCS of the new cube.

DanRyanIrish, line 120:
```suggestion
        return super().update_plot(self.axes)
```

DanRyanIrish:
@Cadair, this now works and the memory issues as well as axes resetting issues have been resolved.  I think it's ready to merge.

ndcube/visualization/descriptor.py
```diff
@@ -44,5 +44,36 @@ def __get__(self, obj, objtype=None):
 
     def __set__(self, obj, value):
         if not isinstance(value, type):
-            raise TypeError("NDCube.plotter can only be set with an uninitialised plotter object.")
+            raise TypeError(
+                "Plotter attribute can only be set with an uninitialised plotter object.")
         setattr(obj, self._attribute_name, value(obj))
+
+
+class SequencePlotterDescriptor(PlotterDescriptor):
```
Cadair, line 52: I don't think you need this new class, just add the special case of `self._default_type == "mpl_sequence_plotter"` to the other one.

ndcube/visualization/mpl_sequence_plotter.py
```diff
@@ -0,0 +1,131 @@
...
+        # Calculate data and wcs for initial animation state and instantiate Animator.
+        data, wcs, plot_axes, coord_params = self._cubes[0].plotter._prep_animate_args(
+            self._cubes[0].wcs, self._plot_axes, self._axes_units, self._data_unit)
+        if not isinstance(wcs, BaseLowLevelWCS):
+            wcs = wcs.low_level_wcs
+        super().__init__(data, wcs, plot_axes, coord_params=coord_params, **base_kwargs)
+
+    def _sequence_slider_function(self, val, artist, slider):
+        self._sequence_idx = int(val)
+        self.data, self.wcs, plot_axes, coord_params = self._cubes[self._sequence_idx].plotter._prep_animate_args(
```
Cadair, line 126: The `plot_axes` and `coord_params` aren't actually being used for anything here right? If not you could replace them with `_` to make that obvious.

ndcube/visualization/mpl_sequence_plotter.py
```diff
@@ -0,0 +1,131 @@
...
+        "Sunpy is required to animate NDCubeSequences. "
+        "Either install sunpy or extract data from sequence and visualize manually.")
+
+from .base import BasePlotter
+from .plotting_utils import prep_plot_kwargs
+
+__all__ = ['MatplotlibSequencePlotter']
+
+
+class MatplotlibSequencePlotter(BasePlotter):
```
Cadair, line 16: Rather than adding this class, you could dispatch on the type of `self._ndcube` in the `MatplotlibPlotter`? Or is this a prelude to more custom sequence specific plotting in the future?

Cadair (commented):
Looks sane, a couple of comments, the main one being not defining a new descriptor class.

Needs tests?

ndcube/visualization/mpl_sequence_plotter.py
DanRyanIrish: Yes, I have already have local implementations for creating a 2-D image and 1-D line plot in the cases where we have a sequence of 1-D cubes and a sequence of shape `(1,)` cubes.  But I was planning on submitting them after 2.0 is released.

DanRyanIrish:
Thanks for the review @Cadair.  I've made the changes you suggested and will merge once the tests pass.

ndcube/ndcube_sequence.py
```diff
@@ -6,10 +6,11 @@
 import numpy as np
 
 from ndcube import utils
+from ndcube.visualization import PlotterDescriptor
 
 __all__ = ['NDCubeSequence']
 
-PLOTTING_NOT_SUPPORTED_ERROR = """NDCubeSequence plotting is no longer supported.
+PLOT_AS_CUBE_NOT_SUPPORTED_ERROR = """NDCubeSequence plot_as_cube is no longer supported.
```
Cadair, line 13: Putting this here now seems redundant?

ndcube/visualization/descriptor.py
```diff
@@ -28,12 +32,18 @@ def __get__(self, obj, objtype=None):
                 try:
```
Cadair, line 32: You could probably save some lines here by putting the if block inside the the try block?

Cadair (approved):

PR #457, opened 2021-08-17
Remove ArithmeticMixin until we WCS it

Cadair:


DanRyanIrish:
:+1: Pending tests passing

DanRyanIrish:
Actually I think this does need a changelog and a breaking one at that.  Although it doesn't affect our codebase it's possible someone was using the arithmetic operations, even though they aren't coordinate aware.  SO for them, NDCube will now break when before it would do something.

DanRyanIrish (approved):

dhomeier:
`specutils` has been using the arithmetic throughout, so is massively broken by this change.

DanRyanIrish:
Thanks for pointing this out.  We clearly merged this too hastily.  Let's find the best solution so `specutils` is not too impacted.

dhomeier:
I think your proposed fix of subclassing directly to `NDArithmeticMixin` at least restores the previous functionality.
I don't think specutils `Spectrum1D.add` at this point intends to provide any functionality to actually co-add spectra co-add spectra on different grids; the only test is just using two spectral axes with different units, and not checking the results.
In fact `spectrum1 + spectrum2` just ignores differences in `spectral_axis` as long as the shapes match.

DanRyanIrish:
Thanks for the clarifications @dhomeier.  In that case we can leave this PR merged as is.

PR #459, opened 2021-08-20
Add util for identifying invariant axes in a WCS

### LINKED ISSUE #432: Detect Invariant Axes as part of NDCube.reproject_to
### Description

After an initial minimal implementation of `NDCube.reproject_to` has been made to work, the method needs to be able to detect invariant axes in the WCSs, i.e. axes over which no reprojection is to be performed.  This will open up a range of possibilities, such as more efficient reproject, supporting more reprojection algorithms, etc.

adwaitbhope:
### Description

Fixes #432.

### DIFF
--- a/ndcube/utils/wcs.py
+++ b/ndcube/utils/wcs.py
@@ -480,7 +480,7 @@ def compare_wcs_physical_types(source_wcs, target_wcs):
     return source_wcs.world_axis_physical_types == target_wcs.world_axis_physical_types
 
 
-def identify_invariant_axes(source_wcs, target_wcs, input_shape):
+def identify_invariant_axes(source_wcs, target_wcs, input_shape, atol=1e-6, rtol=1e-6):
     """
     Performs a pixel to pixel transformation to identify if there are any invariant axes
     between the given source and target WCS objects.
@@ -494,6 +494,12 @@ def identify_invariant_axes(source_wcs, target_wcs, input_shape):
     input_shape: `tuple`
         The array shape of the data.
 
+    atol: `float`
+        The absolute tolerance parameter for comparison.
+
+    rtol: `float`
+        The relative tolerance parameter for comparison.
+
     Returns
     -------
     result: `list`
@@ -505,5 +511,5 @@ def identify_invariant_axes(source_wcs, target_wcs, input_shape):
 
     output_pixel_coords = pixel_to_pixel(source_wcs, target_wcs, *input_pixel_coords)
 
-   return [np.allclose(input_coord, output_coord, atol=1e-6)
+    return [np.allclose(input_coord, output_coord, atol=atol, rtol=rtol)
             for input_coord, output_coord in zip(input_pixel_coords, output_pixel_coords)]
--- a/ndcube/utils/wcs.py
+++ b/ndcube/utils/wcs.py
@@ -8,6 +8,7 @@ import numbers
 from collections import UserDict
 
 import numpy as np
+from astropy.wcs.utils import pixel_to_pixel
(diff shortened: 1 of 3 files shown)


ndcube/utils/wcs.py
```diff
@@ -477,3 +478,35 @@ def compare_wcs_physical_types(source_wcs, target_wcs):
     target_wcs = get_low_level_wcs(target_wcs, 'target_wcs')
 
     return source_wcs.world_axis_physical_types == target_wcs.world_axis_physical_types
+
+
+def identify_invariant_axes(source_wcs, target_wcs, input_shape):
```
adwaitbhope, line 483: Do we need to have an `input_shape` here? Could we just use `pixel_n_dim` from the WCS and pass some static number to `np.arange()` below?

ndcube/utils/wcs.py
```diff
@@ -477,3 +478,35 @@ def compare_wcs_physical_types(source_wcs, target_wcs):
...
+
+    input_pixel_coords = np.meshgrid(*[np.arange(n) for n in input_shape])
+
+    output_pixel_coords = pixel_to_pixel(source_wcs, target_wcs, *input_pixel_coords)
+
+    input_pixel_coords = np.around(input_pixel_coords, decimals=6)
+    output_pixel_coords = np.around(output_pixel_coords, decimals=6)
+
+    return [(input_coord == output_coord).all()
+            for input_coord, output_coord in zip(input_pixel_coords, output_pixel_coords)]
```
DanRyanIrish, line 512:
```suggestion
   return [np.allclose(input_coord, output_coord, atol=1e-6)
            for input_coord, output_coord in zip(input_pixel_coords, output_pixel_coords)]
```

ndcube/utils/wcs.py
DanRyanIrish: I think we do need the `input_shape` here.

DanRyanIrish (commented):
Looks good!  One suggested change at this stage.

ndcube/utils/wcs.py
```diff
@@ -477,3 +478,32 @@ def compare_wcs_physical_types(source_wcs, target_wcs):
...
+    result: `list`
+        A list of booleans denoting whether the axis is invariant or not.
+        Follows the WCS ordering.
+    """
+
+    input_pixel_coords = np.meshgrid(*[np.arange(n) for n in input_shape])
+
+    output_pixel_coords = pixel_to_pixel(source_wcs, target_wcs, *input_pixel_coords)
+
+   return [np.allclose(input_coord, output_coord, atol=1e-6)
```
Cadair, line 508: Why atol? It might be worth exposing atol and rtol as kwargs in this function :thinking:

Cadair:
The approach looks good. :+1:

DanRyanIrish:
> The approach looks good. +1

Good!  Apart from your comment about exposing `atol` and `rtol`, what more needs to be done before merging this?  I *think* it's mergeable as is once that change is made?

DanRyanIrish:
> > The approach looks good. +1
> 
> Good! Apart from your comment about exposing `atol` and `rtol`, what more needs to be done before merging this? I _think_ it's mergeable once that change is made and the tests are passing.

adwaitbhope:
I thought the actual optimization of the `reproject_to` method would be a part of this PR itself. Should we do that here or open a separate PR?

DanRyanIrish:
It can be.  But this seems to me to be a useful tool in its own right doesn't have to be held up waiting for the `reproject_to` optimisation.  Either way, the above changes need to be made.  So make those and we can decide then whether to merge and start a new PR or push on with this one.

DanRyanIrish (approved):
@Cadair, any objections to merging this?

PR #461, opened 2021-08-22
Fix bug in NDCube.axis_world_coords.

DanRyanIrish:
### Description

This PR fixes a bug in `NDCube.axis_world_coords` which caused the coords to be converted to a `numpy.ndarray` when an axes input was given. This caused the unit information to be lost.

Fixes #<Issue Number>


Cadair:
This needs a test to make sure we don't reintroduce it. Also I think I might have a better fix but I am not 100% sure it will work lol

DanRyanIrish:
I've added a test.  If you do have a better implementation that would be great.  Feel free to add it as a suggestion if it's simple enough.

Cadair (approved):

PR #462, opened 2021-08-23
Add method to NDCubeSequence to stack data from cubes.

DanRyanIrish:
### Description

Fixes #<Issue Number>

### DIFF
--- a/ndcube/ndcube_sequence.py
+++ b/ndcube/ndcube_sequence.py
@@ -1,6 +1,7 @@
 import copy
 import numbers
 import textwrap
+import warnings
 
 import astropy.units as u
 import numpy as np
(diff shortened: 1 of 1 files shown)


ndcube/ndcube_sequence.py
```diff
@@ -239,6 +240,67 @@ def explode_along_axis(self, axis):
...
+        masks = np.zeros(dims, dtype=bool)
+        data_unit = None
+        uncert_unit = None
+        data_unit_present = False
+        data_unit_absent = False
+        uncert_unit_present = False
+        uncert_unit_absent = False
+        for i, cube in enumerate(self._sequence.data):
+            # Extract data values, converting to a consistent unit if the cube includes a unit.
+            if not data_unit and cube.unit:
```
hayesla, line 258: a check should also be added to make sure all the cubes in the sequence have the same size - maybe just checking against the first cube dimension?

ndcube/ndcube_sequence.py
```diff
@@ -239,6 +240,67 @@ def explode_along_axis(self, axis):
         # creating a new sequence with the result_cubes keeping the meta and common axis as axis
         return self._new_instance(result_cubes, common_axis=new_common_axis, meta=self.meta)
 
+    def stack_data(self):
+        """Generates a stacked array from a sequence of cubes."""
```
hayesla, line 244: more docs needed here of what the returned arrays are i.e. Returns : data, unites uncertainties

hayesla (changes requested):

PR #468, opened 2021-08-27
Add Mask class and holding and combining multiple masks

### LINKED ISSUE #257: MaskCollection Class
### Description

Currently, the ```NDCube.mask``` implementation is simply a boolean array.  However I there may be many components to a mask and in some cases users may want to keep them separate, e.g. parts of the CCD not illuminated, dusk in the optical path, cosmic rays hits, and also feature masks, e.g. active region, of flare.

Is there currently a class out there, e.g. ```MaskCollection```, that can store all these components and return a master mask which is a combination of them all a a subset of them.  If not, implementing one might not be so hard.

#### Proposed implementation
* ```MaskCollection``` inherits from ```dict```.
* Mask components are stored as boolean arrays or ```Mask``` objects where:
    * ```Mask.mask``` is a boolean array
    * ```Mask.meta``` is a dictionary-like including a description/info entry giving more info on what the mask represents.
* ```MaskCollection.mask_and(masks)``` is a method returning a logical AND combination of ```masks``` which is a list of ```key``` values of masks within the collection.
* ```MaskCollection.mask_or(masks)```: same as ```MaskCollection.mask_and(masks)``` except that a logical OR combination is returned.
* ```MaskCollection.mask``` is a property returning a combined mask using the conditions defined in ```MaskCollection._mask_defaults```.
* ```MaskCollection._mask_defaults```: a **kwargs-like dictionary defining how the ```MaskCollection.mask``` is computed.
    * ```MaskCollection._mask_defaults.masks```: ```list``` of ```key``` names of masks to be combined.
    * ```MaskCollection._mask_defaults.func()```: The function via which the masks are to be combined.  
        * Must take ```MaskCollection._mask_defaults.masks``` as the input.
        * Can be as simple as ```self.mask_or()``` or a complicated combination combining different masks differently.
        * Default is ```self.mask_or```
* ```MaskCollection.set_mask_defaults(masks=None, func=None)```: a method allowing users to set the defaults in ```MaskCollection._mask_defaults```.

DanRyanIrish:
### Description

This `Mask` class enables multiple masks to be stored which can be combined via the logical OR principle, i.e. if any pixel has a `True` value in any component mask, the value in the combined mask will be `True`.  This facilitates masking an `NDCube` based on different criteria, e.g.:
* pixels outside FOV
* bad data, e.g. cosmic ray hits, dust in optical path, etc.
* regions/features of interest
* etc.

Component masks can be activated and deactivated via `Mask.activate`/`Mask.deactivate`.  This allows users to be more nuanced in their identification of good/bad data and to switch more quickly between different masking criteria, e.g. include/exclude different regions of interest.  Component masks can added and removed via `Mask.add`/`Mask.remove`, and combined via `Mask.mask`.

Inspection methods/properties include `Mask.is_active(name)` (returns `True` is the `name` mask is active, `False` otherwise), `Mask.active_masks` (returns a `tuple` of the names of active masks) and `Mask.shape` (returning the shape if the masks).

Fixes #257

### DIFF
--- a/ndcube/mask.py
+++ b/ndcube/mask.py
@@ -1,0 +1,172 @@
+import copy
+import textwrap
+
+all = ["Mask"]
+
+
+class Mask:
+    """A class for holding and combining boolean mask arrays.
+
+    Each mask can be activated or deactivated and thereby included or excluded
+    from the combined mask. See the activate/deactive methods.
+
+    Parameters
+    ----------
+    masks: dict-like or parseable by `dict`
+        The names and arrays of each mask.
+        Each mask array must be the same shape and of boolean type.
+
+    meta: dict-like
+        Any metadata associated with the masks.
+    """
+    def __init__(self, masks, meta=None):
+        self._masks = dict(masks)
+        shape = self.shape
+        for mask in self._masks.values():
+            if mask.shape != shape:
+                print(mask.shape, shape)
+                raise ValueError("All masks must have same shape.")
+                self._masks[key] = mask.astype(bool)
+        self.meta = meta
+        self._active = dict((key, True) for key in self)
+
+    @property
+    def active_masks(self):
+        """Return the names of active masks"""
+        return tuple(key for key, active in self._active.items() if active)
+
+    @property
+    def mask(self):
+        """Return the combined mask.
+
+        The active masks are compared element-wise. Any element which has a value
+        of True is any active mask is given a value of True in the combined mask.
+        """
+        return sum(self[key] for key in self if self.is_active(key)) > 0
+
+    @property
+    def names(self):
+        """Return the names of all masks, whether active or not."""
+        return tuple(self._masks.keys())
+
+    @property
+    def shape(self):
+        """Return the shape of the masks.
+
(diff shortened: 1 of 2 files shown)


Cadair:
Can open worms everywhere...

We would need to talk to NDData and spectral cube people about this.

DanRyanIrish:
Agreed!  I meant to label this as "Discussion Needed".  Have now done so.

DanRyanIrish:
This can introduced in a non-breaking way.  This PR does not force masks on `NDCube`s to be an instance of this class.  So all other code should be unaffected.  I think the only place it would have to be special-cased in in the plotting.

Alternatively we could require that all masks be a subclass of `Mask` and then the standard API for getting the mask would become `cube.mask.mask`.  Presumably this would be a breaking change for those other packages.

Another option is the make `NDCube.mask` a property and store the mask itself elsewhere, e.g. `NDCube._mask`.  The property could call `cube._mask.mask` if mask is an instance of `Mask`, and `cube._mask` otherwise.  This would require changes to the slicing infrastructure, but not too much.

As you say, worms everywhere.

PR #472, opened 2021-09-17
WIP: Reimplement NDCube.crop to "crop by points" and Extend Support to Crop by ExtraCoords

DanRyanIrish:
### Description

This PR reimplements `NDCube.crop` (and `NDCube.crop_by_values`) to use the "crop by points" algorithm.  Instead of taking lower and upper bounds on the world axes and calculating the corners of the bounding box in pixel space, this PR makes `NDCube.crop` return the smallest cube in pixel space that includes the user-input points described using world coords.  It is not assumed that the entered points are corners.  This is degenerate with the bounding box implementation when the world and pixel axes are aligned.  However, when the coordinate systems are rotated, the returned cube can be much smaller.

This PR also extends the crop functionality to enable cropping by extra coords.  As a result, it introduces a new property, `ExtraCoords.cube_wcs`, which returns a WCS of the extra coords that describes the cube itself.  If not all array axes are associated with an extra coord, dummy world axes are added.

Fixes #<Issue Number>


DanRyanIrish:
Failing tests should be fixes once #482 is merged.

PR #484, opened 2021-10-21
Use new mpl-animators package rather than sunpy

Cadair:


Cadair:
This is pending a release of the `mpl_animators` package to pass the figure-tests workflow.

PR #493, opened 2021-11-01
Add `SkyCoord` frame check to `crop` input validation

dhomeier:
### Description
This is a follow-up to #491 adding an additional check for any `SkyCoord` sections of the bounding box to be compatible with the world coordinates of the cube WCS.
Testing for `is_equivalent_frame` here is the more conservative approach compared to `is_transformable_to`, but I have not found an example of a working transformation between two different frames in the `crop` context, so leaving the implementation up for further discussion.

### DIFF
--- a/ndcube/ndcube.py
+++ b/ndcube/ndcube.py
@@ -14,6 +14,7 @@ try:
     import sunpy.coordinates  # pylint: disable=unused-import  # NOQA
 except ImportError:
     pass
+from astropy.coordinates import SkyCoord
 from astropy.wcs import WCS
 from astropy.wcs.utils import _split_matrix
 from astropy.wcs.wcsapi import BaseHighLevelWCS, HighLevelWCSWrapper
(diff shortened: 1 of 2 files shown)


dhomeier:
@Cadair I have rebased since the issue still exists in main, i.e. trying to crop e.g. a cube with Helioprojective WCS to SkyCoord bounds as in the test raises a `ConvertError` somewhere down the line
```python
gsecoord = <GeocentricSolarEcliptic Coordinate (obstime=None): (lon, lat) in deg
    (-0.00333, -0.00111111)>
heeframe = <HeliocentricEarthEcliptic Frame (obstime=None)>

    @frame_transform_graph.transform(FunctionTransformWithFiniteDifference,
                                     GeocentricSolarEcliptic, HeliocentricEarthEcliptic)
    @_transformation_debug("GSE->HEE")
    def gse_to_hee(gsecoord, heeframe):
        """
        Convert from Geocentric Solar Ecliptic to Heliocentric Earth Ecliptic
        """
        # First transform the GSE coord to the HEE obstime
        int_coord = _transform_obstime(gsecoord, heeframe.obstime)
    
        if int_coord.obstime is None:
>           raise ConvertError("To perform this transformation, the coordinate"
                               " frame needs a specified `obstime`.")
E           astropy.coordinates.errors.ConvertError: To perform this transformation, the coordinate frame needs a specified `obstime`.
```

Issue #496, opened 2021-11-04
Potential bug with `crop_by_values`


hayesla:
### Description

Came across an issue when trying to use `ndcube.NDCube.crop_by_values()` such that it either returns the same NDCube or otherwise errors with `ValueError: Input points causes cube to be cropped to a single pixel. This is not supported.`

It still could be something I'm doing wrong. Also documentation in examples uses should be added to the docs

### Steps to Reproduce

Here's an example, of trying to crop a NDCube which contains the data from an AIA image.

```python
>>> import sunpy.map
>>> from sunpy.data.sample import AIA_171_IMAGE
>>> from ndcube import NDCube
>>> from astropy import units as u
>>> from astropy.coordinates import SkyCoord
>>> from astropy.wcs.utils import wcs_to_celestial_frame


>>> my_map = sunpy.map.Map(AIA_171_IMAGE)
>>> my_ndcube = NDCube(my_map.data, my_map.wcs)

```
and `crop` works:

```python
>>> bl = SkyCoord(500*u.arcsec, -400*u.arcsec, frame=wcs_to_celestial_frame(my_ndcube.wcs))
>>> tr = SkyCoord(1100*u.arcsec, -200*u.arcsec, frame=wcs_to_celestial_frame(my_ndcube.wcs))
>>> my_ndcube.crop(bl, tr).dimensions
[85, 250]pix
```
however when trying to use `crop_by_values` there seems to be a bug

```python
>>> my_ndcube.crop_by_values((500*u.arcsec, -400*u.arcsec), (1100*u.arcsec, -200*u.arcsec)).dimensions
~/opt/miniconda3/envs/july_flare/lib/python3.9/site-packages/ndcube/utils/cube.py in get_crop_item_from_points(points, wcs, crop_by_values)
    195     # If item will result in a scalar cube, raise an error as this is not currently supported.
    196     if result_is_scalar:
--> 197         raise ValueError("Input points causes cube to be cropped to a single pixel. "
    198                          "This is not supported.")
    199     return tuple(item)

ValueError: Input points causes cube to be cropped to a single pixel. This is not supported.
```
 whats even weirder - if I change the bottom left y coord from -400'' to -500'', an error isn't raised, but it returns a NDCube with dimensions of [0,0] 🤷 

```python
>>> my_ndcube.crop_by_values((500*u.arcsec, -500*u.arcsec), (1100*u.arcsec, -200*u.arcsec))

[...]

PR #497, opened 2021-11-08
Converting units to wcs.world_axis_units in crop_by_values

hayesla:
### Description

Following #496 this PR patches an issue with the conversion of quantity input units to the `wcs.world_axis_units` of the ndcube when using `crop_by_values`

### DIFF
--- a/changelog/497.bugfix.rst
+++ b/changelog/497.bugfix.rst
@@ -1,0 +1,1 @@
+Patch to convert quantity objects passed to `crop_by_coords` to the units given in the `wcs.world_axis_units`
--- a/changelog/497.bugfix.rst
+++ b/changelog/497.bugfix.rst
@@ -1,1 +1,1 @@
-Patch to convert quantity objects passed to `crop_by_coords` to the units given in the `wcs.world_axis_units`
+Patch to convert quantity objects passed to ``crop_by_coords`` to the units given in the ``wcs.world_axis_units``.
--- a/ndcube/ndcube.py
+++ b/ndcube/ndcube.py
@@ -7,6 +7,7 @@ from collections.abc import Mapping
 
 import astropy.nddata
 import astropy.units as u
+from astropy.units import UnitsError
 import numpy as np
 
 try:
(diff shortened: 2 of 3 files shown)


hayesla:
need to add a test too

ndcube/ndcube.py
```diff
@@ -548,6 +548,9 @@ def _get_crop_by_values_item(self, *points, units=None, wcs=None):
                     raise ValueError(f"Unit '{unit}' of coordinate object {j} in point {i} is "
                                      f"incompatible with WCS unit '{wcs.world_axis_units[j]}'")
```
DanRyanIrish, line 549: Lines 545 - 549 can be removed now that the unit will be checked implicitly by the conversion.

DanRyanIrish:
> need to add a test too

Also needs a changelog and the codestyle to be fixed.

ndcube/ndcube.py
hayesla: well I guess this checks they are compatible? what if you passed `km` for example or something else that isnt for example convertible to `u.deg`

hayesla: but I could tidy this up to check before converting

DanRyanIrish: The unit conversion would error and tell you that km is not compatible with deg.

If you wanted to try to give a more helpful error message you could do
```python
if value is not None:
    try:
        points[i][j] = points[i][j].to(wcs.world_axis_units[j])
    except UnitsError:
        raise UnitsError(f"Unit '{unit}' of coordinate object {j} in point {i} is "
                         f"incompatible with WCS unit '{wcs.world_axis_units[j]}'"
```
You'd have to import `UnitsError` from `astropy.units`.

ndcube/ndcube.py
```diff
@@ -543,10 +544,11 @@ def _get_crop_by_values_item(self, *points, units=None, wcs=None):
                             f"index: {i}; coord type: {type(value)}; unit: {unit}")
                     points[i][j] = u.Quantity(value, unit=unit)
                 elif value is not None:
```
DanRyanIrish, line 546:
```suggestion
                if value is not None:
```
This conversion needs to be done even if `value_is_float` as there's no guarantee that the user will enter the values in the `wcs_axis_units`.

DanRyanIrish (changes requested):
Also needs the codestyle to be fixed.

hayesla:
ok - I think that final test that's failing isn't related to this PR right?

DanRyanIrish:
> ok - I think that final test that's failing isn't related to this PR right?

No, that's not you.

changelog/497.bugfix.rst
```diff
@@ -0,0 +1 @@
+Patch to convert quantity objects passed to `crop_by_coords` to the units given in the `wcs.world_axis_units`
```
Cadair, line 1:
```suggestion
Patch to convert quantity objects passed to ``crop_by_coords`` to the units given in the ``wcs.world_axis_units``.
```

DanRyanIrish (approved):

PR #498, opened 2021-11-10
Fix the use of `axes_units=` when using the NDCube animator

SolarDrew:
### Description

`world_axis_physical types` is undefined. It breaks things.


DanRyanIrish:
Thanks @SolarDrew.  This looks like the right fix.  Can you add a changelog and a test that executes that code?

Cadair (approved):

DanRyanIrish (approved):

SolarDrew:
As usual, not before Stuart does :P

PR #499, opened 2021-11-10
Write Model for Length-1 Lookup Tables

DanRyanIrish:
### Description

Fixes #

### DIFF
--- a/ndcube/extra_coords/table_coord.py
+++ b/ndcube/extra_coords/table_coord.py
@@ -16,6 +16,14 @@ from astropy.wcs.wcsapi.wrappers.sliced_wcs import combine_slices, sanitize_slic
 __all__ = ['TimeTableCoordinate', 'SkyCoordTableCoordinate', 'QuantityTableCoordinate']
 
 
+@models.custom_model
+def length1_lookup_table(x, lookup_table, fill_value=np.nan):
+    if x.to_value(u.pix) == 0:
+        return lookup_table[0]
+    else:
+        return fill_value
+
+
 def _generate_generic_frame(naxes, unit, names=None, physical_types=None):
     """
     Generate a simple frame, where all axes have the same type and unit.
@@ -56,18 +64,21 @@ def _generate_tabular(lookup_table, interpolation='linear', points_unit=u.pix, *
     if len(points) == 1:
         points = points[0]
 
-    kwargs = {
-        'bounds_error': False,
-        'fill_value': np.nan,
-        'method': interpolation,
-        **kwargs
-    }
+    if len(lookup_table) == 1:
+        t = length1_lookup_table(lookup_table, fill_value=kwargs.get("fill_value", np.nan))
+    else:
+        kwargs = {
+            'bounds_error': False,
+            'fill_value': np.nan,
+            'method': interpolation,
+            **kwargs
+        }
 
-    t = TabularND(points, lookup_table, **kwargs)
+        t = TabularND(points, lookup_table, **kwargs)
 
-    # TODO: Remove this when there is a new gWCS release
-    # Work around https://github.com/spacetelescope/gwcs/pull/331
-    t.bounding_box = None
+        # TODO: Remove this when there is a new gWCS release
+        # Work around https://github.com/spacetelescope/gwcs/pull/331
+        t.bounding_box = None
 
     return t
 
--- a/ndcube/extra_coords/table_coord.py
+++ b/ndcube/extra_coords/table_coord.py
@@ -10,45 +10,83 @@ import numpy as np
 from astropy.coordinates import SkyCoord
 from astropy.modeling import models
 from astropy.modeling.models import tabular_model
+from astropy.modeling.tabular import _Tabular
 from astropy.time import Time
 from astropy.wcs.wcsapi.wrappers.sliced_wcs import combine_slices, sanitize_slices
 
 __all__ = ['TimeTableCoordinate', 'SkyCoordTableCoordinate', 'QuantityTableCoordinate']
 
 
-@models.custom_model
-def length1_lookup_table(x, lookup_table=[0]*u.pix, fill_value=np.nan):
-    """Generate a length-1 lookup table model.
+class _Length1Tabular(_Tabular):
+    def __init__(self, points=None, lookup_table=None, pixel_width=None,
+                 method='linear', bounds_error=True, fill_value=np.nan, **kwargs):
+        if len(lookup_table) != 1:
+            raise ValueError("lookup_table must have length 1.")
+        super().__init__(points=points, lookup_table=lookup_table, method=method,
+                         bounds_error=bounds_error, fill_value=fill_value, **kwargs)
+        self._pixel_width = pixel_width  # Width of pixel in world units.
 
-    If the requested pixel coordinate corresponds to the 0th pixel,
-    the 0th element of the table is returned. Otherwise the fill value
-    is returned.
+    def evaluate(self, x):
+        output = np.full(x.shape, self.fill_value)
+        output[np.logical_and(x >= -0.5, x < 0.5)] = self.lookup_table[0].value
+        return output * self.lookup_table.unit
 
-    Parameters
-    ----------
-    x: `astropy.units.Quantity` in pixel units.
-        The pixel coordinate(s) for which the value of the lookup is desired.
-    lookup_table: `astropy.units.Quantity`
-        The world value corresponding to the 0th pixel.
-        Must be length-1, not scalar.
-    fill_value: `float`
-        The value returned for pixel indices outside the 0th pixel
-    """
-    if not isinstance(x, u.Quantity):
-        raise TypeError("x must be an astropy Quantity with pixel units.")
-    if not (isinstance(lookup_table, u.Quantity) and lookup_table.shape == (1,)):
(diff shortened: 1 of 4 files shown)


DanRyanIrish:
Something like this @Cadair?

ndcube/extra_coords/table_coord.py
```diff
@@ -16,6 +16,14 @@
 __all__ = ['TimeTableCoordinate', 'SkyCoordTableCoordinate', 'QuantityTableCoordinate']
 
 
+@models.custom_model
+def length1_lookup_table(x, lookup_table, fill_value=np.nan):
+    if x.to_value(u.pix) == 0:
```
Cadair, line 21: This is too narrow. A pixel has a width!

ndcube/extra_coords/table_coord.py
```diff
@@ -56,18 +64,21 @@ def _generate_tabular(lookup_table, interpolation='linear', points_unit=u.pix, *
...
+        kwargs = {
+            'bounds_error': False,
+            'fill_value': np.nan,
+            'method': interpolation,
+            **kwargs
+        }
+
+        t = TabularND(points, lookup_table, **kwargs)
+
+        # TODO: Remove this when there is a new gWCS release
```
Cadair, line 79: Is this done now?

ndcube/extra_coords/table_coord.py
DanRyanIrish: So it should be this?
```suggestion
    x_is_scalar = False
    if x.isscalar():
        x_is_scalar = True
        x.reshape((1,))
    output = np.full(x.shape, fill_value)
    x_value = x.to_value(u.pix)
    output[np.logical_or(x_value >= -0.5, x_value < 0.5)] = lookup_table[0].value
    output *= lookup_table[0].unit
    if x_is_scalar:
        return output[0]
    return output
```

ndcube/extra_coords/table_coord.py
DanRyanIrish: Looks like it was included in the 0.16.0 gWCS release.  If we remove this now, we'll have to pin ndcube 2.0.1 to that gWCS version, right?

Cadair: Given 0.16 is nearly a year old, I think we could do that?

ndcube/extra_coords/table_coord.py
Cadair: I think that would work, would need some tests just for the model I think.

DanRyanIrish:
@Cadair does this need a test for anything else?  Or do you think this is good to go now?

Cadair:
I would say we should at least give this a run through ExtraCoords i.e. make a long table, slice the ec object and then test pixel_to_world and axis_world_coords etc

DanRyanIrish:
@Cadair this is ready for another review.  I don't know why the test is failing on the oldest deps.  Have you any idea?

nabobalis (approved):
Fixes the issue on sunraster

ndcube/extra_coords/table_coord.py
```diff
@@ -10,12 +10,85 @@
...
+        _Tabular._id += 1
+        name = f'{name_root}{model_id}'
+
+    model_class = type(str(name), (model,), members)
+    model_class.__module__ = 'ndcube.extra_coords.table_coord'
+    return model_class
+
+
+Length1Tablular = length1_tabular_model(name="Length1Tabular")
+InverseLength1Tabular = length1_tabular_model(name="InverseLength1Tabular", inverse=True)
```
Cadair, line 89: We don't need this function. In Astropy it's there to dynamically generate tables for arbitary numbers of dimensions. We can just directly use the two classes you defined.

```suggestion
```

ndcube/extra_coords/table_coord.py
```diff
@@ -10,12 +10,85 @@
...
+        return output * self.lookup_table.unit
+
+    @property
+    def inverse(self):
+        return InverseLength1Tabular(points=self.points[0], lookup_table=self.lookup_table,
+                                     pixel_width=self._pixel_width, method=self.method,
+                                     bounds_error=self.bounds_error, fill_value=self.fill_value)
+
+
+class _InverseLength1Tabular(_Length1Tabular):
```
Cadair, line 41:
```suggestion
class InverseLength1Tabular(_Length1Tabular):
```

ndcube/extra_coords/table_coord.py
```diff
@@ -10,12 +10,85 @@
...
 from astropy.modeling import models
 from astropy.modeling.models import tabular_model
+from astropy.modeling.tabular import _Tabular
 from astropy.time import Time
 from astropy.wcs.wcsapi.wrappers.sliced_wcs import combine_slices, sanitize_slices
 
 __all__ = ['TimeTableCoordinate', 'SkyCoordTableCoordinate', 'QuantityTableCoordinate']
 
 
+class _Length1Tabular(_Tabular):
```
Cadair, line 20:
```suggestion
class Length1Tabular(_Tabular):
```

ndcube/extra_coords/table_coord.py
```diff
@@ -10,12 +10,71 @@
...
+        if len(lookup_table) != 1:
+            raise ValueError("lookup_table must have length 1.")
+        super().__init__(points=points, lookup_table=lookup_table, method=method,
+                         bounds_error=bounds_error, fill_value=fill_value, **kwargs)
+        self._pixel_width = pixel_width  # Width of pixel in world units.
+
+    def evaluate(self, x):
+        output = np.full(x.shape, self.fill_value)
+        x_unit = self.input_units['x']
+        output[np.logical_and(x >= -0.5*x_unit, x < 0.5*x_unit)] = self.lookup_table[0].value
```
Cadair, line 42: Should this not use `pixel_width`?

ndcube/extra_coords/table_coord.py
```diff
@@ -10,12 +10,71 @@
...
+                                     bounds_error=self.bounds_error, fill_value=self.fill_value)
+
+
+class InverseLength1Tabular(Length1Tabular):
+    def __init__(self, **kwargs):
+        points = kwargs.pop("points", None)
+        lookup_table = kwargs.pop("lookup_table", None)
+        super().__init__(points=lookup_table, lookup_table=points, **kwargs)
+        if self._pixel_width is None:
+            self._pixel_width = 0 * lookup_table.unit
```
Cadair, line 58: Defaulting to 0 here? We aren't overriding this anywhere?

ndcube/extra_coords/table_coord.py
```diff
@@ -10,12 +10,71 @@
...
+    _has_inverse_bounding_box = True
+    _separable = True
+
+    n_inputs = 1
+    n_outputs = 1
+
+    lookup_table = np.zeros([1])
+    points = np.zeros([1])
+
+    def __init__(self, points=None, lookup_table=None, pixel_width=None,
```
Cadair, line 31: `pixel_width` is rather specific given the rest of this model allows anything in the points and lookup_table. `point_width` might be more generic?

ndcube/extra_coords/table_coord.py
DanRyanIrish: No, `pixel_width` is effectively CDELT.  So it's in real world coordinates and needs to be defined by the user.  Here we need the pixel width in pixel units which is always defined as -0.5 - 0.5 pix.

ndcube/extra_coords/table_coord.py
DanRyanIrish: I'm open to the name change.  To clarify though, `pixel_width` is effectively CDELT, i.e it's in real world coords.  Is there a better name than `pixel_width` or `point_width` given this?

ndcube/extra_coords/table_coord.py
DanRyanIrish: It's given by the `pixel_width` kwarg, defined in the init of `Length1Tabular` which this class inherits.

Cadair: but I mean we aren't populating it in our code anywhere, so it's always going to use the default?

ndcube/extra_coords/table_coord.py
Cadair: ah, I see. The question here really becomes do we want these models to be only for a pixel <> world transform? Because at the moment, by virtue of inheriting `_Tabular` it isn't restricted to that.

I would say `value_width` would be better as you have values inside the lookup_table and points which are mapped to those values.

ndcube/extra_coords/table_coord.py
Cadair: Should this also have it's own parameter then? to match what is currently `pixel_width`, we are just assuming the point has a width of 1, it might not (in the general sense).

ndcube/extra_coords/table_coord.py
DanRyanIrish: Unless the user sets the `pixel_width` kwarg, yes.  But that's how default values work no?

DanRyanIrish:
@Cadair: I've replaced `pixel_width` `point_width` and `value_width` as you suggested.  I also used that change to simplify `InverseLength1Tabular` so now most of the code is in `Length1Tabular`.

PR #500, opened 2021-11-11
Update CI and min dependancies

Cadair:
:thinking:


setup.cfg
```diff
@@ -34,9 +34,9 @@ python_requires = >=3.7
 setup_requires =
   setuptools_scm
 install_requires =
-  astropy>=4.1
+  astropy>=4.2
```
Cadair, line 37: Without this change, any dropped dimensions were not added to global coords.

setup.cfg
```diff
@@ -34,9 +34,9 @@ python_requires = >=3.7
...
   gwcs>=0.15
```
DanRyanIrish, line 38:
```suggestion
  gwcs>=0.16
```

tox.ini
```diff
@@ -55,9 +55,18 @@ deps =
     online: pytest-rerunfailures
     online: pytest-timeout
 
+    # Oldest Dependacies
```
DanRyanIrish, line 58:
```suggestion
    # Oldest Dependancies
```

tox.ini
```diff
@@ -55,9 +55,18 @@ deps =
...
+    oldestdeps: sunpy<3.1
+    oldestdeps: astropy<4.3
+    oldestdeps: gwcs<0.16
+    oldestdeps: numpy<1.18
+    oldestdeps: matplotlib<3.3
+    oldestdeps: mpl_animators<1.1
```
DanRyanIrish, line 64: Should these be `==` or `>`?

tox.ini
```diff
@@ -55,9 +55,18 @@ deps =
...
+    oldestdeps: reproject==0.7.1
+
     # Figure tests need a tightly controlled environment
     figure: matplotlib==3.3.2
-    figure-!devdeps: astropy==4.1.0
+    figure-!devdeps: astropy==4.2.0
```
DanRyanIrish, line 69: Will this fix the failing fig tests?

changelog/500.bugfix.rst
```diff
@@ -0,0 +1,2 @@
+Bump the minimum version of astropy to 4.2 to correctly support capturing
+dropped world dimensions into global coords when slicing the WCS.
```
DanRyanIrish, line 2: Mention gwcs 0.16 too?

setup.cfg
Cadair: Thinking about this a bit more, we should do it in a separate PR for 2.1 rather than a bug fix.

tox.ini
Cadair: yes

tox.ini
Cadair: This is a little odd, but it's done with `<` so that it picks up whatever the newest bugfix release of the minimum supported minor version is. i.e. `<3.1` will pick up `3.0.2` when it's released without us having to change anything.

DanRyanIrish (approved):

PR #534, opened 2022-08-18
Add functools call to copy `plotter.plot` doc and signature to `plot`

### LINKED ISSUE #530: The docstring / function signature for `NDCube.plot` is useless
### Describe the bug

```
help(mycube.plot)
```

isn't much use because all the actual code is in `mycube.plotter.plot`, we should dynamically copy the signature and docstring.

SolarDrew:
Closes #530

## PR Description

Dynamically copies the docstring and function signature from `NDCube.plotter.plot()` to `my_ndcube_instance.plot()` on instantiation. I'm not sure if this is the best way to achieve this or if it will work for different plotters, but it's the only approach I've found so far that works because a) you can't write a method's `__doc__` once it's been bound to an instance, and b) having `self.plotter` not defined until instantiation makes things weird and difficult.


ndcube/ndcube.py
```diff
@@ -230,6 +232,8 @@ def __init__(self, data, wcs=None, uncertainty=None, mask=None, meta=None,
                 global_coords = deepcopy(global_coords)
             self._global_coords = global_coords
 
+        functools.update_wrapper(NDCube.plot, self.plotter.plot)
```
Cadair, line 235: My concern with the `NDCube.plot` part here is that this applies globally to *all* ndcube instances, so if you had two instances with different plotters it would pick the docstring for both of the one that got instantiated last?

SolarDrew: Yes, I worried about that as well and I've pushed a better version.

Cadair:
pre-commit.ci autofix

Cadair:
Is there a way we can test this? Maybe a call to `help()`?

SolarDrew:
Yeah, that should work, we can run it with each subclass and make sure it gets the right docstring.

DanRyanIrish:
Barring an official approval, is this PR ready to go @SolarDrew @Cadair

SolarDrew:
Nearly, I still need to implement the test call to `help()` but assuming the rest passes then once that's done it's ready.

DanRyanIrish:
Probably a silly question.  What's wrong with the docstring of `.plot` just saying something like what it say now?  Maybe more explicitly directing people to the docstring of `NDCube.plotter.plot`?

Cadair:
If you try and figure out what arguments you can pass to plot etc then you can't and you have to go look at a different thing. I found this annoying in the context of being in an interactive session as opposed to in the docs.

ndcube/tests/test_ndcube.py
```diff
@@ -827,3 +827,9 @@ def test_reproject_exact_incompatible_wcs(ndcube_4d_ln_l_t_lt, wcs_4d_lt_t_l_ln,
     with pytest.raises(ValueError):
         _ = ndcube_4d_ln_l_t_lt.reproject_to(wcs_4d_lt_t_l_ln, algorithm='exact',
                                              shape_out=(5, 10, 12, 8))
+
+
+def test_plot_docstring():
+    cube = NDCube([], astropy.wcs.WCS())
+
+    assert cube.plot.__doc__ == cube.plotter.plot.__doc__
```
Cadair, line 835: You could use this: https://docs.python.org/3/library/inspect.html#inspect.signature to compare the signature as well?

SolarDrew: Ah yes, good call.

SolarDrew: So actually it turns out the signatures aren't the same anyway, so this won't work right now. From what I remember of figuring out how to do this in the first place I think the signature was slightly more fiddly to copy accross, and I'm not sure how much difference it makes to the user if the docstring is there. So I guess it's up to @DanRyanIrish whether we want to incur a short delay while I fix that or merge it as it and open a new issue for the signature to come back to.

Cadair: Some of the signature did get copied:
```
In [8]: inspect.signature(ds.plotter.plot)
Out[8]: <Signature (axes=None, plot_axes=None, axes_coordinates=None, axes_units=None, data_unit=None, wcs=None, **kwargs)>

In [9]: inspect.signature(ds.plot)
Out[9]: <Signature (plot_axes=None, axes_coordinates=None, axes_units=None, data_unit=None, wcs=None, **kwargs)>
```
Not entirely sure where the bonus `axes=` comes from though.

ndcube/visualization/descriptor.py
```diff
@@ -18,35 +20,76 @@ def __set_name__(self, owner, name):
...
+        # If we have no default type then just return None
+        else:
+            return
+
+    def __get__(self, obj, objtype=None):
+        if obj is None:
+            return
+
+        if getattr(obj, self._attribute_name, None) is None:
+            self._resolve_default_type(obj)
```
Cadair, line 77: We don't need both `_resolve_type` and `_resolve_default_type` we should make this use the current `_resolve_type` implementation (although I think `_resolve_default_type` is a better name).

ndcube/visualization/descriptor.py
```diff
@@ -18,35 +21,60 @@ def __set_name__(self, owner, name):
...
-                else:
-                    # If we have no default type then just return None
-                    return
+                    return MatplotlibSequencePlotter
             except ImportError as e:
-                raise ImportError(MISSING_MATPLOTLIB_ERROR_MSG) from e
+                if raise_error:
+                    raise ImportError(MISSING_MATPLOTLIB_ERROR_MSG) from e
+                else:
+                    warnings.warn(MISSING_MATPLOTLIB_ERROR_MSG)
```
Cadair, line 44: I wouldn't even bother doing this here. I would just pass silently, we will error if and when they call `.plot()` or access `.plotter`.

ndcube/visualization/descriptor.py
```diff
@@ -1,3 +1,6 @@
+import functools
+import warnings
```
Cadair, line 2:
```suggestion
```

Cadair (approved):

Cadair:
pre-commit.ci autofix

Issue #535, opened 2022-08-24
Cannot update NDCollection when aligned_axes are not set


jmbhughes:
### Describe the bug

If you are not using `aligned_axes` for an `NDCollection` and you try to `update` the collection with a new collection, you get an error. It should be possible if both `NDCollection`s are not using `aligned_axes`. 

### To Reproduce

```py
from ndcube import NDCollection, NDCube
import numpy as np
import astropy.wcs
import astropy.units as u
import astropy.wcs
import numpy as np

def generate_test_cube():
    linewidth_data = np.random.rand(4, 4) / 2 # dummy data
    linewidth_wcs = astropy.wcs.WCS(naxis=2)
    linewidth_wcs.wcs.ctype = 'HPLT-TAN', 'HPLN-TAN'
    linewidth_wcs.wcs.cunit = 'deg', 'deg'
    linewidth_wcs.wcs.cdelt = 0.5, 0.4
    linewidth_wcs.wcs.crpix = 2, 2
    linewidth_wcs.wcs.crval = 0.5, 1
    linewidth_cube = NDCube(linewidth_data, linewidth_wcs)
    return linewidth_cube

my_collection = NDCollection([("one", generate_test_cube()), ("two", generate_test_cube())])
other_collection = NDCollection([("three", generate_test_cube())])

my_collection.update(other_collection)
```
### Screenshots

<img width="905" alt="Screen Shot 2022-08-24 at 8 54 55 AM" src="https://user-images.githubusercontent.com/13068381/186451210-e84f243e-4f54-44a3-ba02-93a079e425bd.png">

### System Details

ndcube version 2.0.2

SunPy Installation Information:

General
#######
OS: Mac OS 10.16
Arch : 64bit, (i386)
SunPy : 3.0.1

Required Dependices
###################
astropy: 4.2.1
parfive: 1.3.0
numpy: 1.20.1

Optional Dependices
###################
asdf: 2.10.0
beautifulsoup4: 4.9.3
dask: 2022.2.1
drms: 0.6.2
glymur: Missing, need glymur!=0.9.0,>=0.8.18; extra == "all"
h5netcdf: Missing, need h5netcdf>=0.8.1; extra == "all"
h5py: Missing, need h5py>=3.1.0; extra == "all"
matplotlib: 3.3.4
pandas: 1.2.4
python-dateutil: 2.8.1
scikit-image: 0.18.1
scipy: 1.6.2
sqlalchemy: 1.4.36
[...]

PR #538, opened 2022-08-26
corrects bug 535

jmbhughes:
## PR Description

Bug #535 identifies that an NDCollection without aligned_axes cannot be updated. This was because the assertion for compatible axes did not consider the scenario where aligned_axes were not being used. Thus, I updated the `ncube.utils.colleciton.assert_aligned_axes_compatible` and the `ndcube.ndcollection.NDCollection.update` methods to take this into account by checking for this explicitly. Tests were added for this and passed successfully.

### DIFF
--- a/ndcube/ndcollection.py
+++ b/ndcube/ndcollection.py
@@ -258,15 +258,19 @@ class NDCollection(dict):
             new_data = list(collection.values())
             key_data_pairs = zip(new_keys, new_data)
             new_aligned_axes = collection.aligned_axes
+
         # Check aligned axes of new inputs are compatible with those in self.
         # As they've already been sanitized, only one set of aligned axes need be checked.
+        first_old_aligned_axes = self.aligned_axes[self._first_key] if self.aligned_axes is not None else None
+        first_new_aligned_axes = new_aligned_axes[new_keys[0]] if new_aligned_axes is not None else None
         collection_utils.assert_aligned_axes_compatible(
             self[self._first_key].dimensions, new_data[0].dimensions,
-            self.aligned_axes[self._first_key], new_aligned_axes[new_keys[0]]
+            first_old_aligned_axes, first_new_aligned_axes
         )
         # Update collection
         super().update(key_data_pairs)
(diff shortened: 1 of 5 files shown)


ndcube/ndcollection.py
```diff
@@ -258,15 +258,19 @@ def update(self, *args):
...
+        first_new_aligned_axes = new_aligned_axes[new_keys[0]] if new_aligned_axes is not None else None
         collection_utils.assert_aligned_axes_compatible(
             self[self._first_key].dimensions, new_data[0].dimensions,
-            self.aligned_axes[self._first_key], new_aligned_axes[new_keys[0]]
+            first_old_aligned_axes, first_new_aligned_axes
         )
         # Update collection
         super().update(key_data_pairs)
-        self.aligned_axes.update(new_aligned_axes)
+        if first_old_aligned_axes is not None:  # since the above assertion passed, both are aligned_axes are not None
```
DanRyanIrish, line 272:
```suggestion
        if first_old_aligned_axes is not None:  # since the above assertion passed, if one aligned axes is not None, both are not None
```

ndcube/utils/collection.py
```diff
@@ -150,10 +150,16 @@ def assert_aligned_axes_compatible(data_dimensions1, data_dimensions2, data_axes
...
 
     """
-    # Confirm same number of aligned axes.
-    if len(data_axes1) != len(data_axes2):
-        raise ValueError("Number of aligned axes must be equal: "
-                         f"{len(data_axes1)} != {len(data_axes2)}")
-    # Confirm dimension lengths of each aligned axis is the same.
-    if not all(data_dimensions1[np.array(data_axes1)] == data_dimensions2[np.array(data_axes2)]):
-        raise ValueError("All corresponding aligned axes between cubes must be of same length.")
+    # there is a mismatch where one collection uses aligned_axes and the other does not
```
DanRyanIrish, line 153:
```suggestion
    # If one set of aligned axes is None and the other isn't, they are defined as not compatible.
```

PR #539, opened 2022-08-26
aligned_axis_physical_types returns None when no aligned_axes

jmbhughes:
## PR Description

Danny pointed out that `__str__` will error out when `aligned_axes` is `None` because `aligned_axis_physical_types` would raise a `ValueError`. Thus, I changed so that instead of raising a `ValueError` in this scenario `aligned_axis_physical_types` will now just return `None`.


DanRyanIrish:
Looks great.  A changelog is needed

DanRyanIrish (approved):

PR #450, opened 2021-07-29
Implement Superpixel on NDCube

DanRyanIrish:
### Description

This PR implements a `superpixel` method on `NDCube` (now renamed `rebin`) that downsamples the grid by a factor that is an integer fraction of the cube's length in the relevant dimension.  It enables the user to provide the function that dictates how the value in the superpixel is derived from its constituent pixels.  To support this, an `interpolate` method has been introduced to `ExtraCoords`.

For reviewing purposes, the core pieces of this PR are `NDCube.rebin` and `ExtraCoords.interpolate`.

This PR depends on #449 

### To Do
- [x] Discuss/decide how to handle uncertainty
- [x] Handle ExtraCoords
- [x] Handle SkyCoord ExtraCoord
- [x] Add tests
- [x] Add changelog for ExtraCoord.interpolate
- [x] Test error propagation
- [x] Generalise SkyCoord interpolation as suggested in conversation thread.
- [x] Use numpy.interp to remove scipy dependency
- [x] Rename `superpixel` to `rebin`
- [x] Propagate uncertainties to `max` and `min` versions. If multiple pixels with max/min value, preserve the larger uncertainty
- [x] Add and standard deviation version.  Drop uncertainty
- [x] Write tests for min/max versions
- [x] Resolve near-name conflict with astropy `NDArithmeticMixin` kwarg `propagate_uncertainties` which has a different behaviour.
- [x] Implement extra_coords interpolate on TableCoordinate objects
- [x] Add resample method to ExtraCoords leveraging TableCoord interpolate
- [x] Replace use of ExtraCoords.interpolate with ExtraCoords.resample in this PR and bugfix ExtraCoords.resample
- [x] Delete ExtraCoords.interpolate
- [x] Change `use_masked_values` to `operation_ignores_mask`

PR #541, opened 2022-08-26
Add basic arithmetic methods

wtbarnes:
This PR adds basic support for arithmetic methods on NDCube: add, subtract, multiply, divide. It explicitly does not support operations between NDCube objects (or more precisely anything that is an NDData instance). 

There was also some discussion about to what extent we can support operations with Dask arrays. There is no issue if our NDCube is already backed by a Dask array. However, if the other object that we are performing the operation with is a Dask array, it will be eagerly evaluated because of the use of `Quantity`. This is probably not necessary and we should be checking possible other operands in a slightly different way.

### TODOs

- [x] Fix the tests
- [ ] ~~Gallery example~~
- [x] Changelog 
- [x] Create issues for supporting pow and rtruediv
- [x] Create "downstream" issues on sunpy
- [x] Test for `NotImplementedError`
- [x] Scale uncertainties as part of multiplication


DanRyanIrish:
@wtbarnes: Summarising our discussion on the sunpy call:
* Current implementation of casting input to a `Quantity` forces input to be read into memory and not supportive of `dask`.
* Therefore, we should start by only supporting Quantities and non-unit objects, e.g. floats, arrays, etc.  So the new implementation should be as follows:
  * Check if input has a unit using `hasatrr(input, "unit")`
  * If not, apply operation assuming input is a zero-uncertainty unitless scalar/array
  * If it does have a unit, check if input is a `Quantity` using `isinstance`.
  * If not, raise `NotImplementedError`
  * If so, confirm the unit is compatiable with the `NDCube`'s unit.  If not, raise Error.
  * Work out scaling factor between units.
  * Apply operation to data with input value * scaling factor.  This should mean that the resulting data is still in the units of the original `NDCube`.
* In the future we can work out ways of being less restrictive the types we support.  But this should account for many users' needs without painting ourselves into a corner and so if worth implementing.

Feel free to push back if you feel this doesn't represent what we agreed.

DanRyanIrish:
Any movement or planned movement on this @wtbarnes?

wtbarnes:
Sorry I have not had any time to revisit this since the discussion you summarized above. I should have some time this week or next to make the refactor suggested in the above comment.

wtbarnes:
The logic here is not very clean and could probably be refactored a bit to avoid some repetition, but I believe this accomplishes what is laid out above. Currently, I'm not explicitly excluding NDCube or NDData instances, but they're implicitly caught by the "has-unit-but-is-not-a-quantity" case.

changelog/541.feature.rst
```diff
@@ -0,0 +1,3 @@
+Add basic arithmetic methods between `~ndcube.NDCube` objects and broadcastable arrays
+and `~astropy.unit.Quantity` objects. Operations between two `~ndcube.NDCube` objects
```
DanRyanIrish, line 2: And scalars, right?

ndcube/ndcube.py
```diff
@@ -792,3 +801,90 @@ def plot(self, *args, **kwargs):
...
+            new_unit = self.unit
+        return self._new_instance_from_op(new_data, new_unit)
+
+    def __rmul__(self, value):
+        return self.__mul__(value)
+
+    def __truediv__(self, value):
+        return self.__mul__(1/value)
+
+    def __rtruediv__(self, value):
```
DanRyanIrish, line 874: Remind me why this can't also be ```return self.__truediv__(value)```.  Will all this checking and handling then ultimately be done by ```self.__mul__```

ndcube/ndcube.py
```diff
@@ -792,3 +801,90 @@ def plot(self, *args, **kwargs):
...
+                              meta=self.meta,
+                              uncertainty=self.uncertainty)
+        if self.extra_coords is not None:
+            new_cube._extra_coords = deepcopy(self.extra_coords)
+        return new_cube
+
+    def __neg__(self):
+        return self._new_instance_from_op(-self.data, self.unit)
+
+    def __pow__(self, value):
```
DanRyanIrish, line 821: I think this needs to propagate uncertainties.  See #450 for my current attempt on how to do this.  If this is too much to agree on by Friday, I suggest not implementing `__pow__` until we've agreed an implementation for uncertainty propagation.

ndcube/tests/test_ndcube.py
```diff
@@ -827,3 +827,133 @@ def test_reproject_exact_incompatible_wcs(ndcube_4d_ln_l_t_lt, wcs_4d_lt_t_l_ln,
...
+    cube_quantity = u.Quantity(ndcube_2d_ln_lt_units.data, ndcube_2d_ln_lt_units.unit)
+    new_cube = ndcube_2d_ln_lt_units ** 2
+    check_arithmetic_value_and_units(new_cube, cube_quantity ** 2)
+
+
+def test_map_arithmetic_neg(ndcube_2d_ln_lt_units):
+    check_arithmetic_value_and_units(
+        -ndcube_2d_ln_lt_units,
+        u.Quantity(-ndcube_2d_ln_lt_units.data, ndcube_2d_ln_lt_units.unit),
+    )
```
DanRyanIrish, line 959: Should include tests ensuring `NotImplementedError` is raised when it should be.

ndcube/ndcube.py
```diff
@@ -792,3 +801,90 @@ def plot(self, *args, **kwargs):
                 "no default plotting functionality is available.")
 
         return self.plotter.plot(*args, **kwargs)
+
+    def _new_instance_from_op(self, new_data, new_unit):
```
DanRyanIrish, line 805: This also needs to carry over `global_coords`

ndcube/ndcube.py
```diff
@@ -792,3 +801,90 @@ def plot(self, *args, **kwargs):
...
+
+    def __mul__(self, value):
+        if hasattr(value, 'unit'):
+            if isinstance(value, u.Quantity):
+                # NOTE: if the cube does not have units, set the unit
+                # to dimensionless such that we can perform arithmetic
+                # between the two.
+                cube_unit = u.Unit('') if self.unit is None else self.unit
+                new_data = self.data * value.to_value()
+                new_unit = cube_unit * value.unit
```
DanRyanIrish, line 860: I think this highlights that we probably want an `NDCube.to` method as users will likely want normalize data e.g. with units of `u.m / u.cm`.  Further use cases include converting between spectral units.  However, this is probably beyond the scope of this PR.

ndcube/ndcube.py
```diff
@@ -756,6 +756,15 @@ class NDCube(NDCubeBase):
         Default is False.
 
     """
+    # Enabling the NDCube reflected operators is a bit subtle.  The NDCube
```
DanRyanIrish, line 759: My feeling is that we probably want all these arithmetic methods on `NDCubeBase` rather than `NDCube`.  That way `NDCube` will inherit them.  Are there any arguments against? @wtbarnes @Cadair ?

DanRyanIrish (commented):
This is looking great @wtbarnes.  I haven't reviewed the tests but have left a few comments and one discussion point (about `NDCube` vs `NDCubeBase`).  This PR isn't the end of the story but is a big leap forward and I believe isn't far off now.

changelog/541.feature.rst
wtbarnes: I was grouping scalars with "broadcastable arrays" but yes I can note that explicitly.

ndcube/ndcube.py
wtbarnes: `__truediv__` is `cube / value` while `__rtruediv__` is `value/cube` so the implicit assumption in `__mul__` is that `cube` is in the numerator. This is not true here. Actually though, this could be implemented as `__pow__(-1).__mul__(value)__`...

ndcube/ndcube.py
wtbarnes: Oh yes, good point! I think at this stage, we should just explicitly check that `self.uncertainties is None` and if not, throw a `NotImplemented`.

ndcube/tests/test_ndcube.py
wtbarnes: Yes, good point. I will add those.

ndcube/ndcube.py
wtbarnes: I will add that in. Is it a kwarg or does it need to be attached after instantiating like `extra_coords`?

ndcube/ndcube.py
wtbarnes: I'm in favor of this, but I agree it is out of scope here.

ndcube/ndcube.py
wtbarnes: I think we briefly talked about this at the coordination meeting and I *thought* the consensus was that it should go on `NDCube` rather than `NDCubeBase` but perhaps I am mistaken.

ndcube/ndcube.py
DanRyanIrish: I actually disagree (in a friendly tone :) ).  I think if we give people something that returns `NotImplementedError` in some cases it may be confusing and frustrating.  Then once we do implement it fully, it may take some time and effort to change the pwerception that power doesn't work properly.  If we don't release it, `NDCube` hasn't gotten any worse and when we do release `pow` it should be a much more positive users experience.  I'm not aware that we have any users asking explictly for `pow` at the moment...

ndcube/ndcube.py
DanRyanIrish: It needs to be attached.  I think the way to do it is attach the hidden version of global_coords to the new object.
`new_cube._global_coords = self._global_coords`

ndcube/ndcube.py
DanRyanIrish: Oh.  I don't remember that.  Perhaps @Cadair can have the casting vote on which memory is right.

ndcube/ndcube.py
DanRyanIrish: Of course you’re right!

ndcube/ndcube.py
ayshih: I don't know if @wtbarnes intended to write `NotImplemented`, which is different from `NotImplementedError`.  I think situationally returning `NotImplemented` depending on the particulars of an instance rather than just the types is a bad idea.

wtbarnes: I specifically wrote `NotImplemented` and not `NotImplementedError`. My only reasoning for this is this is what we return in the arithmetic operators on Map when operations are not supported. I recall that we had a discussion about which to return when the Map arithmetic operators were implemented and the consensus was `NotImplemented`. However, I've lost the thread on what exactly the reasoning was. I have no opinion here but we should probably be consistent across Map and NDCube.

ayshih: The problem with returning `NotImplemented` in only a subset of situations of using a particular operator (i.e., only when `self.certainties` is `False`) is that there is no way to provide a message to the user explaining why it doesn't work for that particular instance.  It'll simply report that there's no supported operator for those two types, which would actually be a lie.  `NotImplemented` is really intended for an operator that will *never* work given the types.

wtbarnes: Ah ok. Well that's certainly not the case here then (and also not the case in Map either). I would propose then that we actually return `NotImplementedError` and provide the reasoning in the error message.

wtbarnes: I think if we are going to support uncertainties, we should be making use of the error propagation provided by astropy; https://docs.astropy.org/en/stable/api/astropy.nddata.NDUncertainty.html#astropy.nddata.NDUncertainty.propagate

wtbarnes: I'm struggling with how exactly to deal with this, given that `uncertainty` is not even required to adhere to a particular type. If we do not want to support `__pow__` for only cubes *without* uncertainties, then I think it makes sense to just drop it for now.

DanRyanIrish: If you set uncertainty to an array it gets converted to `UnknownUncertainty` by `NDCube`,s parent class instantiator.  And `UnknownUncertainty` prevents error propagation.  So you're right in that sense that error propagation is not always possible.  So I think the solution is to check the type.  If `uncertainty` is `UnknownUncertainty` or `None`, then drop uncertainties.  Otherwise `uncertainty` must be a valid uncertainty type that supports propagation.

As I said before, I think it's reasonable to drop `__pow__` from this PR and save it for its own where these issues do not hold up the other arithmetic operations.

Cadair: > Ah ok. Well that's certainly not the case here then (and also not the case in Map either). I would propose then that we actually return `NotImplementedError` and provide the reasoning in the error message.

If we are going to deviate from `Map` here then I would like to change this in `Map` as well, so that we don't end up slit brained (also it might generate more discussion on a sunpy PR).

Cadair (dismissed):
I still need to review this properly, but I want to keep the behaviour of NDCube and Map the same (in all things) at the moment, so if we are going to deviate from map for `NotImplemeted` vs `NotImplementedError` then I would want a PR open to sunpy core with the same change before we merge this one.

ayshih:
To be clear, the deviation from `Map` is not `NotImplemented` versus `NotImplementedError`, because `Map` does not return either for the situation under discussion.  If there are uncertainties on the data, `Map` doesn't care at all, and in fact discards any uncertainties in its arithmetic operations.

ndcube/ndcube.py
DanRyanIrish: After further discussion over :beers: let's leave these on `NDCube` for now as there's no penalty for moving them to `NDCubeBase` later whereas there is if we decide to go the other way.

ndcube/ndcube.py
```diff
@@ -792,3 +801,69 @@ def plot(self, *args, **kwargs):
...
+        if hasattr(value, 'unit'):
+            if isinstance(value, u.Quantity):
+                # NOTE: if the cube does not have units, set the unit
+                # to dimensionless such that we can perform arithmetic
+                # between the two.
+                cube_unit = u.Unit('') if self.unit is None else self.unit
+                new_data = self.data * value.to_value()
+                new_unit = cube_unit * value.unit
+            else:
+                return NotImplementedError('Multiplication is not supported between unitful non-quantity objects.')
```
DanRyanIrish, line 859:
```suggestion
                return NotImplementedError('Multiplication is currently not supported between unitful non-quantity objects.')
```

ndcube/ndcube.py
```diff
@@ -792,3 +801,69 @@ def plot(self, *args, **kwargs):
...
+        else:
+            new_data = self.data * value
+            new_unit = self.unit
+        return self._new_instance_from_op(new_data, new_unit)
```
DanRyanIrish, line 863: For multiplication, I just realised the uncertainties also need to be scaled.

DanRyanIrish (approved):

DanRyanIrish:
After discussion with @ayshih we identified the following remaining issues.
* In fact, unsupported type combinations for the operations should return `NotImplemented`, not `NotImplementedError`.  `NotImplemented` allows other types to override the operators to handle `NDCube`s which we should allow while `NotImplementedError` does not.  Moreover, this would not require a fix in sunpy to maintain alignment with `Map`.
* `__add__` should require unitful values if the `NDCube` has a unit, and require unitless values if `self.unit` is `None`.
* A point to highlight that does not require a change to this PR: There is divergence between this PR and `Map` in that `M̀ap` always drops uncertainties when any operation is applied.  This PR does not do that.  This is a shortcoming of `Map` that would be addressed by rebasing onto `NDCube`.  `Map` suddenly not dropping uncertainties should cause no issues as any code written for `Map`s without uncertainties should not break if there are uncertainties.

I will leave a few code suggestions in inline comments that should address the above issues.

ndcube/ndcube.py
```diff
@@ -792,3 +801,74 @@ def plot(self, *args, **kwargs):
...
+                # NOTE: if the cube does not have units, set the unit
+                # to dimensionless such that we can perform arithmetic
+                # between the two.
+                cube_unit = u.Unit('') if self.unit is None else self.unit
+                value_unit = value.unit
+                value = value.to_value()
+                new_data = self.data * value
+                new_unit = cube_unit * value_unit
+            else:
+                raise NotImplementedError('Multiplication is not currently supported between unitful non-quantity objects.')
```
DanRyanIrish, line 861:
```suggestion
                return NotImplemented
```

ndcube/ndcube.py
```diff
@@ -792,3 +801,74 @@ def plot(self, *args, **kwargs):
...
+                # NOTE: if the cube does not have units, we cannot
+                # perform arithmetic between a unitful quantity.
+                # This forces a conversion to a dimensionless quantity
+                # so that an error is thrown if value is not dimensionless
+                cube_unit = u.Unit('') if self.unit is None else self.unit
+                new_data = self.data + value.to_value(cube_unit)
+            else:
+                # NOTE: This explicitly excludes other NDCube objects and NDData objects
+                # which could carry a different WCS than the NDCube
+                raise NotImplementedError('Addition between NDCube objects and non-quantity objects is not currently supported.')
```
DanRyanIrish, line 835:
```suggestion
                return NotImplemented
```

ndcube/ndcube.py
```diff
@@ -792,3 +801,74 @@ def plot(self, *args, **kwargs):
...
+        else:
```
DanRyanIrish, line 836:
```suggestion
        elif self.unit not in {None, u.Unit("")}:
            raise TypeError("Cannot add a unitless object to an NDCube with a unit.")
        else:
```

ndcube/ndcube.py
```diff
@@ -756,6 +756,15 @@ class NDCube(NDCubeBase):
         Default is False.
 
     """
+    # Enabling the NDCube reflected operators is a bit subtle.  The NDCube
+    # reflected operator will be used only if the Quantity non-reflected operator
+    # returns NotImplemented.  The Quantity operator strips the unit from the
+    # Quantity and tries to combine the value with the GenericMap using NumPy's
```
wtbarnes, line 762:
```suggestion
    # Quantity and tries to combine the value with the NDCube using NumPy's
```

wtbarnes:
Just one minor correction in a comment, but otherwise I think this is ready to go

DanRyanIrish (approved):

DanRyanIrish:
Thanks so much @wtbarnes for this!!!!

PR #542, opened 2022-08-26
Add example for making NDCube from solar radio images

murphp30:
## PR Description
Adds example for making NDCube from solar radio images as discussed at Sunpy Coordination Meeting 2022.
 
I wanted to showcase something that came out of the hack day at the Sunpy Coordination Meeting 2022, creating a single NDCube for all solar radio images in a given frequency.  This uses an experimental class in the dksit python package https://github.com/DKISTDC/dkist/blob/4d03dcd96719448b91f2ec38bf831e055fd8c773/dkist/wcs/models.py#L218
and as such will likely need to be modified at some point.


murphp30:
Looking at this again and there's a bug that needs fixing. I'll mark as WIP in the meantime.

examples/ndcube_from_radio_images_example.py
```diff
@@ -0,0 +1,152 @@
...
+    beam_semi_major_axis = 0.5*header['BMAJ']*u.deg #nothing in fits to suggest degrees, you just have to know
+    beam_semi_minor_axis = 0.5*header['BMIN']*u.deg
+    beam_area = ((np.pi*beam_semi_major_axis*beam_semi_minor_axis)/(4*np.log(2)))
+    data = data *(u.Jy/beam_area)
+    equiv = u.brightness_temperature(freq)
+    data = data.to(u.K, equivalencies=equiv)
+    
+    return data, obstime
+
+with Pool() as pool:
```
DanRyanIrish, line 60: A comment here saying what `Pool` does and why you're using it might be helpful for new users.

examples/ndcube_from_radio_images_example.py
```diff
@@ -0,0 +1,152 @@
...
+
+##############################################################################
+# We first need to load some data. These exist on a machine somewhere and 
+# I don't know how to share them. 
+# Here we take every image at a single frequency
+
+ims = glob.glob("/data/mpearse/radio_images/SB158/*image.fits")
+ims.sort()
+
+def get_data(file):
```
DanRyanIrish, line 42: Would this example be just as helpful if you created fake data in this function, e.g. using `numpy.random` or `numpy.ones`, and explicitly defined the time, frequency, etc., instead of reading it from a file?  That way you wouldn't have to worry about how to share the files.

murphp30: :+1: definitely, yeah! That's much easier.

examples/ndcube_from_radio_images_example.py
murphp30: Good idea.

DanRyanIrish:
@murphp30 Could as well as the bugfix, the data generation and extra comments, could you also add a changelog and make the precommit code formatting changes?

examples/ndcube_from_radio_images_example.py
```diff
@@ -0,0 +1,152 @@
...
+# telescope pointing is updated to stay somewhat centred on the sun.
+# Thus, we will need create look up tables for the different reference 
+# coordinates and corresponding rotation matrices. We will also need the
+# location of the NenuFAR array.
+
+nenufar_ITRF = np.array((4323915,165533.67,4670321.7))
+array_loc = EarthLocation.from_geocentric(*nenufar_ITRF, u.m)
+
+def get_lookup_tables(file):
+    with fits.open(file) as hdu:
```
DanRyanIrish, line 76: Same comment as above.  Could these values be generated here rather than reading them in from a file.  I think the point of this example is to show how to create a WCS with a time-varying CRVAL rather than reading a FITS file.  Generating the data would make this easier to follow and remove the issue of sample data file handling in the package.

murphp30: Yeah, I've set it so that the data and lookup tables are just random numpy arrays of the correct shape. This also removes the need to use mulitproccessing and fits header parsing!

murphp30:
Sorry for the flurry of commits 😬 I don't understand why I'm failing the "fix end of files" check, the script ends with a blank line.

DanRyanIrish:
Hi @murphp30.  Is this ready for another review?  The doc build is failing because you use the `dkist` package which isn't installed.  @nabobalis is there a way to make this example work without making ndcube depend on `dkist` which could otherwise end up in a circular import nightmare?  Or would the models @murphp30 is using have to be upstreamed from `dkist`?

nabobalis:
If it's just docs, you can add dkist to that part of the setup.cfg and it will install it

murphp30:
Hey yes, I think this could do with another review for sure. I don't prentend to fully understand what I'm doing so please let me know if I've not described something accurately/correctly.
One thing in particular actually, which I touch on in a comment in line 92-93, when I try something like `cube.wcs.world_to_pixel(66.47644385*u.arcsec, 7.89916619*u.arcsec, Time("2022-05-19T12:27:36.4"))` I get a pixel coordinate with the correct values but the unit is `pix2`.
As for including `dkist` for the docs, is that just a simple as adding `dkist` somehwere under [options.extras_require]
in setup.cfg?

nabobalis:
>  As for including `dkist` for the docs, is that just a simple as adding `dkist` somehwere under [options.extras_require] in setup.cfg?

Pretty much, if you add it to this section: https://github.com/sunpy/ndcube/blob/main/setup.cfg#L51 
it should work.

PR #544, opened 2022-08-26
NDCube from FITS file example for gallery

hayesla:
## PR Description

Adding a simple example of how to read data from a FITS file and create a NDCube.

This uses the astropy example data - but happy to change this


DanRyanIrish:
Hi @hayesla.  This looks great.  Can you add a changelog and make the precommit changes?  I think this will be good to merge after that.

hayesla:
thanks @DanRyanIrish - I've made those fixes now. Not sure what the tests are failing on, doesn't seem to be related to this (unless I'm missing something!)

DanRyanIrish:
@hayesla Yeh, those fails aren't you.  Astropy has deprecated an kwargs in `reproject` which we haven't changed yet.

DanRyanIrish (approved):

PR #548, opened 2022-09-06
Reproject 0.9 changes

nabobalis:
I set the old defaults to suppress the warnings, I am unsure if this is a good idea.

Q: Do you want to bump reprojects version?
A: NO MUST WORK WITH 0.8

### DIFF
--- a/ndcube/ndcube.py
+++ b/ndcube/ndcube.py
@@ -614,48 +614,85 @@ class NDCubeBase(NDCubeSlicingMixin, NDCubeABC):
             raise ValueError(f"The 'algorithm' argument must be one of "
                              f"{', '.join(order_compatibility.keys())}.")
 
-    def reproject_to(self, target_wcs, algorithm='interpolation', shape_out=None, order='bilinear',
-                     output_array=None, parallel=False, return_footprint=False):
+    def reproject_to(self, target_wcs, algorithm='interpolation', return_footprint=False, **reproject_args):
         """
-        Reprojects this NDCube to the coordinates described by another WCS object.
+        Reprojects this `~nducbe.NDCube` to the coordinates described by another WCS object.
 
         Parameters
         ----------
-        algorithm: `str`
-            The algorithm to use for reprojecting. This can be any of: 'interpolation', 'adaptive',
-            and 'exact'.
-
         target_wcs : `astropy.wcs.wcsapi.BaseHighLevelWCS`, `astropy.wcs.wcsapi.BaseLowLevelWCS`,
             or `astropy.io.fits.Header`
             The WCS object to which the ``NDCube`` is to be reprojected.
 
+        algorithm: `str`
+            The algorithm to use for reprojecting.
+            This can be any of: 'interpolation', 'adaptive', and 'exact'.
+
+        return_footprint : `bool`
+            If `True`` the footprint is returned in addition to the new `~ndcube.NDCube`.
+            Defaults to `False`.
+
         shape_out: `tuple`, optional
-            The shape of the output data array. The ordering of the dimensions must follow NumPy
-            ordering and not the WCS pixel shape.
+            The shape of the output data array.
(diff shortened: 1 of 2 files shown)


ndcube/ndcube.py
```diff
@@ -640,7 +640,7 @@ def reproject_to(self, target_wcs, algorithm='interpolation', shape_out=None, or
             algorithm is selected).
             For 'interpolation' algorithm, this can be any of: 'nearest-neighbor', 'bilinear',
             'biquadratic', and 'bicubic'.
-            For 'adaptive' algorithm, this can be either 'nearest-neighbor' or 'bilinear'.
+            For 'adaptive' algorithm, this has no affect.
```
nabobalis, line 643:
```suggestion
            For 'adaptive' algorithm, this has no effect.
```

ndcube/ndcube.py
```diff
@@ -709,8 +709,13 @@ def reproject_to(self, target_wcs, algorithm='interpolation', shape_out=None, or
                                     return_footprint=return_footprint)
 
         elif algorithm == 'adaptive':
+            if order:
+                warnings.warn(
+                    "Order is ignored for the adaptive algorithm",
+                    UserWarning
+                )
```
nabobalis, line 716: Maybe its better to not raise a warning and just let it be ignored?

ndcube/ndcube.py
```diff
@@ -709,8 +709,13 @@ def reproject_to(self, target_wcs, algorithm='interpolation', shape_out=None, or
...
             data = reproject_adaptive(self, output_projection=target_wcs, shape_out=shape_out,
-                                      order=order, return_footprint=return_footprint)
+                                      return_footprint=return_footprint, kernel="Gaussian", boundary_mode="ignore")
```
nabobalis, line 718: Extra defaults to suppress warnings from reproject.

Cadair:
Two comments also made in chat:

* Need to maintain compatibility with <0.9
* We should align map and ndcube (by changes in either).

I will look at this more tomorrow.

ndcube/ndcube.py
```diff
@@ -614,64 +614,103 @@ def _validate_algorithm_and_order(self, algorithm, order):
             raise ValueError(f"The 'algorithm' argument must be one of "
                              f"{', '.join(order_compatibility.keys())}.")
 
-    def reproject_to(self, target_wcs, algorithm='interpolation', shape_out=None, order='bilinear',
-                     output_array=None, parallel=False, return_footprint=False):
+    def reproject_to(self, target_wcs, algorithm='interpolation', return_footprint=False, **reproject_args):
```
nabobalis, line 617: I here you like breaking API.

ndcube/ndcube.py
```diff
@@ -694,7 +733,16 @@ def reproject_to(self, target_wcs, algorithm='interpolation', shape_out=None, or
...
             raise ValueError('Given target_wcs is not compatible with this NDCube, the physical types do not match.')
 
-        # If shape_out is not specified explicity, try to extract it from the low level WCS
+        # LOL AT THIS
+        shape_out = reproject_args.get("shape_out", None)
+        order = reproject_args.get("order", 'bilinear')
+        output_array = reproject_args.get("output_array", None)
+        parallel = reproject_args.get("parallel", False)
+        reproject_args.get("kernel", "Gaussian")
+        reproject_args.get("boundary_mode", "ignore")
```
nabobalis, line 742: This is one way I guess?

nabobalis:
Doc build is fixed in another PR.

ndcube/ndcube.py
nabobalis: I can probably pop these, pass the final `reproject_args` to reproject instead?

Cadair:
@MeeseeksDev hello

Cadair:
I am going to adopt this PR.

PR #552, opened 2022-09-09
Fix ndcube with reproject 0.9

Cadair:
Taken over from #548

### DIFF
--- a/ndcube/ndcube.py
+++ b/ndcube/ndcube.py
@@ -610,9 +610,9 @@ class NDCubeBase(NDCubeSlicingMixin, NDCubeABC):
 
         algorithm: `str` {'interpolation', 'adaptive', 'exact'}
             The algorithm to use for reprojecting.
-            When set to 'interpolation' ~`reproject.reproject_interp` is used,
-            when set to 'adaptive' ~`reproject.reproject_adaptive` is used and
-            when set to 'exact' ~`reproject.reproject_exact` is used.
+            When set to `'interpolation'` `~reproject.reproject_interp` is used,
+            when set to `'adaptive'` `~reproject.reproject_adaptive` is used and
+            when set to `'exact'` `~reproject.reproject_exact` is used.
 
         shape_out: `tuple`, optional
             The shape of the output data array. The ordering of the dimensions must follow NumPy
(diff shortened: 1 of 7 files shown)


ndcube/ndcube.py
```diff
@@ -598,85 +598,74 @@ def explode_along_axis(self, axis):
...
         Returns
         -------
         resampled_cube : `ndcube.NDCube`
             A new resultant NDCube object, the supplied ``target_wcs`` will be the ``.wcs`` attribute of the output ``NDCube``.
 
         footprint: `numpy.ndarray`
-            Footprint of the input array in the output array. Values of 0 indicate no coverage or
-            valid values in the input image, while values of 1 indicate valid values.
+            Footprint of the input array in the output array.
+            Values of 0 indicate no coverage or valid values in the input
```
DanRyanIrish, line 639:
```suggestion
            Values of 0 indicate no coverage or no valid values in the input
```

DanRyanIrish:
To summarise private communications between myself and @Cadair, this PR addresses a  `reproject` `FutureWarning` triggered by `NDCube.reproject_to`.  This fix involves tweaking the API to enable reproject kwargs to be passed on through to the reproject algorithm.  Users will not notice the API change unless they have been providing positional args after the `shape_out` arg/kwarg.  

Presumably this is a tiny number of users.  Nonetheless, accepting this PR would break the API without giving users prior warning.  @Cadair argues that this is acceptable in line with semantic versioning if our next release including this change is a minor release (2.1) and not a bugfix release (2.0.3).  Moreover, this API is an improvement on our current one.

At this stage I have two responses:
1. (Re. whether this API is better):  I agree.
2. (Re. whether breaking the API is acceptable for a 2.1 release): My understanding was that under the guidelines of semantic versioning, users should not have to worry about breaking changes between minor releases.  This is reserved for major releases.  If this is true we should save this change for 3.0 and issue a `FutureWarning` of our own in 2.1 to alert people of the future change.

That being said, this minor change does not *feel* worthy of a major release and my suggested roadmap is a lot of hassle for what presumably is a tiny number of users who will even notice the difference.  Given that I might be persuaded that it may be acceptable to break the rules here...

DanRyanIrish:
On another note, does this change require reproject 0.9?  If so, I don't see where the source code pins reproject to >= 0.9...

DanRyanIrish (approved):

Cadair:
So are we backporting this to `2.0.3` or are we cutting a `2.1` release soon? If we aren't going to backport this we should probably release a bugfix release with a max version of reproject 0.9.

PR #558, opened 2022-09-22
Fix internal imports to avoid weird circular import error

Cadair:
I have no idea why my prompt was throwing me an error without this fix, and now it seems not to be again.

Anyway, this is probably a slight improvement or at least a neutral anyway.


nabobalis (approved):

nabobalis:
Always been a more pro using the full import paths for internal code over the high-level API.

PR #571, opened 2022-11-09
311 testing

nabobalis:


DanRyanIrish (approved):

.github/workflows/ci.yml
```diff
@@ -57,8 +57,11 @@ jobs:
       submodules: false
       coverage: codecov
       toxdeps: tox-pypi-filter
+      libraries: |
+        apt:
+          - libhdf5-dev
```
Cadair, line 62: I don't think we need this? We only install sunpy with core deps to test coordinate stuff so we don't ever import timeseries?
```suggestion
```

PR #584, opened 2023-01-05
Regression with asserting ndcubes inside numpy arrays

Cadair:
I use this in [my dkist code](https://github.com/DKISTDC/dkist/blob/main/dkist/dataset/tiled_dataset.py#L69) where I have a numpy object array of ndcube objects and this change:

https://github.com/sunpy/ndcube/blob/73ca267b87143aebe13766765cab5d47fe7b08ff/ndcube/ndcube.py#L766

caused this to stop working.

Investigation continues but here's a regression test.

### DIFF
--- a/ndcube/ndcube.py
+++ b/ndcube/ndcube.py
@@ -756,15 +756,6 @@ class NDCube(NDCubeBase):
         Default is False.
 
     """
-    # Enabling the NDCube reflected operators is a bit subtle.  The NDCube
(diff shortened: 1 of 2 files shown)


DanRyanIrish (approved):

DanRyanIrish:
Given that the pressing need for this PR has been circumvented by downstream changes I have re-milestoned this to 2.2.

PR #586, opened 2023-01-26
Add unit conversion method to NDCube.

DanRyanIrish:
- [x] Move `NDCubeBase.to` to `NDCube.to` as it relied on `__mul__` which is on `NDCube`.
- [x] Add test a fixture using a dask array and check that the data array type is still dask after the operation.
- [x] Add dask to test dependencies

### DIFF
--- a/ndcube/ndcube.py
+++ b/ndcube/ndcube.py
@@ -805,7 +805,6 @@ class NDCube(NDCubeBase):
     def _new_instance_from_op(self, new_data, new_unit, new_uncertainty):
         # This implicitly assumes that the arithmetic operation does not alter
         # the WCS, mask, or metadata.
-        new_uncertainty = deepcopy(self.uncertainty)
         new_cube = type(self)(new_data,
                               unit=new_unit,
                               wcs=self.wcs,
--- a/ndcube/ndcube.py
+++ b/ndcube/ndcube.py
@@ -709,26 +709,6 @@ class NDCubeBase(NDCubeSlicingMixin, NDCubeABC):
 
         return resampled_cube
 
-    def to(self, new_unit, **kwargs):
-        """Convert instance to another unit.
-
-        Converts the data, uncertainty and unit and returns a new instance
-        with other attributes unchanged.
-
-        Parameters
-        ----------
-        new_unit: `astropy.unit.Unit`
-            The unit to convert to.
-        kwargs:
-            Passed to the unit conversion method, self.unit.to.
-
-        Returns
-        -------
-        : `ǸDCube`
-            A new instance with the new unit and data and uncertainties scales accordingly.
-        """
-        return self * (self.unit.to(new_unit, **kwargs) * new_unit / self.unit)
-
 
 class NDCube(NDCubeBase):
     """
@@ -895,3 +875,23 @@ class NDCube(NDCubeBase):
 
     def __truediv__(self, value):
         return self.__mul__(1/value)
+
+    def to(self, new_unit, **kwargs):
+        """Convert instance to another unit.
+
+        Converts the data, uncertainty and unit and returns a new instance
+        with other attributes unchanged.
+
+        Parameters
+        ----------
+        new_unit: `astropy.unit.Unit`
+            The unit to convert to.
+        kwargs:
+            Passed to the unit conversion method, self.unit.to.
+
+        Returns
+        -------
+        : `ǸDCube`
+            A new instance with the new unit and data and uncertainties scales accordingly.
+        """
+        return self * (self.unit.to(new_unit, **kwargs) * new_unit / self.unit)
--- a/ndcube/ndcube.py
+++ b/ndcube/ndcube.py
@@ -802,11 +802,10 @@ class NDCube(NDCubeBase):
 
         return self.plotter.plot(*args, **kwargs)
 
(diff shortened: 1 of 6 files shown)


ndcube/ndcube.py
```diff
@@ -857,16 +879,15 @@ def __mul__(self, value):
...
             else:
                 return NotImplemented
         else:
-            new_data = self.data * value
             new_unit = self.unit
-        new_cube = self._new_instance_from_op(new_data, new_unit)
-        if new_cube.uncertainty is not None:
-            new_cube.uncertainty.array *= value
+        new_data = self.data * value
+        new_uncertainty = (type(self.uncertainty)(self.uncertainty.array * value)
```
wtbarnes, line 888: Is it guaranteed that the uncertainty has the same units as `data`? For example, could `data` be in J and uncertainty be expressed in mJ?

wtbarnes (commented):
This looks good to me. Using `__mul__` to do the unit conversion is pretty slick as well.

There are no obvious places where this would cause problems with Dask given that the array is not being converted to a `Quantity`.

My one other comment is I wonder whether this can be used with the `data_unit` kwarg when plotting. Currently, this is done manually by first converting the data to a `Quantity` and letting `Quantity` do the conversion.

I would also suggest using the Dask ndcube (~~which I think already lives in conftest??~~ Nope, definitely not) to implement a test of `to` with a Dask array.

ndcube/ndcube.py
DanRyanIrish: It is guaranteed as there is only one unit associated with the cube.  Of course the user could enter them differently but the interpretation of an NDCube is that they are the same by definition.

DanRyanIrish:
> I would also suggest using the Dask ndcube (~which I think already lives in conftest??~ Nope, definitely not) to implement a test of `to` with a Dask array.

So a new dask-based NDCube fixture would have to be implemented?  And then how would such a test be different from the test that's already there apart from the initial cube being a dask one?

DanRyanIrish:
> My one other comment is I wonder whether this can be used with the `data_unit` kwarg when plotting. Currently, this is done manually by first converting the data to a `Quantity` and letting `Quantity` do the conversion.

It should be possible. Alternatively, does this method make the `data_unit` kwarg obsolete as user could just do
`>>> cube.to(data_unit).plot()`

wtbarnes:
> > My one other comment is I wonder whether this can be used with the `data_unit` kwarg when plotting. Currently, this is done manually by first converting the data to a `Quantity` and letting `Quantity` do the conversion.
> 
> It should be possible. Alternatively, does this method make the `data_unit` kwarg obsolete as user could just do `>>> cube.to(data_unit).plot()`

Oh that's a good point! Yeah, actually it probably does...

wtbarnes:
> > I would also suggest using the Dask ndcube (~which I think already lives in conftest??~ Nope, definitely not) to implement a test of `to` with a Dask array.
> 
> So a new dask-based NDCube fixture would have to be implemented? And then how would such a test be different from the test that's already there apart from the initial cube being a dask one?

Yes to a new fixture. And I think I was thinking that a new test with this fixture would also test that the data array is still a Dask array even after calling `to`, i.e. it preserves the laziness of the operation.

DanRyanIrish:
@nabobalis How can I add `dask` to ndcube's test dependencies without adding it as dependency for using the package?  I tried adding it in locally here in [setup.cfg](https://github.com/sunpy/ndcube/blob/main/setup.cfg#L51) and [tox.ini](https://github.com/sunpy/ndcube/blob/main/tox.ini#L116) but it didn't work.

DanRyanIrish:
@wtbarnes This is ready for another review based on your feedback yesterday.

wtbarnes:
> @nabobalis How can I add `dask` to ndcube's test dependencies without adding it as dependency for using the package? I tried adding it in locally here in [setup.cfg](https://github.com/sunpy/ndcube/blob/main/setup.cfg#L51) and [tox.ini](https://github.com/sunpy/ndcube/blob/main/tox.ini#L116) but it didn't work.

Can't you just add it to the list of test dependencies here? https://github.com/sunpy/ndcube/blob/73ca267b87143aebe13766765cab5d47fe7b08ff/setup.cfg#L47

DanRyanIrish:
> Can't you just add it to the list of test dependencies here?
> 
> https://github.com/sunpy/ndcube/blob/73ca267b87143aebe13766765cab5d47fe7b08ff/setup.cfg#L47

That's what I tried but it didn't work.  Am I missing something?

DanRyanIrish:
> > @nabobalis How can I add `dask` to ndcube's test dependencies without adding it as dependency for using the package? I tried adding it in locally here in [setup.cfg](https://github.com/sunpy/ndcube/blob/main/setup.cfg?rgh-link-date=2023-01-27T09%3A06%3A53Z#L51) and [tox.ini](https://github.com/sunpy/ndcube/blob/main/tox.ini?rgh-link-date=2023-01-27T09%3A06%3A53Z#L116) but it didn't work.
> 
> I am so sorry, I do not know how I missed this message.
> 
> Yeah as Will said, adding them to the tests block in setup.cfg should be enough. I do not see that in the diff, was it there before?

It works now.  Don't know why it didn't before.  Presumably user error.  Thanks!

PR #594, opened 2023-02-27
Fix failing figure tests

DanRyanIrish:
## PR Description
This PR updates the dependency versions to fix a bug causing the figure tests to fail.


nabobalis (approved):

nabobalis:
A hash changed:

```
________________________________________________________________________________ test_plot_1D_cube _________________________________________________________________________________
Hash 040edf223f40754b7a53915da165d10dad9d4456622f1f8217269679c3b22550 doesn't match hash 9cb2c58dc884dc0b5c63afb79341ec23008a31863ae335bb885d8a89abff6b4c in library /home/circleci/project/.tox/py310-figure/lib/python3.10/site-packages/ndcube/visualization/tests/figure_hashes_mpl_353_ft_261_astropy_51_animators_100.json for test test_plotting.test_plot_1D_cube.
```

nabobalis:
Why did the bot not link the results of the figure test?

DanRyanIrish:
So `result` is the figure that was produced and `baseline` is what was expected, right?  It looks to me like `result` is correct.  Has changing the hash made the original `baseline` the same as the `result`?

nabobalis:
> So `result` is the figure that was produced and `baseline` is what was expected, right? 

Yes.

> It looks to me like `result` is correct. Has changing the hash made the original `baseline` the same as the `result`?

The hash is just a check against the hash in the file. The baseline image is only updated on merge when circleci runs on main. Until then, the baseline image will be wrong

PR #605, opened 2023-03-13
Allow passing units as strings to NDCube.to

wtbarnes:
This PR fixes a bug where passing a string representation of a unit to `NDCube` raises the following `TypeError`,

```python-traceback
File ~/mambaforge/envs/coronal-rain-observables/lib/python3.10/site-packages/ndcube/ndcube.py:964, in NDCube.to(self, new_unit, **kwargs)
    946 def to(self, new_unit, **kwargs):
    947     """Convert instance to another unit.
    948 
    949     Converts the data, uncertainty and unit and returns a new instance
   (...)
    962         A new instance with the new unit and data and uncertainties scales accordingly.
    963     """
--> 964     return self * (self.unit.to(new_unit, **kwargs) * new_unit / self.unit)

TypeError: can't multiply sequence by non-int of type 'float'
```

### DIFF
--- a/ndcube/ndcube.py
+++ b/ndcube/ndcube.py
@@ -961,6 +961,7 @@ class NDCube(NDCubeBase):
         : `ǸDCube`
             A new instance with the new unit and data and uncertainties scales accordingly.
         """
+        new_unit = u.Unit(new_unit)
         return self * (self.unit.to(new_unit, **kwargs) * new_unit / self.unit)
 
     def rebin(self, bin_shape, operation=np.mean, operation_ignores_mask=False, handle_mask=np.all,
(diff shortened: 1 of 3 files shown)


ndcube/ndcube.py
```diff
@@ -961,6 +961,7 @@ def to(self, new_unit, **kwargs):
         : `ǸDCube`
             A new instance with the new unit and data and uncertainties scales accordingly.
         """
+        new_unit = u.Unit(new_unit)
```
DanRyanIrish, line 964: This forces the unit to be an astropy

DanRyanIrish (approved):

PR #636, opened 2023-09-09
[main]  Drop Python 3.8, Increase min deps versions, update tox config on CI, update circleci

nabobalis:
figure devdeps failing due to 3.8.0rc1 and the filename for the hash, is incorrect.


docs/installation.rst
```diff
@@ -4,7 +4,7 @@
 Installing ``ndcube``
 *********************
 
-`ndcube` requires Python >=3.8, ``astropy``>=4.2, ``numpy``>=1.1.7 and ``gwcs``>=0.15.
+`ndcube` requires Python >=3.9, ``astropy``>=5.0, ``numpy``>=1.21 and ``gwcs``>=0.18.
```
wtbarnes, line 7: Shouldn't all of these be single ticks?

nabobalis: Maybe if they interlink, I am unsure if they do.

pyproject.toml
```diff
@@ -53,13 +53,13 @@ build-backend = 'setuptools.build_meta'
     enabled = true
 
   [ tool.gilesbot.circleci_artifacts.figure_report ]
-    url = "py310-figure/figure_test_images/fig_comparison.html"
-    message = "Click details to see the figure test comparisons, for figure."
+    url = ".tmp/py310-figure/figure_test_images/fig_comparison.html"
```
DanRyanIrish, line 56: Should these urls really be to `.tmp` locations?

DanRyanIrish:
Looks like the devdeps figtests are still failing, perhaps due to a filepath location issue?

DanRyanIrish:
First things last though:  Thanks so much @nabobalis for tackling this!

docs/installation.rst
Cadair: They do not, double ticks is more correct.

pyproject.toml
Cadair: They are .tmp so that they don't pollute your local directory if you run this locally.

Cadair: That being said, you shouldn't actually need them here. Giles will match on any substring.

nabobalis: I can undo it?

tox.ini
```diff
@@ -28,7 +23,8 @@ setenv =
     COLUMNS = 180
     PYTEST_COMMAND = pytest -vvv -s -raR --pyargs ndcube --cov-report=xml --cov=ndcube --cov-config={toxinidir}/setup.cfg {toxinidir}/docs
     PARFIVE_HIDE_PROGESS = True
-    devdeps: PIP_EXTRA_INDEX_URL = https://pypi.anaconda.org/astropy/simple https://pypi.anaconda.org/scipy-wheels-nightly/simple
+    devdeps: PIP_INDEX_URL = https://pypi.anaconda.org/astropy/simple https://pypi.anaconda.org/scientific-python-nightly-wheels/simple
+    devdeps: PIP_EXTRA_INDEX_URL = https://pypi.org/simple
```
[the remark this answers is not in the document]
nabobalis: I will undo this, this was me playing around seeing what happens if I use INDEX_URL over EXTRA_INDEX for the devdeps build.

PR #642, opened 2023-09-23
Add links to ndcube publications to README and docs.

DanRyanIrish:


nabobalis (approved):

Cadair:
My only thought about this is if we should have a project-wide citation file? i.e. extend [this](https://github.com/sunpy/sunpy/blob/main/sunpy/CITATION.rst) one which get's pulled into the website etc?

nabobalis:
What would be the best way to host it? In our .github repo and add to that? 

Add a file and convert it to html on the sunpy.org website?

Cadair:
Well if we add all the various project ones to the one in core it will "just work"?

nabobalis:
Sounds good

PR #646, opened 2023-10-11
Update NDCollection copy() & pop() to Support None aligned_axes

### LINKED ISSUE #641: AttributeError thrown when calling NDCollection.pop() with None aligned_axes
### Describe the bug

I'm trying to remove a key/value from a `NDCollection` object that has `aligned_axes = None`. The `pop(key)` function correctly removes the key from the `NDCollection`, however an attribute error is then thrown trying to remove the key from the aligned axes. 
`AttributeError: 'NoneType' object has no attribute 'pop'`

I can get around this for now with a try/catch around the `pop()` function call however this is not ideal. 

### To Reproduce

```
import numpy as np
from astropy.wcs import WCS
from ndcube import NDCube, NDCollection

test_cube = NDCube(np.random.random((10,10)), wcs=WCS(naxis=2))
c = NDCollection([("Test", test_cube)])
assert c.aligned_axes == None

test_cube2 = NDCube(np.random.random((10,10)), wcs=WCS(naxis=2))
c.update([("Test2", test_cube2)], c.aligned_axes)
assert c.aligned_axes==None

c.pop("Test2")
```

Results in:
```
AttributeError: 'NoneType' object has no attribute 'pop'
```

### Screenshots

_No response_

### System Details

macOS-13.5.2-arm64-arm-64bit
sys.version_info(major=3, minor=8, micro=16, releaselevel='final', serial=0)
numpy=1.24.2
astropy=5.2.2
ndcube=2.1.3

### Installation method

pip

Alrobbertz:
## PR Description

* Adds `None` check on `self.aligned_axes` to `NDCollection.copy()` and `NDCollection.pop()`
* Adds test data to represent collection with `None` aligned axes in `test_ndcollection.py`
* Updates tests for `copy()` and `pop()` to test with new test data

closes #641

### DIFF
--- a/ndcube/ndcollection.py
+++ b/ndcube/ndcollection.py
@@ -222,8 +222,12 @@ class NDCollection(dict):
         return collection_items, new_aligned_axes
 
     def copy(self):
-        return self.__class__(self.items(), tuple(self.aligned_axes.values()),
-                              meta=self.meta, sanitize_inputs=False)
+        # Aligned axes is not a required parameter and may be None
+        if self.aligned_axes is not None:
+            return self.__class__(self.items(), tuple(self.aligned_axes.values()),
+                                  meta=self.meta, sanitize_inputs=False)
+        else:
+            return self.__class__(self.items(), None, meta=self.meta, sanitize_inputs=False)
 
     def setdefault(self):
         """Not supported by `~ndcube.NDCollection`"""
@@ -244,8 +248,10 @@ class NDCollection(dict):
         """
         # Extract desired cube from collection.
         popped_cube = super().pop(key)
-        # Delete corresponding aligned axes
-        self.aligned_axes.pop(key)
+        # Aligned axes is not a required parameter and may be None
+        if self.aligned_axes is not None:
+            # Delete corresponding aligned axes
+            self.aligned_axes.pop(key)
         return popped_cube
 
     def update(self, *args):
--- a/ndcube/tests/test_ndcollection.py
+++ b/ndcube/tests/test_ndcollection.py
@@ -38,6 +38,7 @@ sequence20 = NDCubeSequence([cube2, cube0])
 aligned_axes = ((1, 2), (2, 0), (1, 2))
 keys = ("cube0", "cube1", "cube2")
 cube_collection = NDCollection([("cube0", cube0), ("cube1", cube1), ("cube2", cube2)], aligned_axes)
+unaligned_collection = NDCollection([("cube0", cube0), ("cube1", cube1), ("cube2", cube2)], aligned_axes=None)
 seq_collection = NDCollection([("seq0", sequence02), ("seq1", sequence20)], aligned_axes="all")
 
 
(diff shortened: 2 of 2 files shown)


Alrobbertz:
Passed pre-commit but failing pep8 here? Some of the lines I changed would look out-of-place compared to their surroundings if made <80 characters. 

Recommendations?

nabobalis:
Ignore it, we stopped using it but it's somehow come back in the last week. Still trying to work out how.

Alrobbertz:
Do you need me to fix anything for the other checks? 

I fixed this on the mission side with a try-catch but figured I'd fix it here for whenever your next release is.

nabobalis:
Just a changelog entry for that check. The figure check is broken due to upstream changes.

changelog/646.bugfix.rst
```diff
@@ -0,0 +1,3 @@
+- Adds `None` check on `self.aligned_axes` to `NDCollection.copy()` and `NDCollection.pop()` which raised a `AttributeError`
```
nabobalis, line 1: We try to keep our changelog user facing, so a simpler description of what the bug fixed was. 

No need to go into much detail about specifics or tests added. 

Thanks!

nabobalis:
Thanks for the PR, I'll get one of the core maintainers to review this soon!

ndcube/ndcollection.py
```diff
@@ -222,8 +222,12 @@ def _generate_collection_getitems(self, item):
...
 
     def copy(self):
-        return self.__class__(self.items(), tuple(self.aligned_axes.values()),
-                              meta=self.meta, sanitize_inputs=False)
+        # Aligned axes is not a required parameter and may be None
+        if self.aligned_axes is not None:
+            return self.__class__(self.items(), tuple(self.aligned_axes.values()),
+                                  meta=self.meta, sanitize_inputs=False)
+        else:
+            return self.__class__(self.items(), None, meta=self.meta, sanitize_inputs=False)
```
DanRyanIrish, line 230: The logic looks correct here.  I think it could be written a little more succinctly.  See below.
```suggestion
        aligned_axes = None if self.aligned_axes is None else tuple(self.aligned_axes.values())
        return self.__class__(self.items(), aligned_axes, meta=self.meta, sanitize_inputs=False)
```

DanRyanIrish (commented):
Hi @Alrobbertz.  Thanks for this PR which definitely catches an important bug.  It looks to me like this PR is in great shape.  I only came up with one suggestion for how to write your change to ``copy`` a bit "nicer" (or at least more succinctly).  If you are happy with my suggested change and the tests are passing, I think this will be ready to merge.

ndcube/ndcollection.py
Alrobbertz: Funny enough, I thought about doing something like this from the start! I was weighing whether I liked the conciseness of the ternary operator of the more verbose clarity of the conditional block.

DanRyanIrish (approved):
lgtm!  Thanks again for the PR @Alrobbertz

PR #664, opened 2024-02-15
Adds sunpy package template functionality

CyclingNinja:
Adds sunpy package template automation using cruft and cookiecutter
Removes some configuration files which would now duplicate functionality


.pre-commit-config.yaml
```diff
@@ -49,15 +15,8 @@ repos:
...
       - id: debug-statements
       - id: check-added-large-files
-        args: ['--enforce-all','--maxkb=1054']
+        args: ["--enforce-all", "--maxkb=1054"]
       - id: end-of-file-fixer
-        exclude: ".*(.fits|.fts|.fit|.header|.txt|tca.*)$"
+        exclude: ".*(.fits|.fts|.fit|.header|.txt|tca.*|.json)$|^CITATION.rst$"
       - id: mixed-line-ending
         exclude: ".*(.fits|.fts|.fit|.header|.txt|tca.*)$"
-  - repo: https://github.com/codespell-project/codespell
```
nabobalis, line 57: Oh codespell didn't make it to the template?

.pre-commit-config.yaml
```diff
@@ -49,15 +15,8 @@ repos:
...
-    rev: v2.2.6
-    hooks:
-      - id: codespell
-        args: ['--config setup.cfg']
-ci:
-  autofix_prs: false
```
nabobalis, line 63: This didn't either? 
We should also disable updates if the plan is to roll them out via the template?

pyproject.toml
```diff
@@ -1,10 +1,89 @@
...
+    "pytest-cov",
+]
+docs = [
+    "sphinx",
+    "sphinx-automodapi",
+    "tomli; python_version <\"3.11\"",
+]
+
+[project.urls]
+repository = "https://docs.sunpy.org/projects/ndcube/"
```
nabobalis, line 33: We should expand these with docs and repository and homepage?

setup.py
```diff
@@ -1,30 +0,0 @@
-#!/usr/bin/env python
```
nabobalis, line 1: I thought this file had to stay?

nabobalis:
Looks like actions did not start?

setup.py
CyclingNinja: I'll stick it back in (first time using the interactive git add functoinality on kraken)

.github/workflows/ci.yml
```diff
@@ -1,3 +1,4 @@
+{% set default_python = '3.10' %}
```
Cadair, line 1: This shouldn't be here :thinking:

.github/workflows/ci.yml
```diff
@@ -15,33 +16,58 @@ on:
...
 
-# Automatically cancel old builds
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
   cancel-in-progress: true
 
 jobs:
   core:
-    uses: OpenAstronomy/github-actions-workflows/.github/workflows/tox.yml@v1
+    uses: OpenAstronomy/github-actions-workflows/.github/workflows/tox.yml@main
```
Cadair, line 25: we should change the template to use v1 not main.

.pre-commit-config.yaml
Cadair, line 1: When doing this manual retemplating we should keep all the extra stuff in files like this until it's updated in the template.

.pre-commit-config.yaml
Cadair: We should add this to the template, asap.

>  We should also disable updates if the plan is to roll them out via the template?

hmm, this is a good and interesting question. Yes, we should roll out updates via the template, this would mean we go from one (pre-commit) to one (template update) auto PR rather than one to two. We should figure out how we can automate bumping the versions in the template though or they will surely get forgotten.

https://github.com/sunpy/package-template/issues/33

pyproject.toml
Cadair: https://github.com/sunpy/package-template/issues/34

.pre-commit-config.yaml
Cadair: https://github.com/sunpy/package-template/issues/35

nabobalis:
Replaced by https://github.com/sunpy/ndcube/pull/672

Issue #591, opened 2023-02-14
NDCube does not support arithmetic via numpy ufuncs


Cadair:
### Describe the bug

Numpy [ufuncs](https://numpy.org/doc/stable/reference/ufuncs.html) are things like  `np.add` and `np.equal`, currently we support arithmetic methods via the standard Python operands like `+` (through `__add__`) but not the equivalent ufunc.

I discovered this issue because the following doesn't work:

```python
cube = NDCube(...)
arr = np.array([cube], dtype=object)
np.equal(arr, cube)
```
In addition some things like `np.maximum` and `np.isinf` might be quite nice to have working without having to access `.data`, also if we implement `__array_ufunc__` we can have those things return quantities.

---
See [Recommendations for implementing binary operations](https://numpy.org/neps/nep-0013-ufunc-overrides.html#recommendations-for-implementing-binary-operations) in the NEP for how I think we need to go about this.

The overview is:

* Implement all the arithmetic methods in `__array_ufunc__`.
* Have the Python operator methods `__add__` etc, call the numpy ufunc, i.e. `__add__` calls `np.add`.
* Ensure that our operator methods return `NotImplemented` if the other operand has `__array_ufunc__ = None`.

PR #666, opened 2024-03-10
Adds the functionality of Ufunc to NDCube

ViciousEagle03:
PR Description
This PR address the issue: #591

I've tried to add the support of Arithmetic operations to NDCube using Numpy Ufunc
I have removed the test cases in `ndcube/tests/test_ndcube.py` in the arithmetic section which doesn't support  `__array_ufunc__`

Note: Before proceeding to add the remaining test cases and Changelog, I wanted to ensure that the proof of concept is correct.

### DIFF
--- a/ndcube/ndcube.py
+++ b/ndcube/ndcube.py
@@ -883,11 +883,44 @@ class NDCube(NDCubeBase):
             new_cube._global_coords = deepcopy(self.global_coords)
         return new_cube
 
+    def __array_ufunc__(self, ufunc, method, *inputs, **kwargs):
+        if method == '__call__':
+            if ufunc == np.add:
+                if isinstance(inputs[0], NDCube):
+                    new_data = inputs[0].data + inputs[1]
+                    return new_data
+                else:
+                    return (self.__radd__(inputs[0]))
+            elif ufunc == np.subtract:
+                return (self.__rsub__(inputs[0]))
+            elif ufunc == np.multiply:
+                if isinstance(inputs[0], NDCube):
+                    new_data = inputs[0].data * inputs[1]
+                    return new_data
+                else:
+                    return (self.__rmul__(inputs[0]))
+            elif ufunc == np.equal:
+                if all(isinstance(inp, NDCube) for inp in inputs):
+                    return (np.equal(inputs[1].data, inputs[0].data))
+                else:
+                    NotImplemented
+            elif ufunc == np.maximum:
+                if all(isinstance(inp, NDCube) for inp in inputs):
+                    new_data = np.maximum(inputs[0].data, inputs[1].data)
+                    return self._new_instance_from_op(new_data, deepcopy(self.unit), deepcopy(self.uncertainty))
+            elif ufunc == np.isinf:
+                if isinstance(inputs[0], NDCube):
+                    return (np.isinf(inputs[0].data))
+        else:
+            return NotImplemented
+
     def __neg__(self):
         return self._new_instance_from_op(-self.data, deepcopy(self.unit),
                                           deepcopy(self.uncertainty))
 
     def __add__(self, value):
+        if getattr(value, '__array_ufunc__', None) is None:
+            return NotImplemented
         if hasattr(value, 'unit'):
             if isinstance(value, u.Quantity):
                 # NOTE: if the cube does not have units, we cannot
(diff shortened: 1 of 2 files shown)


ViciousEagle03:
Hey @DanRyanIrish, Thanks for the feedback 
I think both the ufunc and arithmetic operators APIs should provide the same implementation for the ufuncs and arithmetic operators and I have changed the `__array_ufunc__` method to take in account the same.The ufuncs now also  considers the WCS, uncertainties and mask where appropriate.
Regarding the implementation of np.equal ufunc, I think it is ambiguous to define the implementation of it 
Given the computational expense of comparing the WCSes of two NDCube ,if one is only comparing the data array of the cubes then it can be achieved by just `np.equal(cube1.data, cube2.data)` or if equal means comparing the reference both the cubes then we can check it by `is`.So, you are right to first establish what the np.equal should achieve and if you have any particular implementation you would like me to add to `np.equal`, I would be happy to change the patch accordingly.
I'm a bit uncertain about `np.maximum`,
You see `np.maximum` is a binary ufunc and changing the parameters to one, changing the convention of `np.maximum` could be misleading or confusing, Or did I misunderstood the above?
And if we implement the `np.maximum`, we are returning the two things below, right?
- the maximum value in the NDCube.data array
- the maximum of the data array and the corresponding uncertainty value for each corresponding data

ndcube/ndcube.py
```diff
@@ -886,28 +886,32 @@ def _new_instance_from_op(self, new_data, new_unit, new_uncertainty):
     def __array_ufunc__(self, ufunc, method, *inputs, **kwargs):
         if method == '__call__':
             if ufunc == np.add:
-                if isinstance(inputs[0], NDCube):
+                if isinstance(inputs[0], NDCube) and kwargs.get("dunder"):
```
DanRyanIrish, line 889: Is `"dunder"` a standard kwarg or custom for this implementation?  And what does it mean?

ndcube/ndcube.py
```diff
@@ -886,28 +886,32 @@ def _new_instance_from_op(self, new_data, new_unit, new_uncertainty):
...
                     new_data = inputs[0].data + inputs[1]
                     return new_data
```
DanRyanIrish, line 891: Does this mean that `np.add(cube1, array2)` would return an array?

This looks different to the implementation for `np.sub` which appears like it would return an `NDCube`.

DanRyanIrish (commented):
Hi @ViciousEagle03.  Thanks for the update to your PR and apologies for the delay in my reply.  I am still consumed with another project.  I am still trying to understand exactly the API you're implementing.  I've left some in line comments in places where I'm confused.

Perhaps it would be a good idea to list out as a comment on this PR the exact behaviour you are trying to implement.  For example, the output types of 
* `np.add(cube1, array2)`
* `np.add(cube1, cube2)`
* `np.add(array1, cube2)`
and what attributes are considered in the calculation, i.e. `.data`, `.mask`, `.uncertainty`.  And the same for `np.subtract`, and `np.multiply`.

ndcube/ndcube.py
ViciousEagle03: `np.add(cube1, array2)` would return an NDCube.
The `np.subtract` utilises the dunder method `__sub__` and in turn the __sub__ dunder method uses the `__add__` dunder method and since `np.subtract` utilises the `__add__` so we don't need to check the way we did in `np.add` as `__add__` dunder method will call the `np.add` which in turn will do the checking and hence both returns the same type.

ndcube/ndcube.py
ViciousEagle03: I have used the `dunder` (not a standard kwarg) kwarg here to distinguish between the operators and numpy ufuncs to properly handle operator overloading.

ViciousEagle03:
Hey @DanRyanIrish,
Apologies for the delay, I have introduced the kwarg `dunder` to handle the operator overloading properly and to leverage the existing python operators. Since there is little documentation available regarding the implementation of `array_ufunc` functionality for complex custom objects so the way i have implemented it might be a little confusing so i will try to explain the API I have implemented here,
If array2 supports `array_ufunc`
• Calling `np.add(cube1, array2)` would return an NDCube.
• Attempting `np.add(cube1, cube2)` would return `NotImplented`, triggering a `TypeError`.
• Calling `np.add(array1, cube2)` would also return an NDCube.
By default, `ufuncs` triggers the array_ufunc mechanism, while arithmetic operators invoke the respective dunder methods.
To ensure both ufuncs and arithmetic operators account for masks and uncertainties appropriately and follow the same API, I've used dunder to specify when an arithmetic  operator is being used.
For example in case of, `np.add(cube1, array2)`, the control proceeds as below:
`array_ufunc()`[dunder=false] --> `__add__()` --> `array_ufunc()` [dunder=True] --> now it proceeds similarly as `+` operator.
All numpy ufuncs leverage the respective existing dunder methods, which ensures the implementation is consistent.

PR #708, opened 2024-05-20
Asdf-Support

ViciousEagle03:
## This PR adds ASDF support for ndcube objects.

[GitHub Project Board Link](https://github.com/orgs/sunpy/projects/12/views/1)

This Pull Request is currently a `Work In Progress` and will be continuously updated as we expand support for serializing `ndcube` objects to ASDF.


ndcube/asdf/converters/ndcube_converter.py
```diff
@@ -0,0 +1,30 @@
+from asdf.extension import Converter
+import numpy as np
+
+
+class NDCubeConverter(Converter):
+    tags = ["tag:sunpy.org:ndcube/ndcube/NDCube-*"]
+
+    @property
+    def types(self):
```
braingram, line 9: Strings matching the "class path" can be used here instead of the actual types. Here's an example for the `astropy.column.Column` converter:
https://github.com/astropy/asdf-astropy/blob/1b286108042c1201199e9931cd2a5b52cc57bdce/asdf_astropy/converters/table/table.py#L11
Using a string here has the benefit that `ndcube` does not need to be imported when the `Converter` is "indexed" (which will occur during the first `asdf.open`). There is a small note on this in the [asdf docs](https://asdf.readthedocs.io/en/latest/asdf/extending/converters.html#entry-point-performance-considerations).

ndcube/asdf/converters/ndcube_converter.py
```diff
@@ -0,0 +1,30 @@
...
+        from ndcube.ndcube import NDCube
+        return [NDCube]
+
+    def select_tag(self, obj, tags, ctx):
```
braingram, line 13: I would leave this out for now. As you've noted in the comments (which are very helpful!) the ordering here can sometimes lead to confusion. The tag versions can be managed by versioning the extension (in different manifests) so that for a given extension there is only one tag version (and we won't need a `select_tag`).

ndcube/asdf/converters/ndcube_converter.py
```diff
@@ -0,0 +1,30 @@
...
+        # Sort the tags in reverse alphabetical order and pick the first (i.e.
+        # the one with the highest version). This assumes that all the tags for
+        # this converter are named the same other than the version number.
+        tags = list(sorted(tags, reverse=True))
+        return tags[0]
+
+    def from_yaml_tree(self, node, tag, ctx):
+        from ndcube import ndcube
+        data = np.asanyarray(node["data"])
```
braingram, line 22: Was `asanyarray` needed because `NDCube` won't accept `NDArrayType`?

ndcube/asdf/converters/ndcube_converter.py
```diff
@@ -0,0 +1,30 @@
...
+        wcs = node["wcs"]
+        return(ndcube.NDCube(data, wcs))
+
+    def to_yaml_tree(self, ndcube, tag, ctx):
+        node = {}
+        node["data"] = np.asarray(ndcube.data)
```
braingram, line 28: What's the need for `asarray` here? Are there circumstances where this could lead to a data copy?

ndcube/asdf/resources/schemas/NDCube-0.1.0.yaml
```diff
@@ -0,0 +1,21 @@
...
+title: |
+  Represents the ndcube NDCube object
+
+description:
+  Represents the ndcube NDCube object
+
+type: object
+properties:
+  data:
+    tag: "tag:stsci.edu:asdf/core/ndarray-1.0.0"
```
braingram, line 15:
```suggestion
    tag: "tag:stsci.edu:asdf/core/ndarray-1.*"
```
`tag` keywords can use wildcards which can help to deal with tag versions. In this case [ASDF-standard 1.6.0](https://github.com/asdf-format/asdf-standard/blob/c367d14fae4b5155679e2c40b531140586e25daf/resources/manifests/asdf-format.org/core/core-1.6.0.yaml#L64) will introduce `ndarray-1.1.0`. By using a wilcard here this schema can work with both ASDF-standard <1.6.0 and =1.6.0.

ndcube/asdf/resources/schemas/NDCube-0.1.0.yaml
```diff
@@ -0,0 +1,21 @@
...
+  wcs:
+    $ref: "http://stsci.edu/schemas/gwcs/wcs-1.1.0"
```
braingram, line 17:
```suggestion
    $ref: "http://stsci.edu/schemas/gwcs/wcs-1.*"
```

pyproject.toml
```diff
@@ -71,6 +71,12 @@ exclude = ["ndcube._dev*"]
 [tool.setuptools_scm]
 write_to = "ndcube/_version.py"
 
+[project.entry-points."asdf.resource_mappings"]
+ndcube = "ndcube.asdf.entry_points:get_resource_mappings"
+
+[project.entry-points."asdf.extensions"]
+ndcube = "ndcube.asdf.entry_points:get_extensions"
+
```
braingram, line 79: Would you also enable schema tests for the new schemas? Here's the configuration for `asdf-astropy`:
https://github.com/astropy/asdf-astropy/blob/1b286108042c1201199e9931cd2a5b52cc57bdce/pyproject.toml#L121-L122

I am not familiar with the ndcube tox and github actions CI but would you look into modifying whatever is necessary there to get the schema tests running for the PR? Let me know if there are more details or anything I can do to help with this. Also, perhaps one of the other maintainers has a sense for how these schema tests might be integrated in the CI.

braingram (commented):
Thanks for sharing your work. It looks like a great start! I left a bunch of comments. I did not look over the `ndcube.py` changes.

ndcube/asdf/converters/ndcube_converter.py
ViciousEagle03: I have used `np.asanyarray` to ensure that the data attribute maintains the `numpy.ndarray` type after deserialization. So, when an NDCube ASDF file is deserialized, the `.data` attribute retains its original `numpy.ndarray` type instead of an `asdf.tags.core.ndarray.NDArrayType` type. This would ensures consistency in the data type.

ndcube/asdf/converters/ndcube_converter.py
ViciousEagle03: I used `np.asarray` here also to ensure that the data attribute is consistently a `numpy.ndarray`. I used `np.asarray as a precautionary measure. We could also use `np.array`.

ndcube/asdf/resources/schemas/NDCube-0.1.0.yaml
ViciousEagle03: Right, will do this.

ndcube/asdf/resources/schemas/NDCube-0.1.0.yaml
ViciousEagle03: I think we can't use wildcard for `$ref`, I have used `anyOf` to incorporate different versions, is this fine?

braingram: Good catch. Yes, the wildcard will only work if this is a tag. Any reason not to use the tag instead of a `$ref`? I'll update the suggestion with what I think should work.

pyproject.toml
ViciousEagle03: I have enabled the schema test, @Cadair could you please verify if the configuration is correct?
I followed [this](https://asdf.readthedocs.io/en/latest/asdf/extending/schemas.html#testing-custom-schemas:~:text=%5Btool.pytest.ini_options%5D%0Aasdf_schema_root%20%3D%20%27path/to/schemas%20another/path/to/schemas%27) to enable the tests

ndcube/asdf/converters/ndcube_converter.py
ViciousEagle03: Right, it would be better to remove it to avoid any unnecessary delay in processing.
Apologies for the delay, I was unwell (Due to fluctuating weather conditions).

ndcube/asdf/converters/ndcube_converter.py
braingram: Thanks for the detailed response. I ask because `NDArrayType` is "lazy", it will only read the array data from the file when it's needed (not when the `NDArrayType` is created). If we have an ASDF file "foo.asdf" with an array under key "arr"
```python
>> af = asdf.open("foo.asdf")
>> print(af["arr"])  # print and some other operations (like shape) won't load the data
<array (unloaded) shape: [100, 100] dtype: float64>
```
Passing the `NDArrayType` through `asanyarray` will convert it to a `ndarray` (as you noted) and trigger the data to be loaded. I had a quick look and it appears that `NDCube` will accept the `NDArrayType`. However something in the constructor is triggering the `NDArrayType` to load the array data (this can happen even for seemingly innocuous calls like this call to [hasattr](https://github.com/sunpy/ndcube/blob/ce4d0b13705fcb8128a87402a23ac0378e7c58da/ndcube/ndcube.py#L379)).
Since the conversion to `ndarray` doesn't seem to copy the data (I tested this by providing the result and input to [shares_memory](https://numpy.org/doc/stable/reference/generated/numpy.shares_memory.html#numpy-shares-memory)) and the `NDCube` constructor triggers the data loading it seems like a good idea to include this `asanyarray` so that `NDCube` doesn't have to handle the `NDArrayType`. Does that sound good to you?

ndcube/asdf/converters/ndcube_converter.py
braingram: To clarify my :+1:

[np.array](https://numpy.org/doc/stable/reference/generated/numpy.array.html#numpy-array) by default will copy the data. Given the call arguments here I don't expect [np.asarray](https://numpy.org/doc/stable/reference/generated/numpy.asarray.html#numpy.asarray) will copy the data. asdf doesn't need a copy (it won't modify the data during conversion) so unless this is required by `NDCube` avoiding the copy would be more efficient.

ndcube/asdf/converters/ndcube_converter.py
ViciousEagle03: Yes, it makes sense, the loaded data is getting passed to the `NDCube` and it doesn't need to handle the loading behaviour of `NDArrayType`

ndcube/asdf/converters/ndcube_converter.py
braingram: Thanks for making the update! This is looking great so far. I hope you feel well soon. Please take my frequent comments as excitement about your work :)

ndcube/asdf/converters/ndcube_converter.py
ViciousEagle03: Yeah sure, then I suppose `np.asarray` would be a better choice.

ndcube/asdf/resources/schemas/NDCube-0.1.0.yaml
ViciousEagle03: I was experimenting and went with `$ref`, after reading the docs, I think it would be better to use a tag here. I believe `$ref` would be useful when we reuse the local schemas.

ndcube/asdf/converters/ndcube_converter.py
ViciousEagle03: Thanks for being so informative :)

ndcube/asdf/resources/schemas/ExtraCoords-0.1.0.yaml
```diff
@@ -0,0 +1,34 @@
...
+description:
+  Represents the ndcube ExtraCoords object
+
+type: object
+properties:
+  wcs:
+    tag: "tag:stsci.edu:gwcs/wcs-1.*"
+  mapping:
+    type: array
+  lookup_tables:
```
ViciousEagle03, line 18: I am unsure about the structure of the `lookup_tables` under the properties validator in the `ExtraCoords` schema and would need to test it with more examples. (I'll update it soon)

pyproject.toml
braingram: I don't think this has allowed the schema tests to run in the CI. I think it's related to https://github.com/asdf-format/asdf/pull/1756

The core issue is that the CI run is testing files in the installed package but is running from a [temporary directory](https://github.com/sunpy/ndcube/blob/ce4d0b13705fcb8128a87402a23ac0378e7c58da/tox.ini#L19). Take [this test](https://github.com/sunpy/ndcube/actions/runs/9323888973/job/25668021958?pr=708#step:10:110) as an example:
```
../../.tox/py312/lib/python3.12/site-packages/ndcube/extra_coords/tests/test_extra_coords.py::test_empty_ec PASSED
```
The path to the file contains a bunch of stuff that is very specific to that test run (py312, tox, etc) and this causes the asdf pytest plugin to ignore the schema files since they appear at a similar path which doesn't match the `asdf_schema_root`.

As there are test configurations where we might expect the schema tests to fail (like the [oldest deps job](https://github.com/sunpy/ndcube/blob/ce4d0b13705fcb8128a87402a23ac0378e7c58da/tox.ini#L22), see [this conversation](https://github.com/DKISTDC/dkist/pull/324#issuecomment-2023135345) for more details) we may want to consider adding a new CI job (or jobs) that test just the schemas. Minimally they could be tested against the newest released asdf and perhaps they could also be tested against development versions of asdf and the dependent schema packages. Ignoring the development testing at the moment I think this would involve:
- add a tox factor that runs just the schema tests (defining `asdf_schema_root`, providing a path to the schemas, not running in a temporary directory)
- enabling this factor in the [CI workflow](https://github.com/sunpy/ndcube/blob/ce4d0b13705fcb8128a87402a23ac0378e7c58da/.github/workflows/ci.yml#L54)

ViciousEagle03:
Hey @nabobalis , was there a problem in this branch?

nabobalis:
No I screwed up.

ndcube/asdf/converters/extracoords_converter.py
```diff
@@ -0,0 +1,29 @@
...
+
+
+class ExtraCoordsConverter(Converter):
+    tags = ["tag:sunpy.org:ndcube/extra_coords/extra_coords/ExtraCoords-*"]
+    types = ["ndcube.extra_coords.extra_coords.ExtraCoords"]
+
+    def from_yaml_tree(self, node, tag, ctx):
+        from ndcube.extra_coords.extra_coords import ExtraCoords
+        extra_coords = ExtraCoords()
+        extra_coords._wcs = node.get("wcs")
```
Cadair, line 11: I can't remember if wcs is dynamic if we have lookup tables, so we wouldn't need to serialise it in that case. I can check later.

pyproject.toml
ViciousEagle03: I've included a separate environment for schema testing in tox and created a workflow for it. Should I add a separate job for the schema tests as well? Please let me know. (P.S. I got a bit carried away with the workflow concept.)

tox.ini
```diff
@@ -89,6 +90,16 @@ commands =
     {toxinidir}/docs \
     {posargs}
 
+[testenv:schema]
+description = Run schema tests
+deps =
+    pytest
+    asdf
+set_env =
+    asdf_schema_root = {toxinidir}/ndcube/asdf/resources
```
braingram, line 99: Nice!

pyproject.toml
braingram: The workflow and tox changes you've included so far look good enough to me. I don't think adding a separate job is necessary (but please let me know if I'm missing something that would provide). Thanks!

ndcube/asdf/converters/extracoords_converter.py
ViciousEagle03: ExtraCoords can be initialized by a `~astropy.wcs.wcsapi.BaseLowLevelWCS` object and a `mapping`. Therfore, I included `wcs` and `mapping` as a node key-value pair and when we initialize ExtraCoords with the `lookup_tables`, the `wcs` is indeed dynamic, and we don't need to serialize it. The structure is such that only when we initialize it with a `wcs` object and a `mapping`, it will get serialized. Did I understand it correctly, or am I missing something?

Cadair: You are pretty much right, there are two internal representations possible with ExtraCoords, first you have the case where someone makes an EC object like this:
```python
ec = ExtraCoords()
ec.wcs = my_wcs
```
the other is when you incrementally build up a EC object from tables:
```python
ec = ExtraCoords()
ec.add("coord1", ...)
```

We need to serialise these two cases differently. You can see from the implementation of `ExtraCoords.wcs`: https://github.com/sunpy/ndcube/blob/main/ndcube/extra_coords/extra_coords.py#L287 how this works. So I think we need the same `if ec._wcs` and `if ec._lookup_tables` logic in here as well.

Cadair: Looking below I see this is how you have it anyway! Looks good to me.

ndcube/asdf/converters/globalcoords_converter.py
```diff
@@ -0,0 +1,28 @@
...
+        globalcoords._ndcube = node["ndcube"]
+
+        return globalcoords
+
+    def to_yaml_tree(self, globalcoords, tag, ctx):
+        node = {}
+        node["ndcube"] = globalcoords._ndcube
+        if globalcoords._internal_coords:
+            node["internal_coords"] = dict(globalcoords._internal_coords)
+        # Todo: Include `_all_coords` as a node key to preserve the dropped dimensions
```
Cadair, line 25: I don't think we do need to include all_coords, the ndcube object should have the same `SlicedLowLevelWCS` so the rest of `all_coords` will be dynamically generated as it is before we save it.

ndcube/asdf/entry_points.py
```diff
@@ -0,0 +1,48 @@
...
+    from ndcube.asdf.converters.extracoords_converter import ExtraCoordsConverter
+    from ndcube.asdf.converters.globalcoords_converter import GlobalCoordsConverter
+    from ndcube.asdf.converters.ndcube_converter import NDCubeConverter
+    from ndcube.asdf.converters.tablecoord_converter import (
+        QuantityTableCoordinateConverter,
+        SkyCoordTableCoordinateConverter,
+        TimeTableCoordConverter,
+    )
+
+    ndcube_converters = [NDCubeConverter(),ExtraCoordsConverter(),TimeTableCoordConverter(),QuantityTableCoordinateConverter(),SkyCoordTableCoordinateConverter(),GlobalCoordsConverter()]
```
Cadair, line 43: can you break this line up it's very long

ndcube/asdf/resources/schemas/NDCube-0.1.0.yaml
```diff
@@ -0,0 +1,25 @@
...
+    tag: "tag:stsci.edu:asdf/core/ndarray-1.*"
+  wcs:
+    tag: "tag:stsci.edu:gwcs/wcs-1.*"
+  extra_coords:
+    $ref: "asdf://sunpy.org/ndcube/schemas/ExtraCoords-0.1.0"
+  globa_coords:
+    $ref: "asdf://sunpy.org/ndcube/schemas/GlobalCoords-0.1.0"
+
+required: [data , wcs]
+allowAdditionalProperties: False
```
Cadair, line 24: If I subclass NDCube and want to write my own schema for my subclass with additional properties does this get in the way? @braingram

ndcube/asdf/resources/schemas/QuantityTableCoordinate-0.1.0.yaml
```diff
@@ -0,0 +1,29 @@
...
+    items:
+      tag: "tag:stsci.edu:asdf/unit/quantity-*"
+  names:
+    type: array
+  mesh:
+    type: boolean
+  physical_types:
+    type: array
+
+required: ["table"]
```
Cadair, line 27: I think we need to make unit required as well?

Cadair (commented):
This is coming on great @ViciousEagle03 thanks so much for all your work.

ndcube/asdf/converters/globalcoords_converter.py
ViciousEagle03: Ah, I see. Since I couldn't test it with the sliced ndcube, I was skeptical about whether to add all_coords or not.

ndcube/asdf/entry_points.py
ViciousEagle03: Sure

ndcube/asdf/resources/schemas/QuantityTableCoordinate-0.1.0.yaml
ViciousEagle03: Yes, I should add `unit` as a required property, thanks for pointing it out.

tox.ini
```diff
@@ -89,6 +90,16 @@ commands =
...
+commands=
```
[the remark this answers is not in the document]
ViciousEagle03: Whoop, let me correct that.

ndcube/asdf/resources/schemas/NDCube-0.1.0.yaml
braingram: I think so. A typical way to "subclass" a schema is to `$ref` the parent schema in a `allOf`. Using `allowAdditionalProperties: False` here would prevent the sub-schema from adding properties (because any addition would cause the parent schema validation to fail, causing the `allOf` to fail).

ndcube/asdf/resources/schemas/ExtraCoords-0.1.0.yaml
```diff
@@ -0,0 +1,37 @@
+%YAML 1.1
+---
+$schema: "http://stsci.edu/schemas/yaml-schema/draft-01"
+id: "asdf://sunpy.org/ndcube/schemas/ExtraCoords-0.1.0"
```
braingram, line 4: I suggest we stick with lowercase with `_` separators for schema uris. This helps to avoid issues with filenames vs uris with case-insensitive filesystems.
```suggestion
id: "asdf://sunpy.org/ndcube/schemas/extra_coords-0.1.0"
```

ndcube/asdf/resources/schemas/ExtraCoords-0.1.0.yaml
```diff
@@ -0,0 +1,37 @@
...
+  lookup_tables:
+    type: array
+    items:
+        type: array
+        items:
+            - oneOf:
+              - type: number
+              - type: array
+            - oneOf:
+              - $ref: "asdf://sunpy.org/ndcube/schemas/QuantityTableCoordinate-0.1.0"
```
braingram, line 27: Any reason to not use a `tag` validator here?

ndcube/asdf/resources/schemas/ExtraCoords-0.1.0.yaml
```diff
@@ -0,0 +1,37 @@
...
+              - $ref: "asdf://sunpy.org/ndcube/schemas/SkyCoordTableCoordinate-0.1.0"
+              - $ref: "asdf://sunpy.org/ndcube/schemas/TimeTableCoordinate-0.1.0"
+  dropped_tables:
+    type: array
+  ndcube:
+    tag: tag:sunpy.org:ndcube/ndcube/NDCube-0.1.0
```
braingram, line 33: We'll have to go through and update all of these once the schemas are "stable" at 1.0.0. We might as well start using the wildcard now.
```suggestion
    tag: tag:sunpy.org:ndcube/ndcube/ndcube-0.*
```

ndcube/asdf/resources/schemas/NDCube-0.1.0.yaml
```diff
@@ -0,0 +1,25 @@
...
+
+type: object
+properties:
+  data:
+    tag: "tag:stsci.edu:asdf/core/ndarray-1.*"
+  wcs:
+    tag: "tag:stsci.edu:gwcs/wcs-1.*"
+  extra_coords:
+    $ref: "asdf://sunpy.org/ndcube/schemas/ExtraCoords-0.1.0"
+  globa_coords:
```
Cadair, line 20:
```suggestion
  global_coords:
```

ndcube/asdf/resources/schemas/ExtraCoords-0.1.0.yaml
ViciousEagle03: Sorry, I couldn't find the reference documentation I mentioned earlier. I may have misread something about the $ref in the ASDF schema docs. I have replaced $ref with tags as we discussed.

braingram: Thanks for looking and for pointing out the gap in the documentation. I opened an issue for asdf:
https://github.com/asdf-format/asdf/issues/1791
Feel free to add to it if I missed anything.

ndcube/asdf/resources/schemas/ndcube-0.1.0.yaml
```diff
@@ -0,0 +1,25 @@
...
+    description: "Must be compatible with ASDF serialization/deserialization and supported by NDCube."
+  wcs:
+    tag: "tag:stsci.edu:gwcs/wcs-1.*"
+  extra_coords:
+    tag: "tag:sunpy.org:ndcube/extra_coords/extra_coords/extracoords-0.*"
+  global_coords:
+    tag: "tag:sunpy.org:ndcube/global_coords/globalcoords-0.*"
+
+required: [data , wcs]
+additionalProperties: true
```
ViciousEagle03, line 24: In ASDF how do the `allowAdditionalProperties` and `additionalProperties` differ? @braingram

braingram: The asdf schemas are based off of "draft 4" jsonschema which doesn't define any behavior for `allowAdditionalProperties` (so defining this won't do anything). It does define `additionalProperties` (in a somewhat confusing way):
https://json-schema.org/draft-04/draft-fge-json-schema-validation-00#rfc.section.5.4.4

ViciousEagle03: Ah, thanks for clarifying, I guess [this](https://github.com/sunpy/sunpy/blob/8ac75969de24ee5b03db61560ff7ed4265c0d560/sunpy/io/special/asdf/resources/schemas/generic_map-1.2.0.yaml#L29C1-L29C33) needs to be updated then?

braingram: Good catch, yeah a sunpy issue/PR to remove those is a good idea:
https://github.com/search?q=repo%3Asunpy%2Fsunpy%20allowAdditionalProperties&type=code

ViciousEagle03: Sure

ViciousEagle03: Opened PR #7686, which addresses the above issue.

ViciousEagle03:
> @ViciousEagle03 Would you add some tests to this PR that verify that a variety of `NDCube` objects can be written to and read from an asdf file? For some inspiration, here are some examples in `asdf-astropy` which generate a number of `astropy.Time` objects then perform a number of tests (including writing and reading the times to asdf): https://github.com/astropy/asdf-astropy/blob/main/asdf_astropy/converters/time/tests/test_time.py
> 
> As not all `NDCube` objects are currently supported, it would be helpful to have tests for the unsupported objects that check for helpful errors raised by the converter.


I've added GWCS objects and tests to check the roundtrip serialization of the NDCube. If I have missed something, please suggest any additional tests or improvements to improve the coverage.

ndcube/asdf/converters/tests/test_ndcube_converter.py
```diff
@@ -0,0 +1,43 @@
...
+def test_serialization(ndc, tmp_path):
+    file_path = tmp_path / "test.asdf"
+    with asdf.AsdfFile() as af:
+        af["ndcube_gwcs"] = ndc
+        af.write_to(file_path)
+
+    with asdf.open(file_path) as af:
+        assert_cubes_equal(af["ndcube_gwcs"], ndc)
+
+@pytest.mark.xfail(reason="Serialization of sliced ndcube not supported")
```
ViciousEagle03, line 24: @braingram I have used `@pytest.mark.xfail` to indicate that other `NDCube` objects are not supported for reading and writing to an ASDF file. If you have any suggestions on how to better inform about the unsupported objects, please let me know.

braingram: I think this is a great way to handle it.

braingram:
I'll try to review this today. My general impression is it looks great and I'm curious to hear thoughts from someone who has more (I should say any) real-world ndcube experience and can try saving and loading a few compatible ndcubes. @Cadair any suggestions for a suitable test-subject?

ndcube/asdf/converters/globalcoords_converter.py
```diff
@@ -0,0 +1,26 @@
...
+class GlobalCoordsConverter(Converter):
+    tags = ["tag:sunpy.org:ndcube/global_coords/globalcoords-*"]
+    types = ["ndcube.global_coords.GlobalCoords"]
+
+    def from_yaml_tree(self, node, tag, ctx):
+        from ndcube.global_coords import GlobalCoords
+
+        globalcoords = GlobalCoords()
+        if node.get("internal_coords") is not None:
+            globalcoords._internal_coords = OrderedDict(node.get("internal_coords"))
```
braingram, line 15:
```suggestion
        if "internal_coords" in node:
            globalcoords._internal_coords = node["internal_coords"]
```
See follow-up comment about `OrderedDict`

ndcube/asdf/converters/globalcoords_converter.py
```diff
@@ -0,0 +1,26 @@
+from typing import OrderedDict
+
```
braingram, line 2:
```suggestion
```
See comment below about `OrderedDict`.

ndcube/asdf/converters/globalcoords_converter.py
```diff
@@ -0,0 +1,26 @@
...
+            globalcoords._internal_coords = OrderedDict(node.get("internal_coords"))
+        globalcoords._ndcube = node["ndcube"]
+
+        return globalcoords
+
+    def to_yaml_tree(self, globalcoords, tag, ctx):
+        node = {}
+        node["ndcube"] = globalcoords._ndcube
+        if globalcoords._internal_coords:
+            node["internal_coords"] = dict(globalcoords._internal_coords)
```
braingram, line 24:
```suggestion
            node["internal_coords"] = globalcoords._internal_coords
```
`OrderedDict` is supported by asdf (via a yaml OMAP). Let me know if you ran into any issues trying to save a `OrderedDict`.

ndcube/asdf/converters/ndcube_converter.py
```diff
@@ -0,0 +1,26 @@
...
+class NDCubeConverter(Converter):
+    tags = ["tag:sunpy.org:ndcube/ndcube/ndcube-*"]
+    types = ["ndcube.ndcube.NDCube"]
+
+    def from_yaml_tree(self, node, tag, ctx):
+        from ndcube.ndcube import NDCube
+
+        data = node["data"]
+        wcs = node["wcs"]
+        ndcube = NDCube(data, wcs)
```
braingram, line 13:
```suggestion
        ndcube = NDCube(node["data"], node["wcs"])
```

ndcube/asdf/converters/tablecoord_converter.py
```diff
@@ -0,0 +1,85 @@
...
+
+
+class TimeTableCoordConverter(Converter):
+    tags = ["tag:sunpy.org:ndcube/extra_coords/table_coord/timetablecoordinate-*"]
+    types = ["ndcube.extra_coords.table_coord.TimeTableCoordinate"]
+
+    def from_yaml_tree(self, node, tag, ctx):
+        from ndcube.extra_coords.table_coord import TimeTableCoordinate
+
+        table = node.get("table")
```
braingram, line 11:
```suggestion
```
See related change below.

ndcube/asdf/converters/tablecoord_converter.py
```diff
@@ -0,0 +1,85 @@
...
+        names = node.get("names")
+        physical_types = node.get("physical_types")
+        reference_time = node.get("reference_time")
+        timetablecoordinate = TimeTableCoordinate(
+            table, names=names, physical_types=physical_types, reference_time=reference_time)
```
braingram, line 16:
```suggestion
            node["table"], names=names, physical_types=physical_types, reference_time=reference_time)
```
Since table is required by the schema we can be sure it exists and don't need the `get`.

ndcube/asdf/converters/tablecoord_converter.py
```diff
@@ -0,0 +1,85 @@
...
+
+        return timetablecoordinate
+
+    def to_yaml_tree(self, timetablecoordinate, tag, ctx):
+        node = {}
+        node["table"] = timetablecoordinate.table
+        node["names"] = timetablecoordinate.names
+        node["mesh"] = timetablecoordinate.mesh
```
braingram, line 24: Is this unused for this class?

ndcube/asdf/converters/tablecoord_converter.py
```diff
@@ -0,0 +1,85 @@
...
+
+class QuantityTableCoordinateConverter(Converter):
+    tags = ["tag:sunpy.org:ndcube/extra_coords/table_coord/quantitytablecoordinate-*"]
+    types = ["ndcube.extra_coords.table_coord.QuantityTableCoordinate"]
+
+    def from_yaml_tree(self, node, tag, ctx):
+        from ndcube.extra_coords.table_coord import QuantityTableCoordinate
+
+        unit = node.get("unit")
+        table = node.get("table")
```
braingram, line 40: table and unit are required by the schema so the `get` can be removed (similar to the above comment for `TimeTableCoordinate`).

ndcube/asdf/converters/tablecoord_converter.py
```diff
@@ -0,0 +1,85 @@
...
+
+
+class SkyCoordTableCoordinateConverter(Converter):
+    tags = ["tag:sunpy.org:ndcube/extra_coords/table_coord/skycoordtablecoordinate-*"]
+    types = ["ndcube.extra_coords.table_coord.SkyCoordTableCoordinate"]
+
+    def from_yaml_tree(self, node, tag, ctx):
+        from ndcube.extra_coords.table_coord import SkyCoordTableCoordinate
+
+        table = node.get("table")
```
braingram, line 69: Same as above comments about the required "table".

ndcube/asdf/resources/schemas/global_coords-0.1.0.yaml
```diff
@@ -0,0 +1,30 @@
...
+    type: object
+    patternProperties:
+      .*:
+        type: array
+        items:
+            - type: string
+            - type: object
+              oneOf:
+                - tag: "tag:stsci.edu:asdf/unit/quantity-*"
+                - tag: "tag:astropy.org:astropy/coordinates/skycoord-*"
```
braingram, line 24:
```suggestion
    additionalProperties:
      type: array
      items:
          - type: string
          - type: object
            oneOf:
              - tag: "tag:stsci.edu:asdf/unit/quantity-*"
              - tag: "tag:astropy.org:astropy/coordinates/skycoord-*"
```
Hopefully I didn't mess up the indenting with this suggestion. I think `additionalProperties` should work here.

ndcube/asdf/resources/schemas/ndcube-0.1.0.yaml
```diff
@@ -0,0 +1,25 @@
...
+  data:
+    description: "Must be compatible with ASDF serialization/deserialization and supported by NDCube."
+  wcs:
+    tag: "tag:stsci.edu:gwcs/wcs-1.*"
+  extra_coords:
+    tag: "tag:sunpy.org:ndcube/extra_coords/extra_coords/extracoords-0.*"
+  global_coords:
+    tag: "tag:sunpy.org:ndcube/global_coords/globalcoords-0.*"
+
+required: [data , wcs]
```
braingram, line 23:
```suggestion
required: [data, wcs]
```

braingram (commented):
I left some minor comments. Overall looks good to me! There are a few comments about converters and schemas I'd like to resolve before approving but I suspect the resolution will be quick. Nice work!

ndcube/asdf/converters/tablecoord_converter.py
ViciousEagle03: Sure

ndcube/asdf/converters/tablecoord_converter.py
ViciousEagle03: Ok

ndcube/asdf/converters/tablecoord_converter.py
ViciousEagle03: Sure

ndcube/asdf/converters/tablecoord_converter.py
ViciousEagle03: Sure

ndcube/asdf/converters/globalcoords_converter.py
ViciousEagle03: Thanks for letting me know about the Yaml OMAP.

ndcube/asdf/converters/globalcoords_converter.py
ViciousEagle03: Sure

ndcube/asdf/converters/globalcoords_converter.py
ViciousEagle03: Sure

ndcube/asdf/converters/extracoords_converter.py
```diff
@@ -0,0 +1,29 @@
...
+class ExtraCoordsConverter(Converter):
+    tags = ["tag:sunpy.org:ndcube/extra_coords/extra_coords/extracoords-*"]
+    types = ["ndcube.extra_coords.extra_coords.ExtraCoords"]
+
+    def from_yaml_tree(self, node, tag, ctx):
+        from ndcube.extra_coords.extra_coords import ExtraCoords
+        extra_coords = ExtraCoords()
+        extra_coords._wcs = node.get("wcs")
+        extra_coords._mapping = node.get("mapping")
+        extra_coords._lookup_tables = node.get("lookup_tables")
```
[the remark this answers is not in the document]
ViciousEagle03: Thanks for pointing this out

ndcube/asdf/resources/schemas/global_coords-0.1.0.yaml
ViciousEagle03: Right, `additionalProperties` would be more appropriate here.

ndcube/asdf/converters/tablecoord_converter.py
ViciousEagle03: I have updated the test suite

ndcube/conftest.py
```diff
@@ -318,6 +456,57 @@ def extra_coords_sharing_axis():
...
+@pytest.fixture
+def ndcube_gwcs_3d_ln_lt_l_ec_q_t_gc(gwcs_3d_lt_ln_l):
+    shape = (3, 3, 4)
+    gwcs_3d_lt_ln_l.array_shape = shape
+    data_cube = data_nd(shape)
+    cube =  NDCube(data_cube, wcs=gwcs_3d_lt_ln_l)
+    coord1 = 1 * u.m
+    cube.global_coords.add('name1', 'custom:physical_type1', coord1)
+    cube.extra_coords.add("time", 0, time_lut(shape))
+    cube.extra_coords.add("exposure_lut", 1, range(shape[1]) * u.s)
```
ViciousEagle03, line 501: I have not added the extra_coords with `SkyCoordTableCoordinate` because there seems to be unexpected behavior when using `assert_extra_coords_equal` to verify the roundtrip serialization.

ndcube/asdf/converters/tablecoord_converter.py
braingram: Would you point me to where "mesh" is read? I don't see it in `from_yaml_tree` for this converter.

ndcube/asdf/converters/tablecoord_converter.py
braingram:
```suggestion
        unit = node["unit"]
        table = node["table"]
```

ndcube/asdf/converters/tablecoord_converter.py
braingram:
```suggestion
        table = node["table"]
```

ndcube/asdf/converters/extracoords_converter.py
```diff
@@ -0,0 +1,29 @@
...
+        extra_coords._dropped_tables = node.get("dropped_tables")
```
braingram:
```suggestion
        extra_coords._lookup_tables = node.get("lookup_tables", [])
        extra_coords._dropped_tables = node.get("dropped_tables", [])
```

ndcube/asdf/converters/globalcoords_converter.py
braingram:
```suggestion
```

ndcube/asdf/converters/globalcoords_converter.py
braingram:
```suggestion
        if "internal_coords" in node:
            globalcoords._internal_coords = node["internal_coords"]
```

ndcube/asdf/converters/globalcoords_converter.py
braingram:
```suggestion
            node["internal_coords"] = globalcoords._internal_coords
```

ndcube/asdf/converters/tablecoord_converter.py
braingram:
```suggestion
        table = node["table"]
```

braingram (changes requested):
~I'm a bit confused. Several of the suggestions in my previous comments were marked as resolved with responses that gave me the impression they were "ok" yet the changes were not made. I was going through updating comments but stopped when I realized this PR might not be up-to-date.~

~Are there changes not committed that have the previous suggestions? If not, is there a reason these suggestions were dismissed?~

EDIT: I'm not sure what's up with github but me posting this comment caused the PR to update for me :-/ I'm now seeing the changes made in [cd31d95](https://github.com/sunpy/ndcube/pull/708/commits/cd31d95a1f59ddba2c0b00f27f047155ae5b604c) which I did not previously see.

ndcube/asdf/resources/schemas/ndcube-0.1.0.yaml
```diff
@@ -0,0 +1,25 @@
...
+    description: "Must be compatible with ASDF serialization/deserialization and supported by NDCube."
+  wcs:
+    tag: "tag:stsci.edu:gwcs/wcs-1.*"
+  extra_coords:
+    tag: "tag:sunpy.org:ndcube/extra_coords/extra_coords/extracoords-0.*"
+  global_coords:
+    tag: "tag:sunpy.org:ndcube/global_coords/globalcoords-0.*"
+
+required: [data, wcs]
+additionalProperties: true
```
braingram, line 24: One question about this. What's the reasoning for having this `true` where it's `false` in the other schemas?

braingram (changes requested):
Now that I'm looking at the newest code :facepalm: 

Just a few comments/questions and one request.

Would you add asdf to the dependencies and set a version that makes the oldestdeps job happy?

ndcube/asdf/resources/schemas/ndcube-0.1.0.yaml
ViciousEagle03: We had discussed during the video chat that setting the `additionalProperties` to true would prevent conflicts when subclassing NDCube and writing the schema for the subclass.
Here is the related [comment](https://github.com/sunpy/ndcube/pull/708#:~:text=If%20I%20subclass%20NDCube%20and%20want%20to%20write%20my%20own%20schema%20for%20my%20subclass%20with%20additional%20properties%20does%20this%20get%20in%20the%20way%3F%20%40braingram)

ViciousEagle03:
> Now that I'm looking at the newest code 🤦
> 
> Just a few comments/questions and one request.
> 
> Would you add asdf to the dependencies and set a version that makes the oldestdeps job happy?

Sure, I am on it.

ndcube/asdf/resources/schemas/ndcube-0.1.0.yaml
braingram: Thanks! I forgot about that conversation. Looks good to me!

pyproject.toml
```diff
@@ -16,9 +16,10 @@ authors = [
   { name = "The SunPy Community", email = "sunpy@googlegroups.com" },
 ]
 dependencies = [
-  "astropy>=5.0.6,!=5.1.0",
-  "gwcs>=0.18",
-  "numpy>=1.23.0"
+  "astropy>=5.3",
+  "gwcs>=0.20",
+  "numpy>=1.23.0",
+  "asdf>=2.14.4"
```
ViciousEagle03, line 22: Apologies for the delay @braingram . To ensure the `gwcs` tests pass, we need to use at least version 0.20 of `gwcs` because it includes the necessary [information](https://github.com/spacetelescope/gwcs/releases/tag/0.20.0#:~:text=Serialize%20and%20deserialize%20pixel_shape%20with%20asdf.%20%5B%23480%5D)( The addition of pixel_shape to the wcs schema used during the roundtrip serialization) introduced in that version. `gwcs` version 0.20 requires a minimum of `astropy` version 5.3, and the `asdf` version that passes the tests is at least 2.14.4.

braingram (approved):
Looks good to me. Thanks!

pyproject.toml
```diff
@@ -16,9 +16,10 @@ authors = [
   { name = "The SunPy Community", email = "sunpy@googlegroups.com" },
 ]
 dependencies = [
-  "astropy>=5.0.6,!=5.1.0",
-  "gwcs>=0.18",
-  "numpy>=1.23.0"
+  "astropy>=5.3",
+  "gwcs>=0.20",
+  "numpy>=1.23.0",
```
Cadair, line 21: These are some pretty aggressive version bumps and ndcube nominally follows the [SPEC 0](https://scientific-python.org/specs/spec-0000/) policy of supporting dependencies for two years after their release.

If we really can't get the asdf stuff working without them, one option would be to have stricter version requirements in the `asdf` extra so doing `pip install ndcube[asdf]` would require newer versions of things than `pip install ndcube`

tox.ini
```diff
@@ -90,6 +91,16 @@ commands =
     {toxinidir}/docs \
     {posargs}
 
+[testenv:schema]
```
Cadair, line 94:
```suggestion
[testenv:asdf_schemas]
```

Cadair (commented):
This PR looks amazing @ViciousEagle03 thanks for all your work on it. I have a few small-ish comments and then I think we can merge it.

ndcube/asdf/converters/extracoords_converter.py
```diff
@@ -0,0 +1,28 @@
...
+    types = ["ndcube.extra_coords.extra_coords.ExtraCoords"]
+
+    def from_yaml_tree(self, node, tag, ctx):
+        from ndcube.extra_coords.extra_coords import ExtraCoords
+        extra_coords = ExtraCoords()
+        extra_coords._wcs = node.get("wcs")
+        extra_coords._mapping = node.get("mapping")
+        extra_coords._lookup_tables = node.get("lookup_tables", [])
+        extra_coords._dropped_tables = node.get("dropped_tables")
+        extra_coords._ndcube = node.get("ndcube")
```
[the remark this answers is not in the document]
braingram: Theoretically yes but I believe the converter will need to be updated.
https://asdf.readthedocs.io/en/latest/asdf/extending/converters.html#reference-cycles
That seems like a good test case.

tox.ini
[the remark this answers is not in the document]
ViciousEagle03: The schema tests are ignored when running from a temporary directory due to path mismatches. 
A tox env was added so that the `asdf_schema_root` was set to the correct path and was not run from a `.tmp` directory which ensures the schema is located and tested properly.
[Relevant chat](https://github.com/sunpy/ndcube/pull/708/files#r1624495728)

ndcube/asdf/converters/extracoords_converter.py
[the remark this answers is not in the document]
ViciousEagle03: @braingram, for the current implementation I believe it is still not storing the `NDCube` object twice, right?
And also since the current `from_yaml_tree` doesn't have `yield`, it can still deserialize the extra_coords object, that is, when the extra_coords converter is called the extra_coords.ndcube shows `_PendingValue` and when it gets picked up in the NDCube converter and the `ndcube.extra_coords` is initialized the reference to the ndcube is properly mapped in the `ndcube.extra_coords._ndcube`.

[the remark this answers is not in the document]
braingram: Thanks for testing this out. This looks to be working because `__set__` overwrites `_ndcube` when `_extra_coords` is assigned:
https://github.com/sunpy/ndcube/blob/09a83f14c704e1428bc931a3e02d4228a92743a6/ndcube/ndcube.py#L321
I am not sure what's going on with `__set__`. @Cadair is there ever an instance where:

- the `extra_coords` attached to a `NDCube` will reference a different `NDCube`?
- `extra_coords` will exist without a `NDCube`?

braingram:
One other minor request. Would you rename the `asdf` directory to `_asdf` (to make it more explicit that this is private API)?

ndcube/asdf/converters/ndcube_converter.py
```diff
@@ -0,0 +1,24 @@
...
+        ndcube._global_coords = node["global_coords"]
+
+        return ndcube
+
+    def to_yaml_tree(self, ndcube, tag, ctx):
+        node = {}
+        node["data"] = ndcube.data
+        node["wcs"] = ndcube.wcs
+        node["extra_coords"] = ndcube.extra_coords
+        node["global_coords"] = ndcube.global_coords
```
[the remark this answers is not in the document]
ViciousEagle03: I have added warnings for the `NDCube` parameters that wouldn't be serialized to `ASDF`.

ViciousEagle03:
Apologies for the delay, @nabobalis would you mind taking a look? As I mentioned earlier, the `oldestdeps` is failing.

nabobalis:
> Apologies for the delay, @nabobalis would you mind taking a look? As I mentioned earlier, the `oldestdeps` is failing.

It is creating a warning which is failing the CI:

`AsdfWarning: Converter handles multiple tags for this extension, but does not implement a select_tag method. This previously worked because Converter subclasses inherited the now removed select_tag. This will be an error in a future version of asdf`

I would just ignore the warning being created by adding it to the pytest.ini list.

ViciousEagle03:
> > Apologies for the delay, @nabobalis would you mind taking a look? As I mentioned earlier, the `oldestdeps` is failing.
> 
> It is creating a warning which is failing the CI:
> 
> `AsdfWarning: Converter handles multiple tags for this extension, but does not implement a select_tag method. This previously worked because Converter subclasses inherited the now removed select_tag. This will be an error in a future version of asdf`
> 
> I would just ignore the warning being created by adding it to the pytest.ini list.

Hey @nabobalis, thanks for the help. Even after adding the `asdf.exceptions.AsdfWarning`, some of the tests are still failing due to `AssertionError`, as the pixel_shape serialization of `gWCS` was [added](https://github.com/sunpy/ndcube/pull/708#:~:text=The%20addition%20of%20pixel_shape%20to%20the%20wcs%20schema%20used%20during%20the%20roundtrip%20serialization) in version `0.20`. I wonder if it might be simpler to skip the tests if the `gWCS` version is `<0.20`, rather than adding asdf as an optional dependency.

ndcube/asdf/converters/ndcube_converter.py
```diff
@@ -0,0 +1,50 @@
...
+
+        attributes = ['uncertainty', 'mask', 'unit']
+        for attr in attributes:
+            if getattr(ndcube, attr) is not None:
+                warnings.warn(f"Attribute '{attr}' is present but not being saved in ASDF serialization.", UserWarning)
+
+        if len(ndcube.meta) > 0:
```
Cadair, line 47: I think we should be able to save meta without any other support, until #455 gets merged that is. Meta at the moment is just a dict of any contents. If someone has a very complex meta dict, so some of it can't be saved to ASDF then I think it's fair to either let them add tags or just pop those things from the meta first.

ndcube/asdf/converters/ndcube_converter.py
```diff
@@ -0,0 +1,50 @@
...
+            Warns if the NDCube object has attributes 'uncertainty', 'mask', 'meta',
+            or 'unit' that are present but not being saved in the ASDF serialization.
+            This ensures that users are aware of potentially important information
+            that is not included in the serialized output.
+        """
+        node = {}
+        node["data"] = ndcube.data
+        node["wcs"] = ndcube.wcs
+        node["extra_coords"] = ndcube.extra_coords
+        node["global_coords"] = ndcube.global_coords
```
Cadair, line 40:
```suggestion
        node["global_coords"] = ndcube.global_coords
        node["meta"] = ndcube.meta
```

ndcube/asdf/converters/ndcube_converter.py
```diff
@@ -0,0 +1,50 @@
...
+
+
+class NDCubeConverter(Converter):
+    tags = ["tag:sunpy.org:ndcube/ndcube/ndcube-*"]
+    types = ["ndcube.ndcube.NDCube"]
+
+    def from_yaml_tree(self, node, tag, ctx):
+        from ndcube.ndcube import NDCube
+
+        ndcube = NDCube(node["data"], node["wcs"])
```
Cadair, line 13:
```suggestion
        ndcube = NDCube(node["data"], node["wcs"], meta=node.get("meta"))
```

ndcube/asdf/converters/ndcube_converter.py
```diff
@@ -0,0 +1,50 @@
...
+        ndcube._extra_coords = node["extra_coords"]
+        ndcube._global_coords = node["global_coords"]
```
Cadair, line 15: These two are not required in the schema, so I think they should be optional:
```suggestion
        if "extra_coords" in node:
            ndcube._extra_coords = node["extra_coords"]
        if "global_coords" in node:
            ndcube._global_coords = node["global_coords"]
```

ndcube/asdf/converters/extracoords_converter.py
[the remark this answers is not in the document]
Cadair: >    the `extra_coords` attached to a `NDCube` will reference a different `NDCube`?

No, that should be explicitly prohibited by the descriptor.

>    `extra_coords` will exist without a `NDCube`?

I think this might be possible yes.

ndcube/asdf/resources/schemas/ndcube-0.1.0.yaml
```diff
@@ -0,0 +1,25 @@
...
+properties:
+  data:
+    description: "Must be compatible with ASDF serialization/deserialization and supported by NDCube."
+  wcs:
+    tag: "tag:stsci.edu:gwcs/wcs-1.*"
+  extra_coords:
+    tag: "tag:sunpy.org:ndcube/extra_coords/extra_coords/extracoords-0.*"
+  global_coords:
+    tag: "tag:sunpy.org:ndcube/global_coords/globalcoords-0.*"
+
```
Cadair, line 22: I think we can add meta support:
```suggestion
  meta:
    type: object

```

Cadair (approved):
This looks great to me. All my comments are minor.

ndcube/asdf/converters/extracoords_converter.py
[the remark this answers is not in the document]
braingram: > > `extra_coords` will exist without a `NDCube`?
> 
> I think this might be possible yes.

Is there a minimal example that would show this? Perhaps it can be used for a test case for the converter.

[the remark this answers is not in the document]
Cadair: I think you can just instantiate one standalone and then add some coordinates to it with the same API as you can if it's attached to the cube.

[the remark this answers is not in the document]
Cadair: turns out you can't: https://github.com/sunpy/ndcube/issues/753

Cadair (approved):

PR #710, opened 2024-05-25
Coords Class for NDCube

DanRyanIrish:
## PR Description

This PR is a sandbox for an idea for a `NDCoords` class linked to `NDCube` through which various coordinate-associated methods and properties can be accessed.

It is envisioned to be attached to `NDCube` as a property or attribute `NDCube.coords`.


hayesla:
in theory, i think this is a good idea, and will be very helpful for users.

PR #713, opened 2024-05-29
Make TableCoordinates support pixel edges and centers

DanRyanIrish:
## PR Description

This PR enables support for custom pixel edges in `QuantityTableCoordinate` and `TimeTableCoordinate` objects by storing the underlying grids as including pixel edges and centers, not just centers as heretonow.

The motivation is that world coordinates at pixel edges are often wanted.  When the pixel widths are uniform, the edges can be found from the centers.  However, when this is not the case, explicit knowledge of the edges is required.


DanRyanIrish:
@Cadair Can you advise on whether this is the best way to go about this?

DanRyanIrish:
Make `TableCoordinate` require the input pixel grid to which the input world grid corresponds, and then keep that as the underlying storage.  Enable shortcuts for `"center"` and `"corners"`.

Issue #725, opened 2024-06-12
Create an analysis subpackage


DanRyanIrish:
### Describe the feature

This issue proposes a new `analysis` subpackage for functions that take `NDCubeABC` instances and perform some analysis task.  The scope of this subpackage should be as follows:
* functions must only require the NDCube 2 API in order to carry out the entire task;
* functions do not have to be valid for all possible `NDCube` instances, but they must be general enough to be beyond the scope of an astronomy-sub-field-specific package.  For example:
  * fitting a spectral line is beyond scope because it requires an understanding of a model and a fitter API which is not included in the NDCube 2 API;
  * Differentially rotating an image of the Sun is beyond scope because this is specific to the field of solar physics;
  * cropping just the celestial axes of a cube is within scope because, although it only applies to cubes with celestial axes, the task can be performed exclusively using the NDCube 2 API and is useful across multiple sub-fields of astronomy.

The subpackage should contain modules which encompass certain types of tasks.  As a start, this issue proposes the following modules:
* `ndcube.analysis.rebin`
    * This can hold a `rebin` function equivalent to `NDCube.rebin`, but can also hold a function to irregularly rebin a cube, which, due to technical limitations at this time, can only be performed when all WCS axes are independent.
* `ndcube.analysis.crop`
  * This can hold a `crop` function equivalent to `NDCube.crop`, but can also contain functions to only crop the celestial axes, e.g. `crop_celestial`. The advantage of this is that more user-friendly APIs can be exposed to the user.  The same could be done for `crop_temporal`, `crop_spectral` etc.

## `NDCube` methods vs. analysis functions
Similar to `numpy`, `NDCube` can have methods that simply call an analysis function, e.g., `rebin` and `crop`.  But when should `NDCube` have a method as well as an analysis function?  The criterion for a method to exist should be that:
* Methods should apply to all possible `NDCube` instances, i.e. it should be completely agnostic to the number of dimensions and physical types associated with them.  For example:
  * `NDCube.crop` should exist, even if `ndcube.analysis.crop.crop()` also exists.  However, while `ndcube.analysis.crop.crop_celestial` is a valid anaylsis function, it should not be a method because it is only valid for cubes with celestial axes.

nabobalis:
Isn't this what specutils is but for 1D?

PR #726, opened 2024-06-12
Irregular rebinning function and Analysis Subpackage

DanRyanIrish:
## PR Description

Following the structure outline by #725, this PR creates an `analysis` subpackage and a `rebin` module.  Included in the `rebin` module is the implementation of the `rebin` method, which now simply calls the `rebin` function.

However, crucially, it also includes a `rebin_by_edges` function, which allows users to rebin an `NDCube` which irregular bin sizes along an axis.  This is a very important functionality for certain analyses, e.g. solar X-ray imaging spectroscopy.

Due to technical limitations, `ndcube` can currently only support this if the WCS axes are independent, and so is not within the scope of an `NDCube` method.  However, if this limitation is overcome in the future, an `NDCube.rebin_by_edges` method can be implemented as a wrapper around this function.

To Do
- [ ] Restore function signature to `NDCube.rebin`.  Explicitly list out inputs, rather than using `*args` and `**kwargs``.
- [ ] Incorporate feature request in #737 
- [ ] Write tests for `rebin_by_edges`
- [ ] Fix doc test
- [ ] Currently, wcs and extra coords are lumped together into combined wcs. Separate these so output still has same coords in WCS and extra_coords

To Consider
- [ ] Change name of `rebin_by_edges` to `rebin_by_corners`


settwi:
Hey Dan this looks really good. I especially like the option to propagate uncertainties (and presumably other metadata, too?). It will save us headaches in other applications.

Occasionally it is nice to do some interpolation of bins, too, like when binning a spectral response down from fine photon bins to coarser count bins, or to "directly" compare data from different instruments with different energy bins. For histograms you want to conserve the "flux" in each bin and the simplest way to do it is just a proportional interpolation. For photon count or flux data I reckon this approach makes sense.

I made a [gist](https://gist.github.com/settwi/60eb4fa884ff9a61cb0d081c93bbb714) a while back to do an interpolating rebinning which has proved useful a few times. Might be nice to add a similar functionality to ndcube. Maybe "rebin_interpolated" or something like that. "Normal" rebinning is just a subset of the interpolating version, but it might make sense to specialize the implementations to save on computation time.

samaloney (commented):
This is GREAT! I still need to actually try it out but just figured get these down now.

ndcube/analysis/rebin.py
```diff
@@ -0,0 +1,413 @@
...
+ARRAY_MASK_MAP[np.ndarray] = np.ma.masked_array
+_NUMPY_COPY_IF_NEEDED = False if np.__version__.startswith("1.") else None
+try:
+    import dask.array
+    ARRAY_MASK_MAP[dask.array.core.Array] = dask.array.ma.masked_array
+except ImportError:
+    pass
+
+
+def rebin(cube, bin_shape, operation=np.mean, operation_ignores_mask=False, handle_mask=np.all,
```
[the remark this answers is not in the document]
DanRyanIrish: It makes sense once there's a `rebin` module.  And it makes the `NDCube` class code base much cleaner as all it does is call this function.

ndcube/analysis/rebin.py
```diff
@@ -0,0 +1,413 @@
...
+        Default=`numpy.mean`.
+
+    Returns
+    -------
+    :
+
+    Example
+    -------
+    >>> import numpy as np
+    >>> from stixpy.product.tools import rebin_by_edges
```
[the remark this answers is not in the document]
DanRyanIrish:
```suggestion
    >>> from ndcube.analysis.rebin import rebin_by_edges
```

DanRyanIrish:
Hi @settwi.  Thanks for the comments.  I definitely agree with the use-cases you pointed out.  From a technical perspective, I __think__ they are already served by `NDCube.reproject_to`, assuming you generate the new WCS separately, which, I grant, can be a more cumbersome API.  I am right though?  Is there a difference between `rebin_interpolated` and `reproject_to` other than a more restrictive and hence more friendly API?

ndcube/analysis/rebin.py
```diff
@@ -0,0 +1,413 @@
+import numpy as np
+
+import astropy.nddata
+
+from ndcube import ExtraCoords, utils
+from ndcube.utils.exceptions import warn_user
+from ndcube.wcs.wrappers import ResampledLowLevelWCS
+
+__all__ = ["rebin"]
```
DanRyanIrish, line 9:
```suggestion
__all__ = ["rebin", "rebin_by_edges"]
```

DanRyanIrish:
We should consider whether `rebin_by_edges` should be `rebin_by_corners` to be consistent with the `axis_world_coords(corners=True)` API.

ndcube/analysis/rebin.py
```diff
@@ -0,0 +1,413 @@
...
+
+    # Build new WCS.
+    ec = ExtraCoords()
+    for name, axis, coord, physical_type in zip(
+        new_coord_names, new_coord_pix_idxs, new_coord_values, new_coord_physical_types
+    ):
+        if len(axis) != 1:
+            raise NotImplementedError("Rebinning multi-dimensional coordinates not supported.")
+        ec.add(name, axis, coord, physical_types=physical_type)
+    new_wcs = ec.wcs
```
Cadair, line 368: Using extra coords as a way to do this is a bit "weird" using table coordinate directly might be more applicable.

DanRyanIrish: Would this require extra code to parse whether it should be a `QuantityTableCoordinate`, `TimeTableCoordinate`, etc.?  If so, `ExtraCoords` already does this for us.

PR #729, opened 2024-06-17
Add tests for numpy dev and update for numpy 2

nabobalis:
Needs backporting and probably a release.

### DIFF
--- a/docs/explaining_ndcube/coordinates.rst
+++ b/docs/explaining_ndcube/coordinates.rst
@@ -225,7 +225,7 @@ The physical types of extra coordinates are also returned by `~ndcube.NDCube.arr
 .. code-block:: python
 
   >>> my_cube.array_axis_physical_types
-  [(np.str_('custom:pos.helioprojective.lat'), np.str_('custom:pos.helioprojective.lon'), np.str_('time')), (np.str_('custom:pos.helioprojective.lat'), np.str_('custom:pos.helioprojective.lon')), (np.str_('em.wl'),)]
+  [('custom:pos.helioprojective.lat', 'custom:pos.helioprojective.lon', 'time'), ('custom:pos.helioprojective.lat', 'custom:pos.helioprojective.lon'), ('em.wl',)]
 
 The values of the extra coordinates at each array index can be retrieved using and combination of :meth:`ndcube.NDCube.axis_world_coords` and `ndcube.NDCube.combined_wcs`.
 See :ref:`combined_wcs` below.
@@ -312,8 +312,7 @@ The values of dropped coordinates at the position where the `~ndcube.NDCube` was
 
   >>> my_2d_cube = my_cube[:, :, 0]
   >>> my_2d_cube.array_axis_physical_types  # Note the wavelength axis is now gone.
-  [(np.str_('custom:pos.helioprojective.lat'), np.str_('custom:pos.helioprojective.lon'), np.str_('time')), (np.str_('custom:pos.helioprojective.lat'), np.str_('custom:pos.helioprojective.lon'))]
-
+  [('custom:pos.helioprojective.lat', 'custom:pos.helioprojective.lon', 'time'), ('custom:pos.helioprojective.lat', 'custom:pos.helioprojective.lon')]
   >>> # The wavelength value at the slicing location is now in the GLobalCoords object.
   >>> list(my_2d_cube.global_coords.keys())
   ['distance', 'em.wl']
--- a/docs/explaining_ndcube/coordinates.rst
+++ b/docs/explaining_ndcube/coordinates.rst
@@ -225,7 +225,7 @@ The physical types of extra coordinates are also returned by `~ndcube.NDCube.arr
 .. code-block:: python
 
   >>> my_cube.array_axis_physical_types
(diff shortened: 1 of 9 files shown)


tox.ini
```diff
@@ -110,26 +111,3 @@ deps =
 commands =
     pre-commit install-hooks
     pre-commit run --color always --all-files --show-diff-on-failure
-
-# This env requires tox-conda.
```
nabobalis, line 114: Not sure we use this anymore for anything?

nabobalis:
I also imagine the tests will fail on oldest deps now.

How do we want to handle that?

docs/explaining_ndcube/coordinates.rst
```diff
@@ -225,7 +225,7 @@ The physical types of extra coordinates are also returned by `~ndcube.NDCube.arr
 .. code-block:: python
 
   >>> my_cube.array_axis_physical_types
-  [('custom:pos.helioprojective.lat', 'custom:pos.helioprojective.lon', 'time'), ('custom:pos.helioprojective.lat', 'custom:pos.helioprojective.lon'), ('em.wl',)]
```
Cadair, line 228: This is our code, so I think rather than doing this we should just fix the code here:

https://github.com/sunpy/ndcube/blob/62bb3693046470b2db1ce39f7241144973789a8f/ndcube/ndcube.py#L428-L429

pytest.ini
```diff
@@ -49,3 +49,5 @@ filterwarnings =
     # This is raised by the Windows and mac os build for visualization.rst
     ignore:FigureCanvasAgg is non-interactive, and thus cannot be shown:UserWarning
 remote_data_strict = True
+doctest_subpackage_requires =
```
Cadair, line 52: oh nice.

tox.ini
```diff
@@ -44,10 +44,11 @@ deps =
...
     # Astropy is installed from the nightly wheels
     devdeps: astropy>=0.0dev0
-    devdeps: matplotlib>=0.0dev0
-    devdeps: git+https://github.com/sunpy/sunpy
-    devdeps: git+https://github.com/sunpy/mpl-animators
     devdeps: git+https://github.com/spacetelescope/gwcs
+    devdeps: git+https://github.com/sunpy/mpl-animators
+    devdeps: git+https://github.com/sunpy/sunpy
+    devdeps: matplotlib>=0.0dev0
+    devdeps: numpy>=0.0.dev0
```
Cadair, line 51: why must you do these things :stuck_out_tongue:

Cadair (changes requested):
We can make `array_axis_physical_types` better, we don't need to change all the doctests.

ndcube/wcs/wrappers/tests/test_resampled_wcs.py
```diff
@@ -17,7 +17,29 @@ def celestial_wcs(request):
     return request.getfixturevalue(request.param)
 
 
-EXPECTED_2D_REPR = """
+EXPECTED_2D_REPR_NUMPY2 = """
```
Cadair, line 20: I opened https://github.com/astropy/astropy/issues/16585 to track this.

nabobalis: Do you want me to undo this and wait for astropy?

tox.ini
nabobalis: I just sorted it in order of letter

docs/explaining_ndcube/coordinates.rst
nabobalis: Oh it's code in ndcube, I assumed it was astropy. 

I can change the code here instead then.

ndcube/wcs/wrappers/tests/test_resampled_wcs.py
Cadair: There is now an astropy PR, so maybe? :thinking:

tox.ini
Cadair: It makes it marginally less likely to conflict with the template if we keep the things that are in the template at the top. This is true for things like the [pyproject.toml](https://github.com/sunpy/streamtracer/pull/156/files#diff-50c86b7ed8ac2cf95bd48334961bf0530cdc77b5a56f852c5c61b89d735fd711L54-R57) as well.

nabobalis: The template should be in order as well

ndcube/ndcube.py
```diff
@@ -425,7 +425,7 @@ def array_axis_physical_types(self):
         wcs = self.combined_wcs
         world_axis_physical_types = np.array(wcs.world_axis_physical_types)
         axis_correlation_matrix = wcs.axis_correlation_matrix
-        return [tuple(world_axis_physical_types[axis_correlation_matrix[:, i]])
+        return [tuple(world_axis_physical_types[axis_correlation_matrix[:, i]].tolist())
```
nabobalis, line 428: This seemed to work.

ndcube/wcs/wrappers/tests/test_resampled_wcs.py
nabobalis: But we should still need this for oldest deps?

Cadair (approved):

Issue #714, opened 2024-05-29
Allow crop to crop to a single pixel


hayesla:
### Describe the bug

Philosophically this is wrong right? you should  be able to crop to a single pixel (@DanRyanIrish  says so too ;) )

https://github.com/sunpy/ndcube/blob/main/ndcube/utils/cube.py#L200

related to this https://github.com/sunpy/sunpy/pull/7605

DanRyanIrish:
The philosophy of `crop` is that the resulting cube should include all input points.  (This is different to slicing, which excludes the upper edge of the interval.)  Therefore, it's logically consistent for `crop` to return a shape-1 cube if the inputs are all in the same pixel.  So instead of raising an error, the code quoted above should be something like:
```python
if result_is_scalar:
    item = tuple(slice(idx, idx + 1) for idx in item)
```

Two things to consider here:
1. The above special case would mean that all dimensions are kept, they are just length-1.  Normally, length-1 dimensions are dropped because it is more convenient to users.
2. We could institute a rule that, like `rebin`, `crop` never drops dimensions.  Length-1 dimensions are the norm.  This will be more cumbersome in some cases, but the behaviour will be consistent.  And now that there is a `squeeze()` method, it's easy to subsequently drop all length-1 dimensions.

I think either case is viable.  It's a philosophical decision.  Does anyone have thoughts?

DanRyanIrish:
A compromise to avoid this being a breaking change is to add a `keepdims=False` kwarg to crop which means no dimensions are dropped.  At least this would mean that cropping to a single pixel with `keepdims=True` would work.

PR #732, opened 2024-06-20
Add a keepdims kwarg to crop and crop_by_value to keep length-1 dimensions

samaloney:
Add a `keep_dims` kwarg to `crop` and `crop_by_values` to allowing length-1 dimensions to be kept see #714 

- [x] Docs
   - [x] Doc strings
   - [x] Main docs (https://sunpy--732.org.readthedocs.build/projects/ndcube/en/732/explaining_ndcube/slicing.html#cropping-with-real-world-coordinates)
- [x] Change log
- [x] Tests

### DIFF
--- a/ndcube/ndcube.py
+++ b/ndcube/ndcube.py
@@ -186,7 +186,8 @@ class NDCubeABC(astropy.nddata.NDDataBase):
     @abc.abstractmethod
     def crop(self,
              *points: Iterable[Any],
-             wcs: BaseHighLevelWCS | ExtraCoordsABC | None = None
+             wcs: BaseHighLevelWCS | ExtraCoordsABC | None = None,
+             keepdims: bool = False,
              ) -> "NDCubeABC":
         """
         Crop using real world coordinates.
@@ -215,6 +216,9 @@ class NDCubeABC(astropy.nddata.NDDataBase):
             could be used it is expected that either the ``.wcs`` or
             ``.extra_coords`` properties will be used.
 
+        keepdims: `bool`, optional
+            If `True` keep length-1 dimensions rather than dropping.
+
         Returns
         -------
         `~ndcube.ndcube.NDCubeABC`
@@ -231,7 +235,8 @@ class NDCubeABC(astropy.nddata.NDDataBase):
     def crop_by_values(self,
                        *points: Iterable[u.Quantity | float],
                        units: Iterable[str | u.Unit] | None = None,
-                       wcs: BaseHighLevelWCS | ExtraCoordsABC | None = None
+                       wcs: BaseHighLevelWCS | ExtraCoordsABC | None = None,
+                       keepdims: bool = False
                        ) -> "NDCubeABC":
         """
         Crop using real world coordinates.
@@ -263,6 +268,9 @@ class NDCubeABC(astropy.nddata.NDDataBase):
             Will default to the ``.wcs`` property if not given. While any valid WCS
             could be used it is expected that either the ``.wcs`` or
             ``.extra_coords`` properties will be used.
+
+        keepdims: `bool`, optional
+            If `True` keep length-1 dimensions rather than dropping.
 
         Returns
         -------
@@ -554,14 +562,14 @@ class NDCubeBase(NDCubeABC, astropy.nddata.NDData, NDCubeSlicingMixin):
         CoordValues = namedtuple("CoordValues", identifiers)
         return CoordValues(*axes_coords[::-1])
 
-    def crop(self, *points, wcs=None):
+    def crop(self, *points, wcs=None, keepdims=False):
         # The docstring is defined in NDCubeABC
         # Calculate the array slice item corresponding to bounding box and return sliced cube.
-        item = self._get_crop_item(*points, wcs=wcs)
+        item = self._get_crop_item(*points, wcs=wcs, keepdims=keepdims)
         return self[item]
 
     @utils.cube.sanitize_wcs
-    def _get_crop_item(self, *points, wcs=None):
+    def _get_crop_item(self, *points, wcs=None, keepdims=False):
         # Sanitize inputs.
         no_op, points, wcs = utils.cube.sanitize_crop_inputs(points, wcs)
         # Quit out early if we are no-op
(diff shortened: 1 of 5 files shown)


ndcube/ndcube.py
```diff
@@ -215,6 +216,9 @@ def crop(self,
             could be used it is expected that either the ``.wcs`` or
             ``.extra_coords`` properties will be used.
 
+        keep_dims: `bool`, optional
+            If `True` then coordinates will be kept along size  dimension
```
samaloney, line 220:
```suggestion
            If `True` then size 1 dimensions will be kept rather than dropped.
```

ndcube/utils/cube.py
```diff
@@ -121,6 +121,9 @@ def get_crop_item_from_points(points, wcs, crop_by_values):
         Denotes whether cropping is done using high-level objects or "values",
         i.e. low-level objects.
 
+    keep_dims : `bool`
+        If true keep size 1 dimensions rather than dropping
```
DanRyanIrish, line 125: I find this a little clearer.
```suggestion
        If False, return item that will drop length-1 dimensions.  Otherwise item will keep length-1 dimensions.
```

DanRyanIrish (changes requested):
Thanks for this PR @samaloney.  It looks good, but `keepdims` should also be added to `crop_by_values`.  And a test for that similar to he one you already have.

DanRyanIrish:
Additionally, the kwargs should be `keepdims` to be consistent with `numpy`

ndcube/ndcube.py
```diff
@@ -215,6 +216,9 @@ def crop(self,
             could be used it is expected that either the ``.wcs`` or
             ``.extra_coords`` properties will be used.
 
+        keepdims: `bool`, optional
+            If `True` keep length-1 dimensions rather than dropping.
```
DanRyanIrish, line 220:
```suggestion
            If `False` and if cropping results in length-1 dimensions, these are sliced away in output cube.
            If `True`, length-1 dimensions are kept.  Default=False
```

ndcube/ndcube.py
```diff
@@ -264,6 +269,9 @@ def crop_by_values(self,
...
```
DanRyanIrish, line 273:
```suggestion
            If `False` and if cropping results in length-1 dimensions, these are sliced away in output cube.
            If `True`, length-1 dimensions are kept.  Default=False
```

ndcube/utils/cube.py
```diff
@@ -121,6 +121,9 @@ def get_crop_item_from_points(points, wcs, crop_by_values):
         Denotes whether cropping is done using high-level objects or "values",
         i.e. low-level objects.
 
+    keep_dims : `bool`
+        If `False`, return item that will drop length-1 dimensions otherwise, item will keep length-1 dimensions.
```
DanRyanIrish, line 125:
```suggestion
        If `False`, returned item will drop length-1 dimensions otherwise, item will keep length-1 dimensions.
```

DanRyanIrish (changes requested):

DanRyanIrish (approved):

DanRyanIrish:
Are you happy for me to merge this @samaloney ?

ndcube/ndcube.py
```diff
@@ -186,7 +186,8 @@ def axis_world_coords_values(self,
     @abc.abstractmethod
     def crop(self,
              *points: Iterable[Any],
-             wcs: BaseHighLevelWCS | ExtraCoordsABC | None = None
+             wcs: BaseHighLevelWCS | ExtraCoordsABC | None = None,
+             keepdims: bool = False,
```
Cadair, line 190: Do we want to add this to the ABC or just NDCube?

ndcube/ndcube.py
```diff
@@ -584,16 +594,16 @@ def _get_crop_item(self, *points, wcs=None):
                         raise TypeError(f"{type(value)} of component {j} in point {i} is "
                                         f"incompatible with WCS component {comp[j]} "
                                         f"{classes[j]}.")
-            return utils.cube.get_crop_item_from_points(points, wcs, False)
+            return utils.cube.get_crop_item_from_points(points, wcs, False, keepdims)
```
Cadair, line 597: :sparkles: pedantry :sparkles: 
```suggestion
            return utils.cube.get_crop_item_from_points(points, wcs, False, keepdims=keepdims)
```

ndcube/ndcube.py
```diff
@@ -626,7 +636,7 @@ def _get_crop_by_values_item(self, *points, units=None, wcs=None):
                         raise UnitsError(f"Unit '{points[i][j].unit}' of coordinate object {j} in point {i} is "
                                          f"incompatible with WCS unit '{wcs.world_axis_units[j]}'") from err
 
-        return utils.cube.get_crop_item_from_points(points, wcs, True)
+        return utils.cube.get_crop_item_from_points(points, wcs, True, keepdims)
```
Cadair, line 639:
```suggestion
        return utils.cube.get_crop_item_from_points(points, wcs, True, keepdims=keepdims)
```

Cadair (approved):
That was pleasantly easy

Cadair:
We should add narrative docs as well?

DanRyanIrish:
> We should add narrative docs as well?

It probably would be good to add a couple sentences and a code snippet at the end of this section: https://github.com/sunpy/ndcube/blob/main/docs/explaining_ndcube/slicing.rst#cropping-with-real-world-coordinates
It doesn't have to be much.  Something like:
```
By default, crop and crop_by_values discard length-1 dimensions to make the resulting cube more wieldy.
However, there are cases where it is preferable to keep the number of dimensions the same.
In this case, uses can set the keepdims kwarg to True in either crop or crop_by_values.

(Add code snippet example showing shape of cropped cube with length-1 dimensions.)

One use case for `keepdims=True` is when cropping leads to a cube with only one array element.
Because cropping an `NDCube` to a scalar is not allowed, such an operation would normally raise an error.
But if `keepdims=True`, a valid NDCube is returned with N length-1 dimensions.

(Perhaps another code snippet example.)
```

ndcube/ndcube.py
```diff
@@ -554,14 +564,14 @@ def axis_world_coords_values(self, *axes, pixel_corners=False, wcs=None):
         CoordValues = namedtuple("CoordValues", identifiers)
         return CoordValues(*axes_coords[::-1])
 
-    def crop(self, *points, wcs=None):
+    def crop(self, *points, wcs=None, keepdims=False):
```
nabobalis, line 567: Is there utility in making either this or everything to be enforced as keywords arguments?

DanRyanIrish: I don't see the need to be that restrictive.  Unless there's a good argument for it?

nabobalis: Well if you ever see the future of where you keep adding keywords, the ordering matters and people will assume that order for you. E.g plot in sunpy.map first arg not being ax for example.

I would personally go on the side of restrictive but its your library.

DanRyanIrish (approved):

DanRyanIrish:
Thanks @samaloney!!

ndcube/ndcube.py
Cadair: This is already kwarg only @nabobalis because of the `*points`.

PR #744, opened 2024-07-08
Rough attempt at ignoring units for rebin

nabobalis:
One way to do it.

Context:
sunpy.map.superpixel requireds astropy units as input.

rebin does not enforce this but a lot of superpixel code has units attached, so making that work with those would be a better idea than making users drop the pixel units.

### DIFF
--- a/changelog/744.bugfix.rst
+++ b/changelog/744.bugfix.rst
@@ -1,0 +1,1 @@
+:meth:`ndcube.NDcube.rebin` ``bin_shape`` argument now accepts a `astropy.units.Quantity` as input if the units are convertible to pixels.
--- a/changelog/744.bugfix.rst
+++ b/changelog/744.bugfix.rst
@@ -1,1 +1,1 @@
-:meth:`ndcube.NDcube.rebin` ``bin_shape`` argument now accepts a `astropy.units.Quantity` as input if the units are convertible to pixels.
+:meth:`ndcube.NDCube.rebin` ``bin_shape`` argument now accepts a `astropy.units.Quantity` as input if the units are convertible to pixels.
--- a/ndcube/ndcube.py
+++ b/ndcube/ndcube.py
@@ -1024,7 +1024,7 @@ class NDCube(NDCubeBase):
             Must be the same length as number of dimensions in data.
             Each element must be in int. If they are not they will be rounded
             to the nearest int. If provided as a `~astropy.units.Quantity` the
-            units are ignored, just the values are used as the shape.
+            units have to be convertible to pixels.
         operation : function
             Function applied to the data to derive values of the bins.
             Default is `numpy.mean`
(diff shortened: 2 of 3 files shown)


ndcube/ndcube.py
```diff
@@ -1122,6 +1122,8 @@ def my_propagate(uncertainty, data, mask, **kwargs):
         """
         # Sanitize input.
         new_unit = new_unit or self.unit
+        if isinstance(bin_shape, u.Quantity):
+            bin_shape = bin_shape.value
```
Cadair, line 1126: I would use to_value(u.pix)

Cadair: Or if we really don't care about units then use np.asarray and not give a hoot

changelog/744.bugfix.rst
```diff
@@ -0,0 +1 @@
+:meth:`ndcube.NDcube.rebin` ``bin_shape`` argument now accepts a `astropy.units.Quantity` as input by stripping the units and converting to a `numpy.ndarray` of integers.
```
nabobalis, line 1:
```suggestion
:meth:`ndcube.NDCube.rebin` ``bin_shape`` argument now accepts a `astropy.units.Quantity` as input by stripping the units and converting to a `numpy.ndarray` of integers.
```

ndcube/ndcube.py
```diff
@@ -1122,6 +1123,8 @@ def my_propagate(uncertainty, data, mask, **kwargs):
         """
         # Sanitize input.
         new_unit = new_unit or self.unit
+        if isinstance(bin_shape, u.Quantity):
+            bin_shape = bin_shape.to_value()
```
Cadair, line 1127: Surely we only want to allow units of pixels? or maybe pixels and dimensionless?

nabobalis: Up to you, I can check its pixel and error if not and then remove the unit.

ndcube/ndcube.py
```diff
@@ -1019,11 +1019,12 @@ def rebin(self, bin_shape, operation=np.mean, operation_ignores_mask=False, hand
 
         Parameters
         ----------
-        bin_shape : array-like
+        bin_shape : array-like, astropy.units.Quantity
```
Cadair, line 1022:
```suggestion
        bin_shape : array-like, `astropy.units.Quantity`
```

Cadair (approved):

Cadair:
@DanRyanIrish do you have philosophical objections to this?

nabobalis:
I just spoke to him via carrier pigeon, he said he approves.

DanRyanIrish (approved):

DanRyanIrish:
> @DanRyanIrish do you have philosophical objections to this?

I'm ok with this

PR #747, opened 2024-07-22
Update dep for source wheel build

nabobalis:
Should fix any CI issues.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -99,7 +99,7 @@ jobs:
       if: |
         github.event_name != 'pull_request' && (
           github.ref_name != 'main' ||
-          github.event_name == 'workflow_dispatch' 
+          github.event_name == 'workflow_dispatch'
         ) || (
           github.event_name == 'pull_request' &&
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -96,19 +96,28 @@ jobs:
...
   publish_pure:
+      # Build when pushing to any branch except main
+      # publish.yml will only publish if tagged ^v.*
       if: |
-        github.event_name != 'pull_request' && (
-          github.ref_name != 'main' ||
-          github.event_name == 'workflow_dispatch' )
+        (
+          github.event_name != 'pull_request' && (
+            github.ref_name != 'main' ||
```
Cadair, line 104: Why do we not run this (here and in core) when pushing to main? This whole if statement is weird.

Cadair (approved):

Cadair:
pre-commit.ci autofix

.github/workflows/ci.yml
Cadair: Moved this to the template.

PR #767, opened 2024-10-15
Only compute needed dims in _generate_world_coords

### LINKED ISSUE #554: Time to compute `axis_world_coords` scales poorly with dimensionality with off-diagonal elements in PCij
### Describe the performance issue

Running `axis_world_coords` to get the corresponding world coordinates along a pixel axis scales poorly with dimensionality of the data array when there are off diagonal elements in the PCij. 

As an example, take the following WCS for a 3D data cube with two celestial axes and one wavelenth axis. The latitudinal dimension is coupled to the third pixel dimension through a single off diagonal element in the PCij matrix.

```python
import numpy as np
import astropy.wcs
import ndcube

nx = 1024
ny = 4096
nz = 5
wcs_header = {
    'WCSAXES': 3,
    'CRPIX1': (nx + 1)/2,
    'CRPIX2': (ny + 1)/2,
    'CRPIX3': 1.0,
    'PC2_3': -1.0,
    'CDELT1': 5,
    'CDELT2': 5,
    'CDELT3': 1,
    'CUNIT1': 'arcsec',
    'CUNIT2': 'arcsec',
    'CUNIT3': 'Angstrom',
    'CTYPE1': 'HPLN-TAN',
    'CTYPE2': 'HPLT-TAN',
    'CTYPE3': 'WAVE',
    'CRVAL1': 0.0,
    'CRVAL2': 0.0,
    'CRVAL3': 1.0,

}
test_cube = ndcube.NDCube(np.random.rand(nz, ny, nx), wcs=astropy.wcs.WCS(header=wcs_header),)
```

Note that the resulting `axis_correlation_matrix` is 

```python
array([[ True,  True,  True],
       [ True,  True,  True],
       [False, False,  True]])
```

Running `axis_world_coords` to get the wavelength array that corresponds to the third dimension takes over 2 s,

```python
%%timeit -n 1 -r 1
test_cube.axis_world_coords(0)[1]
2.14 s ± 0 ns per loop (mean ± std. dev. of 1 run, 1 loop each)
```

If I increase `nz` to 20 (i.e. multiply the wavelength dimension by 4), I find that the execution time for `axis_world_coords` also goes up by a factor of 4,

```python
%%timeit -n 1 -r 1
test_cube.axis_world_coords(0)[1]
8.46 s ± 0 ns per loop (mean ± std. dev. of 1 run, 1 loop each)
```

Based on a few data points, it looks like the execution time for `axis_world_coords` scales linearly with `nz`. 

I'm assuming this is because an increasingly large `SkyCoord` array (the first entry in `axis_world_coords(0)` has to be computed and returned as the dimensionality grows. 

However, for much larger values of `nz`, it becomes impossibly slow to compute the corresponding world coordinates of the wavelength array. Is there some way to get this axis world coordinate with getting the corresponding celestial coordinates?

### To Reproduce

import sunpy

sunpy.map.Map(...)
etc

### Proposed fix

_No response_

svank:
@Cadair pointed me to `axis_world_coords` for verifying the order of the polarization axis in a DKIST/ViSP data set (which is an `NDCube`). It works, but for a dataset of shape `(4, 2, 1100, 915, 2555)` (polarization, repetition, scan step, wavelength, spatial), it computes a *lot* of world coordinates in 30 s (since rep, scan step and spatial are correlated), just to return the four Stokes coordinates. I dug in and took a shot at implementing the TODO, to have the underlying `_generate_world_coords` only run `pixel_to_world` for the axes that are being requested, offering a huge speedup in cases like this.

I rearranged `axis_world_coords` and `axis_world_coords_values` to first determine which world coordinate axes will be needed, and then pass those indices into `_generate_world_coords`, which skips processing any pixel dimension that doesn't contribute to those world dimensions.

Does this warrant a new test? I think that would look like verifying that `_generate_world_coords` returns just the throwaway value for unneeded axes.

My use case is
```python
>>> import dkist
>>> ds = dkist.load_dataset("/home/svankooten/globus/pid_2_86/ALMMM/")
>>> ds.axis_world_coords("polarization.stokes")
(StokesCoord(['I', 'Q', 'U', 'V']),)
```

The last line takes 3.23 ms with this PR, and 34.8 s without.

Fixes #554


Cadair:
Thank you so much for just jumping in and fixing this, it's been on my list for ages.

Will review later today.

Cadair:
pre-commit.ci autofix

Cadair (approved):

nabobalis:
SUMMON BACKPORT BOT

nabobalis:
@meeseeks backport to 2.2

nabobalis:
@MeeseeksDev backport to 2.2

PR #773, opened 2024-10-23
Applys explcit includes ruff file

CyclingNinja:
We've selected specific `.ruff.toml` rule categories to apply, this is the inital test of them

PR #774, opened 2024-10-25
All ruff experiment

CyclingNinja:


ndcube/extra_coords/extra_coords.py
```diff
@@ -490,7 +488,7 @@ def resample(self, factor, offset=0, ndcube=None, **kwargs):
         # Else interpolate the lookup table coordinates.
         factor = np.asarray(factor)
         new_grids = []
-        for c, d, f in zip(offset, cube_shape, factor):
+        for c, d, f in zip(offset, cube_shape, factor, strict=False):
```
nabobalis, line 491: I hate this isn't set to true automatically.

nabobalis:
I did not expect that like 90% of the auto changes are just double quotes.

nabobalis:
The output from ruff was so much it had to be truncated. 

Let me run it locally and paste the output here.

CyclingNinja:
> The output from ruff was so much it had to be truncated.
> 
> Let me run it locally and paste the output here.

It is expensive

nabobalis:
The full output from ruff is over 4000 lines.

I would say the vast majority fall under:

SLF001 Private member accessed
D102 Missing docstring in public method (and other doc ones like D105 etc)

We can skip these for test files for sure.

nabobalis:
The changes needed to fix this aren't hard, but it will take a while and it's boring.

CyclingNinja:
> The output from ruff was so much it had to be truncated.
> 
> Let me run it locally and paste the output here.

It is expensive

> The changes needed to fix this aren't hard, but it will take a while and it's boring.

Preach, I'm not sure how else we could do this though?

nabobalis:
> > The changes needed to fix this aren't hard, but it will take a while and it's boring.
> 
> Preach, I'm not sure how else we could do this though?

Either we wait for ruff to autofix them, or someone has to fix it.

Cadair:
I am much more in favour of picking our lints carefully than doing this fwiw.

nabobalis:
Closing in favor of https://github.com/sunpy/ndcube/pull/773

PR #778, opened 2024-11-06
Deprecate passing numerical keys to NDCollection

### LINKED ISSUE #736: NDCollection allows you to set keys as int or float and then gets upset when you slice
If you create a NDCollection object with integers or floats as the keys for the collection `__getitem__` then interprets these numbers as you wanting to slice the aligned axes of the cube. We should disallow numeric keys for the collection.

Cadair:
fixes #736


DanRyanIrish (approved):
Looks good.  Approve, pending tests passing

Issue #585, opened 2023-01-14
`_generate_world_coords` is slow and uses a lot of memory 


nabobalis:
duplicate of #554
 
### Describe the performance issue

In irispy-lmsal, I noticed that one of my RTD examples was dying due to memory consumption.
I ran a profiler and found:
[
![Untitled](https://user-images.githubusercontent.com/1392107/212457627-e09f82e9-ca40-4676-a02c-dfc81e890c78.png)
](url)

I was creating a range of times by calling `.time` on a NDCube object. 
The data array is:

```
(Pdb) sji_1400.data.shape
(400, 417, 388)
```

Now I am using a gWCS so I am unclear if that makes the problem worse.

### To Reproduce

I don't have a concrete self-contained example right now. But I can try to create one if needed.

### Proposed fix

I saw this older PR (https://github.com/sunpy/ndcube/pull/474) but unsure if it would help.

Otherwise, I have no idea what could be done to fix it. Help would be great.

wtbarnes:
I wonder if this is related to #554. What does the correlation matrix look like for the IRIS data you mentioned?

nabobalis:
Mine is:
```
(Pdb) sji_1400.wcs.axis_correlation_matrix
array([[ True,  True,  True],
       [ True,  True,  True],
       [False, False,  True]])
```

Cadair:
This is indeed a duplicate of #554 although thanks for the flame graph that's nice.
 
The fundamental issue here is that when you couple your spatial axes to your other axes you *need* to generate a 3D array of coordinates as the spatial coordinates change along the other axes, so there isn't really anything you can do to reduce the scale of the computation. We would need to profile the code, mostly in Astropy and try and improve performance that way.

Cadair:
The main thing I would ask people running into this is do you really *need* to be doing this computation for your whole array?

nabobalis:
> Can you show me the code you are running, getting all the timestamps shouldn't trigger a full 3D coord set.

The code is a large example which I have but the line is calling `.time` here: https://github.com/LM-SAL/irispy-lmsal/blob/c7aafeaccfa865e6eb6ac013798be7f322b6ebe3/examples/skip_umbral_flashes.py#L112

Cadair:
What's that calling?

nabobalis:
It actually is from sunraster: 
https://github.com/sunpy/sunraster/blob/main/sunraster/spectrogram.py#L361

[...]

PR #780, opened 2024-11-08
Faster shortcut for working out coordinates values for non-correlated WCS 

nabobalis:
Fixes https://github.com/sunpy/ndcube/issues/585 

So in my case, I use a DKIST compound model which has two coupled pixel axes and a time axes. 
If I want just the time axis, the current code will work out the entire grid leading to a very dense and memory intense results. 

So I added a really specific hack which works for my code and seems to not break the rest of the test suite. 

TODO:

Unit tests:
- [ ] EDGE CASES - THERE SHOULD BE SOME?

### DIFF
--- a/changelog/780.bugfix.rst
+++ b/changelog/780.bugfix.rst
@@ -1,0 +1,1 @@
+Added an internal method to shortcut non-correlated axes avoiding the creation of a full coordinate grid, reducing memory use in specific circumstances.
--- a/ndcube/ndcube.py
+++ b/ndcube/ndcube.py
@@ -484,11 +484,17 @@ class NDCubeBase(NDCubeABC, astropy.nddata.NDData, NDCubeSlicingMixin):
         # This bypasses the entire rest of the function below which works out the full set of coordinates
         # This only works for WCS that have the same number of world and pixel dimensions
         if needed_axes is not None and not isinstance(wcs, ExtraCoords) and np.sum(wcs.axis_correlation_matrix[needed_axes]) == 1:
-            lims = (-0.5, self.data.shape[::-1][needed_axes[0]] + 1) if pixel_corners else (0, self.data.shape[::-1][needed_axes[0]])
-            indices = [np.arange(lims[0], lims[1]) if wanted else [0] for wanted in wcs.axis_correlation_matrix[needed_axes][0]]
+            # Account for non-pixel axes affecting the value of needed_axes
+            # Only works for one axis
+            if np.max(wcs.axis_correlation_matrix[needed_axes][0].shape) == needed_axes[0]:
+                needed_axis = needed_axes[0] - 1
+            else:
+                needed_axis = needed_axes[0]
+            lims = (-0.5, self.data.shape[::-1][needed_axis] + 1) if pixel_corners else (0, self.data.shape[::-1][needed_axis])
+            indices = [np.arange(lims[0], lims[1]) if wanted else [0] for wanted in wcs.axis_correlation_matrix[needed_axis]]
             world_coords = wcs.pixel_to_world_values(*indices)
             if units:
-                world_coords = world_coords << u.Unit(wcs.world_axis_units[needed_axes[0]])
+                world_coords = world_coords << u.Unit(wcs.world_axis_units[needed_axis])
             return world_coords
 
         # Create a meshgrid of all pixel coordinates.
@@ -582,23 +588,19 @@ class NDCubeBase(NDCubeABC, astropy.nddata.NDData, NDCubeSlicingMixin):
         # Docstring in NDCubeABC.
         if isinstance(wcs, BaseHighLevelWCS):
             wcs = wcs.low_level_wcs
-
         orig_wcs = wcs
         if isinstance(wcs, ExtraCoords):
             wcs = wcs.wcs
-
+            if not wcs:
+                return ()
         world_indices = utils.wcs.calculate_world_indices_from_axes(wcs, axes)
-
         axes_coords = self._generate_world_coords(pixel_corners, orig_wcs, world_indices, units=True)
-
         world_axis_physical_types = wcs.world_axis_physical_types
-
         # If user has supplied axes, extract only the
         # world coords that correspond to those axes.
         if axes:
             axes_coords = [axes_coords[i] for i in world_indices]
             world_axis_physical_types = tuple(np.array(world_axis_physical_types)[world_indices])
-
         # Return in array order.
         # First replace characters in physical types forbidden for namedtuple identifiers.
         identifiers = []
--- a/ndcube/ndcube.py
+++ b/ndcube/ndcube.py
@@ -503,40 +503,36 @@ class NDCubeBase(NDCubeABC, astropy.nddata.NDData, NDCubeSlicingMixin):
         pixel_shape = self.data.shape[::-1]
         if pixel_corners:
             pixel_shape = tuple(np.array(pixel_shape) + 1)
-            pixel_ranges = [np.arange(i) - 0.5 for i in pixel_shape]
+            ranges = [np.arange(i) - 0.5 for i in pixel_shape]
         else:
-            pixel_ranges = [np.arange(i) for i in pixel_shape]
+            ranges = [np.arange(i) for i in pixel_shape]
 
         # Limit the pixel dimensions to the ones present in the ExtraCoords
         if isinstance(wcs, ExtraCoords):
-            pixel_ranges = [pixel_ranges[i] for i in wcs.mapping]
+            ranges = [ranges[i] for i in wcs.mapping]
             wcs = wcs.wcs
             if wcs is None:
                 return []
+
         # This value of zero will be returned as a throwaway for unneeded axes, and a numerical value is
         # required so values_to_high_level_objects in the calling function doesn't crash or warn
         world_coords = [0] * wcs.world_n_dim
         for (pixel_axes_indices, world_axes_indices) in _split_matrix(wcs.axis_correlation_matrix):
-            if (
-                needed_axes is not None
-                and len(needed_axes)
-                and all(
-                    world_axis not in needed_axes
-                    for world_axis in world_axes_indices
-                )
-            ):
+            if (needed_axes is not None
+                    and len(needed_axes)
+                    and not any(world_axis in needed_axes for world_axis in world_axes_indices)):
                 # needed_axes indicates which values in world_coords will be used by the calling
                 # function, so skip this iteration if we won't be producing any of those values
                 continue
             # First construct a range of pixel indices for this set of coupled dimensions
-            pixel_ranges_subset = [pixel_ranges[idx] for idx in pixel_axes_indices]
-            # Then get a set of non-correlated dimensions
+            sub_range = [ranges[idx] for idx in pixel_axes_indices]
+            # Then get a set of non correlated dimensions
             non_corr_axes = set(range(wcs.pixel_n_dim)) - set(pixel_axes_indices)
             # And inject 0s for those coordinates
             for idx in non_corr_axes:
-                pixel_ranges_subset.insert(idx, 0)
-            # Generate a grid of broadcast-able pixel indices for all pixel dimensions
-            grid = np.meshgrid(*pixel_ranges_subset, indexing='ij')
+                sub_range.insert(idx, 0)
+            # Generate a grid of broadcastable pixel indices for all pixel dimensions
+            grid = np.meshgrid(*sub_range, indexing='ij')
             # Convert to world coordinates
             world = wcs.pixel_to_world_values(*grid)
             # TODO: this isinstance check is to mitigate https://github.com/spacetelescope/gwcs/pull/332
@@ -549,6 +545,7 @@ class NDCubeBase(NDCubeABC, astropy.nddata.NDData, NDCubeSlicingMixin):
                 array_slice[wcs.axis_correlation_matrix[idx]] = slice(None)
                 tmp_world = world[idx][tuple(array_slice)].T
                 world_coords[idx] = tmp_world
+
         if units:
             for i, (coord, unit) in enumerate(zip(world_coords, wcs.world_axis_units)):
                 world_coords[i] = coord << u.Unit(unit)
@@ -560,27 +557,35 @@ class NDCubeBase(NDCubeABC, astropy.nddata.NDData, NDCubeSlicingMixin):
         # Docstring in NDCubeABC.
         if isinstance(wcs, BaseHighLevelWCS):
             wcs = wcs.low_level_wcs
+
         orig_wcs = wcs
         if isinstance(wcs, ExtraCoords):
             wcs = wcs.wcs
             if not wcs:
                 return ()
+
         object_names = np.array([wao_comp[0] for wao_comp in wcs.world_axis_object_components])
         unique_obj_names = utils.misc.unique_sorted(object_names)
         world_axes_for_obj = [np.where(object_names == name)[0] for name in unique_obj_names]
+
         # Create a mapping from world index in the WCS to object index in axes_coords
         world_index_to_object_index = {}
         for object_index, world_axes in enumerate(world_axes_for_obj):
             for world_index in world_axes:
                 world_index_to_object_index[world_index] = object_index
+
         world_indices = utils.wcs.calculate_world_indices_from_axes(wcs, axes)
         object_indices = utils.misc.unique_sorted(
             [world_index_to_object_index[world_index] for world_index in world_indices]
         )
+
         axes_coords = self._generate_world_coords(pixel_corners, orig_wcs, world_indices, units=False)
+
         axes_coords = values_to_high_level_objects(*axes_coords, low_level_wcs=wcs)
+
         if not axes:
             return tuple(axes_coords)
+
         return tuple(axes_coords[i] for i in object_indices)
 
     @utils.cube.sanitize_wcs
@@ -588,19 +593,25 @@ class NDCubeBase(NDCubeABC, astropy.nddata.NDData, NDCubeSlicingMixin):
         # Docstring in NDCubeABC.
         if isinstance(wcs, BaseHighLevelWCS):
             wcs = wcs.low_level_wcs
+
         orig_wcs = wcs
         if isinstance(wcs, ExtraCoords):
             wcs = wcs.wcs
             if not wcs:
                 return ()
+
         world_indices = utils.wcs.calculate_world_indices_from_axes(wcs, axes)
+
         axes_coords = self._generate_world_coords(pixel_corners, orig_wcs, world_indices, units=True)
+
         world_axis_physical_types = wcs.world_axis_physical_types
+
         # If user has supplied axes, extract only the
         # world coords that correspond to those axes.
         if axes:
             axes_coords = [axes_coords[i] for i in world_indices]
             world_axis_physical_types = tuple(np.array(world_axis_physical_types)[world_indices])
+
         # Return in array order.
         # First replace characters in physical types forbidden for namedtuple identifiers.
         identifiers = []
--- a/ndcube/ndcube.py
+++ b/ndcube/ndcube.py
@@ -481,8 +481,7 @@ class NDCubeBase(NDCubeABC, astropy.nddata.NDData, NDCubeSlicingMixin):
         """Unitful representation of the NDCube data."""
         return u.Quantity(self.data, self.unit, copy=_NUMPY_COPY_IF_NEEDED)
 
-
-    def _generate_independent_world_coords(self, pixel_corners, wcs, pixel_axes, units):
+    def _generate_independent_world_coords(self, pixel_corners, wcs, needed_axes, units):
         """
         Generate world coordinates for independent axes.
 
@@ -495,8 +494,8 @@ class NDCubeBase(NDCubeABC, astropy.nddata.NDData, NDCubeSlicingMixin):
             If one needs pixel corners, otherwise pixel centers.
         wcs : astropy.wcs.WCS
             The WCS.
-        pixel_axes : array-like
-            The pixel axes.
+        needed_axes : array-like
+            The required pixel axes.
         units : bool
             If units are needed.
 
@@ -505,29 +504,19 @@ class NDCubeBase(NDCubeABC, astropy.nddata.NDData, NDCubeSlicingMixin):
         array-like
             The world coordinates.
         """
-        naxes = len(self.data.shape)
-        pixel_indices = [np.array([0], dtype=int).reshape([1] * naxes).squeeze()] * naxes
-        for pixel_axis in pixel_axes:
-            len_axis = self.data.shape[::-1][pixel_axis]
-            # Define limits of desired pixel range based on whether corners or centers are desired
-            lims = (-0.5, len_axis + 1) if pixel_corners else (0, len_axis)
-            pix_ind = np.arange(lims[0], lims[1])
-            shape = [1] * naxes
-            shape[pixel_axis] = len(pix_ind)
-            pixel_indices[pixel_axis] = pix_ind.reshape(shape)
-        world_coords = wcs.pixel_to_world_values(*pixel_indices)
-        # TODO: Remove NaNs??? These should not be here
-        if np.isnan(world_coords).any():
-            if isinstance(world_coords, tuple| list):
-                world_coords = [world_coord[~np.isnan(world_coord)] for world_coord in world_coords]
-            else:
-                world_coords = world_coords[~np.isnan(world_coords)]
+        needed_axes = np.array(needed_axes).squeeze()
+        if self.data.ndim in needed_axes:
+            required_axes = needed_axes - 1
+        else:
+            required_axes = needed_axes
+        lims = (-0.5, self.data.shape[::-1][required_axes] + 1) if pixel_corners else (0, self.data.shape[::-1][required_axes])
+        indices = [np.arange(lims[0], lims[1]) if wanted else [0] for wanted in wcs.axis_correlation_matrix[required_axes]]
+        world_coords = wcs.pixel_to_world_values(*indices)
         if units:
-            mod = abs(wcs.world_n_dim - naxes) if wcs.world_n_dim > naxes else 0
-            world_coords = world_coords << u.Unit(wcs.world_axis_units[np.squeeze(pixel_axes)+mod])
+            world_coords = world_coords << u.Unit(wcs.world_axis_units[needed_axes])
         return world_coords
 
-    def _generate_dependent_world_coords(self, pixel_corners, wcs, pixel_axes, units):
+    def _generate_dependent_world_coords(self, pixel_corners, wcs, needed_axes, units):
         """
         Generate world coordinates for dependent axes.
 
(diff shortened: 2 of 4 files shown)


nabobalis:
TODO:

Unit tests:
- [ ] IRIS COMPOUND MODEL one?
- [ ] EDGE CASES?

ndcube/ndcube.py
```diff
@@ -444,42 +444,56 @@ def quantity(self):
         return u.Quantity(self.data, self.unit, copy=_NUMPY_COPY_IF_NEEDED)
 
     def _generate_world_coords(self, pixel_corners, wcs, needed_axes=None, *, units):
-        # Create meshgrid of all pixel coordinates.
-        # If user wants pixel_corners, set pixel values to pixel corners.
+        # This will generate only the coordinates that are needed if there is no correlation within the WCS
+        # This bypasses the entire rest of the function below which works out the full set of coordinates
+        # This only works for WCS that have the same number of world and pixel dimensions
```
nabobalis, line 449: This probably shouldn't be a limitation but right now I can't work out a good idea.

ndcube/ndcube.py
```diff
@@ -444,42 +444,56 @@ def quantity(self):
...
+        if not pixel_corners and needed_axes is not None and not isinstance(wcs, ExtraCoords) and np.sum(wcs.axis_correlation_matrix[needed_axes]) == 1 and len(self.data.shape) == wcs.world_n_dim:
+            indices = [np.arange(self.data.shape[::-1][needed_axes[0]]) if wanted else [0] for wanted in wcs.axis_correlation_matrix[needed_axes][0]]
+            world_coords = wcs.pixel_to_world_values(*indices)
+            if units:
+                world_coords = world_coords << u.Unit(wcs.world_axis_units[needed_axes[0]])
+            return world_coords
+
+        # Create a meshgrid of all pixel coordinates.
```
nabobalis, line 457: Everything from this line down is just style changes.

ndcube/ndcube.py
```diff
@@ -444,42 +444,56 @@ def quantity(self):
         return u.Quantity(self.data, self.unit, copy=_NUMPY_COPY_IF_NEEDED)
 
     def _generate_world_coords(self, pixel_corners, wcs, needed_axes=None, *, units):
-        # Create meshgrid of all pixel coordinates.
-        # If user wants pixel_corners, set pixel values to pixel corners.
+        # This will generate only the coordinates that are needed if there is no correlation within the WCS
+        # This bypasses the entire rest of the function below which works out the full set of coordinates
+        # This only works for WCS that have the same number of world and pixel dimensions
+        if not pixel_corners and needed_axes is not None and not isinstance(wcs, ExtraCoords) and np.sum(wcs.axis_correlation_matrix[needed_axes]) == 1 and len(self.data.shape) == wcs.world_n_dim:
```
DanRyanIrish, line 450: For future developers (and reviewer), could you add comments stating why each of these conditions are required?  This my clarify some questions I have, such as:
* why does this only work when `not pixel_corners`?
* Does this only work when there is only one `needed_axis`?
  * If not, what is the significance of the magic `0` index below, e.g. `wcs.axis_correlation_matrix[needed_axes][0]]`

ndcube/ndcube.py
DanRyanIrish: Can you elaborate on what the difficulty is here?

nabobalis: Too stupid to work out the line of code to overcome this limiation.

ndcube/ndcube.py
nabobalis: > For future developers (and reviewer), could you add comments stating why each of these conditions are required? This may clarify some questions I have, such as:

Fair

> * why does this only work when `not pixel_corners`?

Too stupid to work out how to fix it.

> * Does this only work when there is only one `needed_axis`?

Too stupid to work out how to fix it.
   
>   * If not, what is the significance of the magic `0` index below, e.g. `wcs.axis_correlation_matrix[needed_axes][0]]`

I think `wcs.axis_correlation_matrix[needed_axes]` returns a list that has to be escaped hence the [0]. But I should check.

ndcube/ndcube.py
```diff
@@ -444,42 +444,56 @@ def quantity(self):
...
+            indices = [np.arange(self.data.shape[::-1][needed_axes[0]]) if wanted else [0] for wanted in wcs.axis_correlation_matrix[needed_axes][0]]
```
DanRyanIrish, line 451: I think something like below is what you need to make this valid of `pixel_corners` as well.
```suggestion
            lims = (-0.5, self.data.shape[::-1][needed_axes[0]] + 1) if pixel_corners else (0, self.data.shape[::-1][needed_axes[0]])
            indices = [np.arange(lims[0], lims[1]) if wanted else [0] for wanted in wcs.axis_correlation_matrix[needed_axes][0]]
```

ndcube/ndcube.py
DanRyanIrish: OK.  Put another way, why does enforcing this limitation make the problem simpler?  If you just remove this condition, so long as `np.sum(wcs.axis_correlation_matrix[needed_axes]) == 1`, does it still work as intended?

ndcube/ndcube.py
DanRyanIrish: > > * why does this only work when `not pixel_corners`?
> 
> Too stupid to work out how to fix it.

See suggested change below for initial guess at how to solve this.  Can you write a test to check out if this in fact does fix it?

> > * Does this only work when there is only one `needed_axis`?
> 
> Too stupid to work out how to fix it.

Would this work if you made the unwanted dimensions for indices have shape `(1) * n` where `n` is the number of dimensions.  And the indices in the wanted dimensions can have the `np.arange` output.  Since the dimensions are independent, this should result in broadcastable inputs to `wcs.pixel_to_world_values`, and so you should be able to get coords out for multiple independent axes, without blowing up the memory.  At least, I think this is true.

> > * If not, what is the significance of the magic `0` index below, e.g. `wcs.axis_correlation_matrix[needed_axes][0]]`
> 
> I think `wcs.axis_correlation_matrix[needed_axes]` returns a list that has to be escaped hence the [0]. But I should check.

Is this list length 1 because you enforced this condition above: `np.sum(wcs.axis_correlation_matrix[needed_axes]) == 1`?

ndcube/ndcube.py
nabobalis: So I added this and the tests fail:

```
        world = wcs.pixel_to_world_values(0, 0, 0)
        if astropy.__version__ >= "7.0.0.dev":
            np.testing.assert_allclose(world, (14, np.nan, np.nan, -7.0))
            np.testing.assert_allclose(wcs.world_to_pixel_values(*world), (0, np.nan, np.nan))
        else:
>           np.testing.assert_allclose(world, (14, -12, -2.6e+10, -7.0))

../../.tox/py/lib/python3.13/site-packages/ndcube/wcs/wrappers/tests/test_compound_wcs.py:179: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

args = (<function assert_allclose.<locals>.compare at 0x7755ea22eac0>, array([14., nan, nan, -7.]), array([ 1.4e+01, -1.2e+01, -2.6e+10, -7.0e+00]))
kwds = {'equal_nan': True, 'err_msg': '', 'header': 'Not equal to tolerance rtol=1e-07, atol=0', 'strict': False, ...}

    @wraps(func)
    def inner(*args, **kwds):
        with self._recreate_cm():
>           return func(*args, **kwds)
E           AssertionError: 
E           Not equal to tolerance rtol=1e-07, atol=0
E           
E           nan location mismatch:
E            ACTUAL: array([14., nan, nan, -7.])
E            DESIRED: array([ 1.4e+01, -1.2e+01, -2.6e+10, -7.0e+00])
```
this code now seems to give the same output as astropy 7.0 will give in the future?!

ndcube/ndcube.py
nabobalis: The reason I added that was to avoid the hack I wrote breaking here:

```
  File "/home/nabil/Git/ndcube/.tox/py/lib/python3.13/site-packages/ndcube/ndcube.py", line 451, in _generate_world_coords
    lims = (-0.5, self.data.shape[::-1][needed_axes[0]] + 1) if pixel_corners else (0, self.data.shape[::-1][needed_axes[0]])
```
due to
```
ValueError: operands could not be broadcast together with shapes (1,6) (5,)
```

ndcube/ndcube.py
nabobalis: With this, I think it works perfectly thank you. I just have to account for a test change:

```
-> assert u.allclose(coords, [1.01e-09, 1.03e-09, 1.05e-09, 1.07e-09, 1.09e-09] * u.m)
(Pdb++) coords
CoordValues(em_wl=<Quantity [1.01e-09, 1.03e-09, 1.05e-09, 1.07e-09, 1.09e-09, 1.11e-09] m>)
(Pdb++) coords[0]
<Quantity [1.01e-09, 1.03e-09, 1.05e-09, 1.07e-09, 1.09e-09, 1.11e-09] m>
(Pdb++) 
```
I need to escape another dim which I didn't need to before?

ndcube/ndcube.py
nabobalis: In the one case in the test suite, no this is not enough.

`(0, self.data.shape[::-1][needed_axes[0]])` errors with `*** IndexError: tuple index out of range` as the "self.data.shape" is length 3 but the needed_axis is "3". There is no direct translation between the needed world coordinate and the corresponding pixel coordinate in this block.

ndcube/tests/test_ndcube.py
```diff
@@ -237,11 +237,16 @@ def test_axis_world_coords_single(axes, ndcube_3d_ln_lt_l):
 
 @pytest.mark.parametrize("axes", [[-1], [2], ["em"]])
 def test_axis_world_coords_single_pixel_corners(axes, ndcube_3d_ln_lt_l):
+
+    # We go from 4 pixels to 6 pixels when we add pixel corners
```
nabobalis, line 241: Not sure if this correct at all.

ndcube/tests/test_ndcube.py
```diff
@@ -237,11 +237,16 @@ def test_axis_world_coords_single(axes, ndcube_3d_ln_lt_l):
...
+    coords = ndcube_3d_ln_lt_l.axis_world_coords_values(*axes, pixel_corners=False)
+    assert u.allclose(coords[0], [1.02e-09, 1.04e-09, 1.06e-09, 1.08e-09] * u.m)
```
nabobalis, line 243: This grew another dim, maybe an astropy 7 change?

ndcube/tests/test_ndcube.py
```diff
@@ -235,13 +235,32 @@ def test_axis_world_coords_single(axes, ndcube_3d_ln_lt_l):
...
+
     coords = ndcube_3d_ln_lt_l.axis_world_coords_values(*axes, pixel_corners=True)
-    assert u.allclose(coords, [1.01e-09, 1.03e-09, 1.05e-09, 1.07e-09, 1.09e-09] * u.m)
+    assert u.allclose(coords[0], [1.01e-09, 1.03e-09, 1.05e-09, 1.07e-09, 1.09e-09, 1.11e-09] * u.m)
```
nabobalis, line 260: This grew another dim but I do not know why right now.

ndcube/tests/test_ndcube.py
```diff
@@ -235,13 +235,32 @@ def test_axis_world_coords_single(axes, ndcube_3d_ln_lt_l):
...
 
 
+def test_axis_world_coords_crazy(ndcube_3d_wave_lt_ln_ec_time):
+    # This replicates a specific NDCube test in the visualization.rst
+    coords = ndcube_3d_wave_lt_ln_ec_time.axis_world_coords('time', wcs=ndcube_3d_wave_lt_ln_ec_time.combined_wcs)
+    assert len(coords) == 1
+    assert isinstance(coords[0], Time)
+    assert np.all(coords[0] == Time(['2000-01-01T00:00:00.000', '2000-01-01T00:01:00.000', '2000-01-01T00:02:00.000']))
+
+    # This fails and returns the wrong coords
```
nabobalis, line 245: This operation returns very very wrong values.

changelog/780.bugfix.rst
```diff
@@ -0,0 +1 @@
+Added an internal method to shortcut non-correlated axes avoiding the creation of a full coordinate grid, reducing memory use in specific circumstances.
```
nabobalis, line 1: TODO: Better changelog.

ndcube/ndcube.py
```diff
@@ -544,6 +561,8 @@ def axis_world_coords_values(self, *axes, pixel_corners=False, wcs=None):
         orig_wcs = wcs
         if isinstance(wcs, ExtraCoords):
             wcs = wcs.wcs
+            if not wcs:
+                return ()
```
nabobalis, line 565: This was missing and is included in the other version of this method, so I added it.

ndcube/tests/test_ndcube.py
```diff
@@ -235,13 +235,32 @@ def test_axis_world_coords_single(axes, ndcube_3d_ln_lt_l):
     assert u.allclose(coords[0], [1.02e-09, 1.04e-09, 1.06e-09, 1.08e-09] * u.m)
 
 
+def test_axis_world_coords_crazy(ndcube_3d_wave_lt_ln_ec_time):
+    # This replicates a specific NDCube test in the visualization.rst
```
nabobalis, line 239: So my understanding of this and I don't know if this is correct.

This NDCube has a 3D WCS, which is Wavelength, Space, Space and has time as an extra coord.

We pass to `axis_world_coords_values` the combined WCS which has bolted on time at the end and then asking the code to do a WCS look up on that. 

But since its in extra coords, should the code not go into a different path?

ndcube/ndcube.py
DanRyanIrish: Would [this util function](https://github.com/sunpy/ndcube/blob/main/ndcube/utils/wcs.py#L111) solve that problem?

DanRyanIrish (commented):
More review to follow.

ndcube/ndcube.py
```diff
@@ -444,8 +444,25 @@ def quantity(self):
         return u.Quantity(self.data, self.unit, copy=_NUMPY_COPY_IF_NEEDED)
 
     def _generate_world_coords(self, pixel_corners, wcs, needed_axes=None, *, units):
-        # Create meshgrid of all pixel coordinates.
-        # If user wants pixel_corners, set pixel values to pixel corners.
+        # This will generate only the coordinates that are needed if there is no correlation within the WCS
+        # This bypasses the entire rest of the function below which works out the full set of coordinates
+        # This only works for WCS that have the same number of world and pixel dimensions
+        if needed_axes is not None and not isinstance(wcs, ExtraCoords) and np.sum(wcs.axis_correlation_matrix[needed_axes]) == 1:
+            # Account for non-pixel axes affecting the value of needed_axes
```
DanRyanIrish, line 451: Can you better explain this comment.

nabobalis: So the passed in needed_axes is given in world indices but the loop below works on the data shape, so I made a random guess at modifying the needed axes to give a variable which does not error. 

It isn't correct

DanRyanIrish:
Hi @nabobalis.  My reading of this PR is that it should be generalised to make it a bit simpler to implement and follow.  If I understand correctly, you want to special-case the scenario where all wanted world axes (whether 1 or more) correspond to their own unique pixel axis.  In this scenario, you don't have to work out all the coordinates for the entire (m, n) pixel grid, but instead just then 1-D dimensions corresponding to the desired world axes.  This saves RAM and time.  Am I right so far?

If this is the case, then I think it would help readability, to move your code to a new private method, e.g. `_generate_independent_world_coords()`.  The pre-existing code could be moved to another private method, e.g. `_generate_dependent_world_coords()`, and then `_generate_world_coords()` could simply check whether the needed axes are independent, and then call the appropriate method above.  Something like this:
```python
def _generate_world_coords(self, pixel_corners, wcs, needed_axes=None, *, units):
    if needed_axes is None:
        needed_axes = list(range(wcs.n_world_dims))
    axes_are_independent = []
    pixel_axes = {}
    for world_axis in needed_axes:
        pix_ax = ndcube.utils.wcs.world_axis_to_pixel_axes(world_axis, wcs.axis_correlation_matrix)
        axes_are_independent.append(len(pix_ax) == 1)
        pixel_axes = pixel_axes.union(set(pix_ax))
    if all(axes_are_independent) and len(pixel_axes) == len(needed_axes):
        return self._generate_independent_world_coords(pixel_corners, wcs, pixel_axes, units)
    else:
        return self._generate_dependent_world_coords(pixel_corners, wcs, pixel_axes, units)
```

This brings us to your new code, i.e. `_generate_independent_world_coords()`.  Since the check whether the needed axes are all independent will be done in the new `_generate_world_coords()`, we don't have to repeat that.  We need to check whether we want pixel centers or corners, then define the pixel grids, where unwanted axes are set to `np.array([0], dtype=int).reshape([1] * ndim)` where `ndim` is the number of dimensions in the data array.  Wanted dimensions are then set to `np.arange(n).astype(int).reshape((1,...,n,...1)` where `n` is the length of the relevant axis.  So something like this:
```python
def _generate_independent_world_coords(pixel_corners, wcs, pixel_axes, units):
    naxes = len(self.shape)
    pixel_indices = [np.array([0], dtype=int).reshape([1] * naxes] * naxes
    for pixel_axis in pixel_axes:
        # Define limits of desired pixel range based on whether corners or centers are desired 
        len_axis = self.data.shape[::-1][pixel_axis]
        lims = (-0.5, len_axis + 1) if pixel_corners else (0, len_axis)
        pix_ind = np.arange(lims[0], lims[1])
        shape = [1] * naxes
        shape[pixel_axis] = len(pix_ind)
        pix_ind = pix_ind.reshape(shape)
        pixel_indices[pixel_axis] = pix_ind
    world_coords = wcs.pixel_to_world_values(*pixel_indices)
    if units:
        # Attach units to each coord here.
    return world_coords
```

I presume there's still some work to actually make this code work, but hopefully it's a good start.  Let me know if you have questions.

(And these private methods should probably have their own docstrings so it's easier for developers to understand them in the future.)

nabobalis:
> Hi @nabobalis. My reading of this PR is that it should be generalised to make it a bit simpler to implement and follow. If I understand correctly, you want to special-case the scenario where all wanted world axes (whether 1 or more) correspond to their own unique pixel axis. In this scenario, you don't have to work out all the coordinates for the entire (m, n) pixel grid, but instead just then 1-D dimensions corresponding to the desired world axes. This saves RAM and time. Am I right so far?
>

Yes.

I will try to attempt your suggestions.

ndcube/ndcube.py
```diff
@@ -585,7 +551,8 @@ def axis_world_coords_values(self, *axes, pixel_corners=False, wcs=None):
             identifier = identifier.replace("-", "__")
             identifiers.append(identifier)
         CoordValues = namedtuple("CoordValues", identifiers)
-        return CoordValues(*axes_coords[::-1])
+        flag = len(axes_coords) == 1 or len(axes_coords.shape) > 1
+        return CoordValues(*axes_coords[::-1]) if flag else CoordValues(axes_coords)
```
nabobalis, line 555: TODO: Find better way

ndcube/ndcube.py
```diff
@@ -443,139 +444,104 @@ def quantity(self):
...
         world_indices = utils.wcs.calculate_world_indices_from_axes(wcs, axes)
         object_indices = utils.misc.unique_sorted(
             [world_index_to_object_index[world_index] for world_index in world_indices]
         )
-
-        axes_coords = self._generate_world_coords(pixel_corners, orig_wcs, world_indices, units=False)
-
+        axes_coords = self._generate_world_coords(pixel_corners, orig_wcs, needed_axes=world_indices, units=False)
+        if not isinstance(axes_coords, list):
+            axes_coords = [axes_coords]
```
nabobalis, line 521: A hack, could be moved into the private method.

ndcube/ndcube.py
```diff
@@ -492,71 +503,88 @@ def _generate_world_coords(self, pixel_corners, wcs, needed_axes=None, *, units)
...
                 world_coords[i] = coord << u.Unit(unit)
+        return world_coords
 
+
+    def _generate_world_coords(self, pixel_corners, wcs, *, needed_axes=None, units=None):
+        # TODO: Workout why I need this twice now.
+        if isinstance(wcs, ExtraCoords):
+            wcs = wcs.wcs
+            if not wcs:
+                return ()
```
nabobalis, line 517: TODO: Figure out

ndcube/ndcube.py
```diff
@@ -566,7 +594,8 @@ def axis_world_coords_values(self, *axes, pixel_corners=False, wcs=None):
             identifier = identifier.replace("-", "__")
             identifiers.append(identifier)
         CoordValues = namedtuple("CoordValues", identifiers)
-        return CoordValues(*axes_coords[::-1])
+        flag = len(axes_coords) == 1 or isinstance(axes_coords, tuple | list)
```
nabobalis, line 597: TODO: Figure out better way

ndcube/ndcube.py
nabobalis: Maybe it does but I don't see how.

nabobalis:
So I reworked it, there are two issues (as far as I can tell).

1. I still need to account for WCS where dimensions of it are greater than the data array
2. There is a bug in the dependent calculation now which breaks the rebin tests. We get back nan values now when we did not before.

I am trying but failing pretty badly on both of them at the moment.

ndcube/ndcube.py
```diff
@@ -443,31 +445,46 @@ def quantity(self):
...
+            shape = [1] * naxes
+            shape[pixel_axis] = len(pix_ind)
+            pixel_indices[pixel_axis] = pix_ind.reshape(shape)
+        world_coords = wcs.pixel_to_world_values(*pixel_indices)
+        # TODO: Remove NaNs??? These should not be here
+        if np.isnan(world_coords).any():
+            if isinstance(world_coords, tuple| list):
+                world_coords = [world_coord[~np.isnan(world_coord)] for world_coord in world_coords]
+            else:
+                world_coords = world_coords[~np.isnan(world_coords)]
```
nabobalis, line 466: TODO: Figure out why there are nans now

ndcube/ndcube.py
nabobalis: It might but I can't work out how.

ndcube/ndcube.py
```diff
@@ -443,31 +445,47 @@ def quantity(self):
...
+        if units:
+            mod = abs(wcs.world_n_dim - naxes) if wcs.world_n_dim > naxes else 0
```
nabobalis, line 468: This was my attempt at working out if there is a difference between world dims and pixel dims.

ndcube/ndcube.py
```diff
@@ -492,71 +551,108 @@ def _generate_world_coords(self, pixel_corners, wcs, needed_axes=None, *, units)
...
+        if needed_axes is None or len(needed_axes) == 0:
+            needed_axes = np.array(list(range(wcs.world_n_dim)),dtype=int)
+        axes_are_independent = []
+        pixel_axes = set()
+        for world_axis in needed_axes:
+            pix_ax = world_axis_to_pixel_axes(world_axis, wcs.axis_correlation_matrix)
+            axes_are_independent.append(len(pix_ax) == 1)
+            pixel_axes = pixel_axes.union(set(pix_ax))
+        if len(pixel_axes) == 1:
+            pixel_axes = list(pixel_axes)
```
DanRyanIrish, line 596:
```suggestion
        pixel_axes = list(pixel_axes)
```
I think this conversion should happen irrespective of the number of axes.

ndcube/ndcube.py
```diff
@@ -492,71 +551,108 @@ def _generate_world_coords(self, pixel_corners, wcs, needed_axes=None, *, units)
...
+        if all(axes_are_independent) and len(pixel_axes) == len(needed_axes):
+            world_coords = self._generate_independent_world_coords(pixel_corners, wcs, pixel_axes, units)
```
DanRyanIrish, line 598: If we keep the code in `_generate_independent_world_coords` exactly as it was before, then we should probably replace `pixel_axes` with `needed_axes` here.  Hopefully not changing the code, but just moving it, should avoid introducing that rebin problem.

ndcube/ndcube.py
```diff
@@ -492,71 +551,108 @@ def _generate_world_coords(self, pixel_corners, wcs, needed_axes=None, *, units)
...
+        else:
+            world_coords = self._generate_dependent_world_coords(pixel_corners, wcs, pixel_axes, units)
+        if len(world_coords) > 1 and isinstance(world_coords, tuple | list):
+            world_coords = [np.squeeze(world_coord) for world_coord in world_coords]
+        else:
+            world_coords = np.squeeze(world_coords)
```
DanRyanIrish, line 604:
```suggestion
```
I think we want to keep the returned type the same irrespective of how many coords are returned.  I suspect this presence of the code is requiring the hack you put in below.

DanRyanIrish (commented):
Initial comments.  Haven't looked at it all yet though.

ndcube/ndcube.py
nabobalis: I removed it

ndcube/ndcube.py
nabobalis: Yeah that fixed the rebin problem.

ndcube/ndcube.py
nabobalis: Updated.

nabobalis:
> Initial comments. Haven't looked at it all yet though.

I removed the workaround and went back to needed_axes. This has solves most of the problems but now the sliced tests all fail.

ndcube/tests/test_ndcube.py
```diff
@@ -252,10 +279,10 @@ def test_axis_world_coords_single_pixel_corners(axes, ndcube_3d_ln_lt_l):
                          indirect=("ndc",))
 def test_axis_world_coords_sliced_all_3d(ndc, item):
     coords = ndc[item].axis_world_coords_values()
-    assert u.allclose(coords, [1.02e-09, 1.04e-09, 1.06e-09, 1.08e-09] * u.m)
+    assert u.allclose(coords[0], [1.02e-09, 1.04e-09, 1.06e-09, 1.08e-09] * u.m)
 
     coords = ndc[item].axis_world_coords()
-    assert u.allclose(coords, [1.02e-09, 1.04e-09, 1.06e-09, 1.08e-09] * u.m)
+    assert u.allclose(coords[0], [1.02e-09, 1.04e-09, 1.06e-09, 1.08e-09] * u.m)
```
Cadair, line 285: Why have these asserts had to change?

nabobalis: I wish I knew, I assume since I broke the code.

nabobalis: But this is common in most of the other tests, you return a tuple of length N and you have to escape it.

nabobalis:
> @nabobalis The issue might be caused by an incorrect handling `ExtraCoords`. Can you confirm by restricting the condition to using your new implementation to when wcs is not an `ExtraCoords` object? If that fixes it, then we will have a strong indication of where the problem is.

Currently the "test_axis_world_coords_wave_ec" test passes but fails on the Extracoords calls when we pass that as the WCS, so that implies to me that you are correct. What I wrote does not handle EC correctly.

nabobalis:
And based on that, I think I worked out what I fucked up. The tests are passing now for me locally.

DanRyanIrish (approved):

PR #784, opened 2024-11-20
Some gwcs dev fixes

Cadair:
Should fix issues introduced by https://github.com/spacetelescope/gwcs/pull/457 and also https://github.com/spacetelescope/gwcs/pull/512

### DIFF
--- a/ndcube/extra_coords/table_coord.py
+++ b/ndcube/extra_coords/table_coord.py
@@ -919,22 +919,22 @@ class MultipleTableCoordinate(BaseTableCoordinate):
         dropped_world_dimensions["world_axis_names"] += [name or None for name in dropped_multi_table.frame.axes_names]
         dropped_world_dimensions["world_axis_physical_types"] += list(dropped_multi_table.frame.axis_physical_types)
         dropped_world_dimensions["world_axis_units"] += [u.to_string() for u in dropped_multi_table.frame.unit]
-        dropped_world_dimensions["world_axis_object_components"] += dropped_multi_table.frame._world_axis_object_components
-        dropped_world_dimensions["world_axis_object_classes"].update(dropped_multi_table.frame._world_axis_object_classes)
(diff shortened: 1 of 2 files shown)


ndcube/extra_coords/table_coord.py
```diff
@@ -919,22 +932,31 @@ def dropped_world_dimensions(self):
...
+        # In gwcs https://github.com/spacetelescope/gwcs/pull/457 the underscore was dropped
+        waocomp = getattr(dropped_multi_table.frame, "world_axis_object_components", getattr(dropped_multi_table.frame, "_world_axis_object_components", []))
+        dropped_world_dimensions["world_axis_object_components"] += waocomp
+        waocls = getattr(dropped_multi_table.frame, "world_axis_object_classes", getattr(dropped_multi_table.frame, "_world_axis_object_classes", {}))
+        dropped_world_dimensions["world_axis_object_classes"].update(waocls)
 
         for dropped in self._dropped_coords:
             # If the table is a tuple (QuantityTableCoordinate) then we need to
             # squish the input
+            # In gwcs https://github.com/spacetelescope/gwcs/pull/457 coordinate_to_quantity was removed
```
DanRyanIrish, line 944: If this is removed and not coming back, should we just commit to our own version, i.e. `self._from_high_level_coordinates`?  Is there any need to use `dropped.frame.from_high_level_coordinates` now, even if it exists because the user has an older version of `gwcs`?  Isn't the new code-flow using `self._from_high_level_coordinates` backwards compatible?

DanRyanIrish (commented):
Looks fine.  One question though.  See inline comment.

ndcube/extra_coords/table_coord.py
Cadair: Yeah it's not coming back. When we depend on a sufficiently new version of gwcs we can drop out all this and rely on the new stuff.

Cadair: Our method only gets used if the new one isn't there.

DanRyanIrish (approved):

Issue #785, opened 2024-11-20
NDCubeSequence.cube_like_dimensions Bug


PCJY:
### Describe the bug

When I call NDCubeSequence.cube_like_dimensions, it raises an error. With the following traceback:

```python
Traceback (most recent call last):
  File "C:\Users\junya\ndcube-sandbox\SliceSequence.py", line 66, in <module>
    print('hereeeeeee',my_sequence.cube_like_dimensions)
  File "C:\Users\junya\ndcube-git\ndcube\ndcube_sequence.py", line 87, in cube_like_dimensions
    dimensions = list(self._dimensions)
AttributeError: 'NDCubeSequence' object has no attribute '_dimensions'. Did you mean: 'dimensions'?
```

### To Reproduce

```python
import astropy.units as u

import astropy.wcs

import numpy as np

from ndcube import NDCube, NDCubeSequence

# Define data arrays.

shape = (4, 4, 5)

data0 = np.random.rand(*shape)

data1 = np.random.rand(*shape)

data2 = np.random.rand(*shape)

data3 = np.random.rand(*shape)

# Define WCS transformations. Let all cubes have same WCS.

wcs = astropy.wcs.WCS(naxis=3)

wcs.wcs.ctype = 'WAVE', 'HPLT-TAN', 'HPLN-TAN'

wcs.wcs.cunit = 'Angstrom', 'deg', 'deg'

wcs.wcs.cdelt = 0.2, 0.5, 0.4

wcs.wcs.crpix = 0, 2, 2

wcs.wcs.crval = 10, 0.5, 1

# Instantiate NDCubes.

cube0 = NDCube(data0, wcs=wcs)

cube1 = NDCube(data1, wcs=wcs)

cube2 = NDCube(data2, wcs=wcs)

cube3 = NDCube(data3, wcs=wcs)

# Instantiate NDCubeSequence. Let the common axis be 0.

my_sequence = NDCubeSequence([cube0, cube1, cube2, cube3], common_axis=0)

print(my_sequence.cube_like_dimensions)
```

### System Details

ndcube dev version

PR #788, opened 2024-11-25
Fixed a bug in NDCubeSequence.cube_like_dimensions.

PCJY:
## PR Description
fixes issue #785

### DIFF
--- a/changelog/788.bugfix.rst
+++ b/changelog/788.bugfix.rst
@@ -1,1 +1,1 @@
-Fixed a bug which caused `NDCubeSequence.cube_like_dimensions` to break.
+Fixed a bug which caused ``NDCubeSequence.cube_like_dimensions`` to break.
--- a/changelog/788.bugfix.rst
+++ b/changelog/788.bugfix.rst
@@ -1,0 +1,1 @@
+Fixed a bug in `NDCubeSequence.cube_like_dimensions`.
--- a/changelog/788.bugfix.rst
+++ b/changelog/788.bugfix.rst
@@ -1,1 +1,1 @@
-Fixed a bug in `NDCubeSequence.cube_like_dimensions`.
+Fixed a bug which caused `NDCubeSequence.cube_like_dimensions` to break.
--- a/ndcube/tests/test_ndcubesequence.py
+++ b/ndcube/tests/test_ndcubesequence.py
@@ -5,6 +5,7 @@ import warnings
 
 import astropy.units as u
 from astropy.time import Time, TimeDelta
+from astropy.tests.helper import assert_quantity_allclose
 
 from ndcube import NDCube, NDCubeSequence
 from ndcube.tests import helpers
@@ -143,8 +144,8 @@ def test_cube_like_shape(ndc, expected_shape):
 def test_cube_like_dimensions(ndc, expected_dimensions):
     with warnings.catch_warnings():
         warnings.simplefilter("ignore")
-        assert all(assert_quantity_allclose(ndc_dim, exp_dim) for ndc_dim, exp_dim in zip(ndc.cube_like_dimensions, expected_dimensions))
-    
+        for ndc_dim, exp_dim in zip(ndc.cube_like_dimensions, expected_dimensions):
+            assert_quantity_allclose(ndc_dim, exp_dim) 
 
 @pytest.mark.parametrize("ndc", (["ndcubesequence_4c_ln_lt_l"]), indirect=("ndc",))
 def test_cube_like_shape_error(ndc):
--- a/ndcube/tests/test_ndcubesequence.py
+++ b/ndcube/tests/test_ndcubesequence.py
@@ -1,6 +1,7 @@
 
 import numpy as np
 import pytest
+import warnings
 
 import astropy.units as u
 from astropy.time import Time, TimeDelta
@@ -140,8 +141,10 @@ def test_cube_like_shape(ndc, expected_shape):
                          ],
                          indirect=("ndc",))
 def test_cube_like_dimensions(ndc, expected_dimensions):
-    assert all(assert_quantity_allclose(ndc_dim, exp_dim) for ndc_dim, exp_dim in zip(ndc.cube_like_dimensions, expected_dimensions))
-
+    with warnings.catch_warnings():
+        warnings.simplefilter("ignore")
+        assert all(assert_quantity_allclose(ndc_dim, exp_dim) for ndc_dim, exp_dim in zip(ndc.cube_like_dimensions, expected_dimensions))
+    
 
 @pytest.mark.parametrize("ndc", (["ndcubesequence_4c_ln_lt_l"]), indirect=("ndc",))
 def test_cube_like_shape_error(ndc):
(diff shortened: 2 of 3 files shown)


PCJY:
Hi @DanRyanIrish Here is my first Pull Request, can you please review it?

nabobalis:
Hello @PCJY, does this happen on a release version of ndcube or just the main version?

DanRyanIrish:
Hi @PCJY.  Thanks so much for this PR.  I agree with @nabobalis that we should also have a unit test for this.  In addition, you should add a changelog file to describe the change that this PR makes.

For an example of a unit test, you can see [here](https://github.com/sunpy/ndcube/blob/main/ndcube/tests/test_ndcubesequence.py#L133).  It tests `NDCubeSequence.cube_like_shape`, so can be very closely adapted to this PR.

For a quick explainer of changelog files, see [here](https://github.com/sunpy/sunpy/tree/main/changelog#readme).  You should add the file to the [ndcube/changelog directory](https://github.com/sunpy/ndcube/tree/main/changelog).

nabobalis:
Does this change also need to occur to the new property as well?

DanRyanIrish:
> Does this change also need to occur to the new property as well?

Which new property @nabobalis ?

DanRyanIrish:
> `warn_deprecated("Replaced by ndcube.NDCubeSequence.cube_like_shape")` ? Or is the new one unitless?

Yes, the new one is unitless.

ndcube/tests/test_ndcubesequence.py
```diff
@@ -134,6 +134,15 @@ def test_cube_like_shape(ndc, expected_shape):
     assert np.all(ndc.cube_like_shape == expected_shape)
 
 
+@pytest.mark.parametrize(("ndc", "expected_dimensions"),
+                         [
+                             ("ndcubesequence_4c_ln_lt_l_cax1", tuple(u.Quantity(d, unit=u.pix) for d in [2. , 12, 4])),
+                         ],
+                         indirect=("ndc",))
+def test_cube_like_dimensions(ndc, expected_dimensions):
+    assert np.all(ndc.cube_like_dimensions == expected_dimensions)
```
DanRyanIrish, line 143: Astropy provides a [function](https://docs.astropy.org/en/latest/api/astropy.tests.helper.assert_quantity_allclose.html) to check two quantities are equal or almost equal with numerical precision.  I think that should be used here.  So something like 
```suggestion
    assert all(assert_quantity_allclose(ndc_dim, exp_dim) for ndc_dim, exp_dim in zip(ndc.cube_like_dimensions, expected_dimensions))
```

DanRyanIrish: If you would like to accept this change, you can do so on GitHub by clicking Commit suggestion.  Or, if you wanted to accepted multiple accepted changes, you can click add to batch (when looking at the "files" tab of the PR), and then commit all the suggested changes at once.

Just be sure to do `git pull me cubeLikeDimIssue` on your local machine afterwards to get this change on your own computer.

changelog/788.bugfix.rst
```diff
@@ -0,0 +1 @@
+Fixed a bug in `NDCubeSequence.cube_like_dimensions`.
```
DanRyanIrish, line 1: This should be a little more descriptive.  For example:

```suggestion
Fixed a bug whereby `NDCubeSequence.cube_like_dimensions` to break.
```

DanRyanIrish:
Hi @PCJY Thanks for these changes.  Currently, the test is failing because of the deprecation warning that this function raises.  Therefore, we have to make this test ignore such warnings.  @nabobalis, do you know how to ignore deprecation warnings for a specific test?

ndcube/tests/test_ndcubesequence.py
```diff
@@ -140,7 +140,7 @@ def test_cube_like_shape(ndc, expected_shape):
                          ],
                          indirect=("ndc",))
 def test_cube_like_dimensions(ndc, expected_dimensions):
-    assert np.all(ndc.cube_like_dimensions == expected_dimensions)
+    assert all(assert_quantity_allclose(ndc_dim, exp_dim) for ndc_dim, exp_dim in zip(ndc.cube_like_dimensions, expected_dimensions))
```
DanRyanIrish, line 143: To avoid the deprecation warning making this test fail, I think you should be able to use Python's [`catch_warnings` context manager](https://docs.python.org/3/library/warnings.html#temporarily-suppressing-warnings)

ndcube/tests/test_ndcubesequence.py
```diff
@@ -134,6 +136,17 @@ def test_cube_like_shape(ndc, expected_shape):
...
+@pytest.mark.parametrize(("ndc", "expected_dimensions"),
+                         [
+                             ("ndcubesequence_4c_ln_lt_l_cax1", tuple(u.Quantity(d, unit=u.pix) for d in [2. , 12, 4])),
+                         ],
+                         indirect=("ndc",))
+def test_cube_like_dimensions(ndc, expected_dimensions):
+    with warnings.catch_warnings():
+        warnings.simplefilter("ignore")
+        for ndc_dim, exp_dim in zip(ndc.cube_like_dimensions, expected_dimensions):
+            assert_quantity_allclose(ndc_dim, exp_dim)
```
DanRyanIrish, line 148: This looks like a perfectly good solution.  The issue before was that `assert_quantity_allclose` returns `None` if the values are the same, not true.  So an alternative fix would have been

```suggestion
        assert all(assert_quantity_allclose(ndc_dim, exp_dim) is None for ndc_dim, exp_dim in zip(ndc.cube_like_dimensions, expected_dimensions))
```

But if this works, we can keep your solution :)

DanRyanIrish:
The pre-commit test failed.  You should run `tox -e codestyle` locally an then commit and push the changes.

In addition, the "This branch is out-of-date with the base branch" warning below the tests means the new updates have been made to the main branch.  So in addition, yout should do `git pull origin main` locally and then push the changes, so the PR is up-to-date.

I'll look into the docs test fail later.

nabobalis:
Doc failure is `<unknown>:37: WARNING: py:obj reference target not found: NDCubeSequence.cube_like_dimensions [ref.obj]`

DanRyanIrish:
> Doc failure is `<unknown>:37: WARNING: py:obj reference target not found: NDCubeSequence.cube_like_dimensions [ref.obj]`

Thanks @nabobalis. Do you know what that means?

nabobalis:
> > Doc failure is `<unknown>:37: WARNING: py:obj reference target not found: NDCubeSequence.cube_like_dimensions [ref.obj]`
> 
> Thanks @nabobalis. Do you know what that means?

Yes I do.

changelog/788.bugfix.rst
```diff
@@ -0,0 +1 @@
+Fixed a bug which caused `NDCubeSequence.cube_like_dimensions` to break.
```
Cadair, line 1:
```suggestion
Fixed a bug which caused ``NDCubeSequence.cube_like_dimensions)` to break.
```

DanRyanIrish (approved):
Congrats @PCJY on your first approved PR to ndcube!!

PR #789, opened 2024-11-25
Update test_compound_wcs.py

nabobalis:
Fix test

### DIFF
--- a/ndcube/wcs/wrappers/tests/test_compound_wcs.py
+++ b/ndcube/wcs/wrappers/tests/test_compound_wcs.py
@@ -172,7 +172,7 @@ def test_shared_pixel_axis_compound_3d(spectral_cube_3d_fitswcs, time_1d_fitswcs
                                                              [False, True, False]])
 
     world = wcs.pixel_to_world_values(0, 0, 0)
-    if astropy.__version__ >= "7.0.0.dev":
+    if astropy.__version__ >= "7.0.0":
         np.testing.assert_allclose(world, (14, np.nan, np.nan, -7.0))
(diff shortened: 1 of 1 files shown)


braingram (approved):
LGTM.

ndcube/wcs/wrappers/tests/test_compound_wcs.py
```diff
@@ -172,7 +172,7 @@ def test_shared_pixel_axis_compound_3d(spectral_cube_3d_fitswcs, time_1d_fitswcs
                                                              [False, True, False]])
 
     world = wcs.pixel_to_world_values(0, 0, 0)
-    if astropy.__version__ >= "7.0.0.dev":
+    if astropy.__version__ >= "7.0.0":
```
wtbarnes, line 175: Is this string comparison doing the right thing? Shouldn't this use `packaging.version.Version`? https://packaging.pypa.io/en/stable/version.html

nabobalis: It does now without the dev tag. But you are correct but Stuart considered this string comparison enough so I left it. For he is the navigator that guides us through the warp.

braingram: [astropy.utils.minversion](https://docs.astropy.org/en/stable/api/astropy.utils.introspection.minversion.html) might also be useful although it's documentation for ".dev" versions is lacking.

I think the issue fixed here is for a string comparison "7.0.0.dev" is greater than "7.0.0" but for a semantic version comparison "7.0.0.dev" is less than "7.0.0".

Cadair (approved):

PR #794, opened 2024-12-11
Added new features to the ndcube.__add__ method

PCJY:
## PR Description

write down the scenarios.

(list, to see what has been covered)

### Testing scenarios without mask
redrafting the tests.

- One and only one of them has a unit. [Error]

- Both NDCube and NDData have unit 
                       Both have uncertainty.
                       NDCube has uncertainty.
                       NDData has uncertainty.
                       Neither has uncertainty.
- Neither has a unit.
                       Both have uncertainty.
                       NDCube has uncertainty.
                       NDData has uncertainty.
                       Neither has uncertainty.

name them again:
test_cube_add_cube_unit_mask_nddata_unc_unit_mask


ndcube/ndcube.py
```diff
@@ -918,6 +919,37 @@ def __neg__(self):
         return self._new_instance(data=-self.data)
 
     def __add__(self, value):
+        if isinstance(value, NDData) and value.wcs is None:
+            if self.unit is not None and value.unit is not None:
+                    value_data = value.data * value.unit.to(self.unit)
```
DanRyanIrish, line 924: I think to be consistent, you should strip the unit after conversion here by using `.to_value` rather than `.to`
```suggestion
                    value_data = value.data * value.unit.to_value(self.unit)
```

ndcube/ndcube.py
```diff
@@ -918,6 +919,37 @@ def __neg__(self):
...
+            # combine the uncertainty
+            new_uncertainty = None
+            if self.uncertainty is not None and value.uncertainty is not None:
+                new_uncertainty = self.uncertainty.propagate(
+                    np.add, value.uncertainty, correlation=0
+                )
+            elif self.uncertainty is not None:
+                new_uncertainty = self.uncertainty
+            elif value.uncertainty is not None:
+                new_uncertainty = value.uncertainty
```
DanRyanIrish, line 941: Alternatively, you can use `else` to set `new_uncertainty` to `None`.
```suggestion
            if self.uncertainty is not None and value.uncertainty is not None:
                new_uncertainty = self.uncertainty.propagate(
                    np.add, value.uncertainty, correlation=0
                )
            elif self.uncertainty is not None:
                new_uncertainty = self.uncertainty
            elif value.uncertainty is not None:
                new_uncertainty = value.uncertainty
            else:
                new_uncertainty = None
```

ndcube/ndcube.py
```diff
@@ -918,6 +919,37 @@ def __neg__(self):
...
+        if isinstance(value, NDData) and value.wcs is None:
+            if self.unit is not None and value.unit is not None:
+                    value_data = value.data * value.unit.to(self.unit)
+            elif self.unit is None:
+                value_data = value.data
+            else:
+                raise TypeError("Cannot add unitless NDData to a unitful NDCube.")
+
+            # addition
+            new_data = self.data + value_data
```
DanRyanIrish, line 931: The addition should be done as part of the masked array addition.  You've already done this below, you just need to extract the added data from the results as well as the mask.

ndcube/ndcube.py
```diff
@@ -918,6 +919,37 @@ def __neg__(self):
...
+            # combine mask
+            self_ma = np.ma.MaskedArray(self.data, mask=self.mask)
+            value_ma = np.ma.MaskedArray(value_data, mask=value.mask)
+            result_ma = self_ma + value_ma
+            new_mask = result_ma.mask
+
+            # return the new NDCube instance
+            return self._new_instance(
+                data=new_data, uncertainty=new_uncertainty, mask=new_mask
+            )
```
DanRyanIrish, line 952: Instead of having a separate `return` here for the `NDData` case, I think we should build a dictionary of `kwargs` that we can give `self._new_instance`, [here](https://github.com/sunpy/ndcube/pull/794/files#diff-929794bb532be1486907b360285433b432f8505ab7e7ff63eca3eab657f7d00dR969).  So, you can create an empty kwargs dictionary at the start of the method, and add the new data, uncertainty, etc. in the relevant place, e.g.
```python
kwargs["uncertainty"] = new_uncertainty
```
Then the final line of the method would become
```python
return self._new_instance(**kwargs)
```

DanRyanIrish: Let me know if this doesn't make sense

ndcube/ndcube.py
```diff
@@ -1035,37 +1035,43 @@ def __neg__(self):
...
             self_ma = np.ma.MaskedArray(self.data, mask=self.mask)
             value_ma = np.ma.MaskedArray(value_data, mask=value.mask)
+
+            # addition
             result_ma = self_ma + value_ma
-            new_mask = result_ma.mask
+
+            # extract new mask and new data
+            kwargs["mask"] = result_ma.mask
+            kwargs["data"] = result_ma.data
```
DanRyanIrish, line 1070: As mentioned in above comment, I think it makes sense to do this before the uncertainty propagation so you can use the `kwargs["data"]` value in that propagation.

ndcube/ndcube.py
```diff
@@ -1034,6 +1035,43 @@ def __neg__(self):
...
+
+            # addition
+            result_ma = self_ma + value_ma
+
+            # extract new mask and new data
+            kwargs["mask"] = result_ma.mask
+            kwargs["data"] = result_ma.data
+
+            # return the new NDCube instance
+            return self._new_instance(**kwargs)
```
DanRyanIrish, line 1073: Move this line to the [end of the method](https://github.com/sunpy/ndcube/pull/794/files#diff-929794bb532be1486907b360285433b432f8505ab7e7ff63eca3eab657f7d00dR1091) and use the `kwargs` approach when handling the other cases, e.g. `Quantity`.  So, for example, [L1082](https://github.com/sunpy/ndcube/pull/794/files#diff-929794bb532be1486907b360285433b432f8505ab7e7ff63eca3eab657f7d00dR1082) would become:
```python
kwargs["data"] = self.data + value.to_value(cube_unit)
```

DanRyanIrish:
A changelog file needs to be added.

And your branch needs to be updated with the latest version of main.

PCJY:
Hi @DanRyanIrish, as we have discussed in our project meetings, below are the issues we encountered and may need further discussions with others in the community:

The issue is mainly around how NumPy handles masks when performing an addition for two NumPy.MaskedArray.
We think the expected outcome for an addition should be: the sum of any value that is not masked by its individual mask.
E.g. [1] ([T]) + [2] ([F]) = [2].

However, from experimentation, it can be seen that
 NumPy returns in this way:
 [1] ([T]) + [2] ([F]) = [1].
Screenshot 2025-01-18 220004

I find this confusing because even if it does combine the mask and then apply it on the result, it should be:
 [1] ([T]) + [2] ([F]) = [-].

Please correct me if there is anything wrong in my understanding.

PCJY:
Secondly, we also encountered some issues around the propagate method: 
it ignores the mask of the objects that are passed in, and still takes into account the uncertainties of the masked elements when it should not have done so.  
Following your guidance and suggestions, this issue is currently being worked on by setting the corresponding entries that should be masked of the uncertainty array to be 0, before passed in to the propagate method.
A clearer example of the issue was implemented as shown in the code below with the screenshot of its output attached.

```
from ndcube import NDCube
import numpy as np
from astropy.nddata import StdDevUncertainty
from astropy.wcs import WCS

data = np.array([[1, 2], [3, 4]])  
uncertainty = StdDevUncertainty(np.array([[0.1, 0.2], [0.3, 0.4]])) 
mask = np.array([[False, True], [False, False]])  
wcs1 = WCS(naxis=2) 
wcs1.wcs.ctype = ["HPLT-TAN", "HPLN-TAN"]

cube = NDCube(data, wcs=wcs1, uncertainty=uncertainty, mask=mask)
print(cube)

def add_operation(cube1, cube2):
    """
    Example function to add two data arrays with uncertainty propagation.
    """
    result_data = cube1.data + cube2.data 
    # Propagate the uncertainties using the NDCube objects
    propagated_uncertainty = cube1.uncertainty.propagate(
        np.add, cube2, result_data=result_data, correlation = 0
    )
    return result_data, propagated_uncertainty

# adding the cube to itself
result_data, propagated_uncertainty = add_operation(cube, cube)

print("Original Data:\n", cube.data)
print("Original Uncertainty:\n", cube.uncertainty.array)
print("Result Data (after addition):\n", result_data)
print("Propagated Uncertainty:\n", propagated_uncertainty.array)
```
Screenshot 2025-01-18 222146

DanRyanIrish:
> I find this confusing because even if it does combine the mask and then apply it on the result, it should be: [1] ([T]) + [2] ([F]) = [-].

This is where I also find numpy masked arrays counter-intuitive.  However, the logic is as follows:
* If one pixel is masked, retain the data of the left-hand operand and set the mask value to `True`.
Notice that order of the operands matters.  Because you did `[1] ([T]) + [2] ([F])`, the results is `[1] ([T])`, which is displayed as `[--]`.  I would expect if you did the operation the other way around (`[2] ([F]) + [1] ([T])`), the result would be `[2] ([T])`.

Notice that this is not the same as the scheme I've proposed in my previous comment, in part because it's confusing, as you've found.

DanRyanIrish:
@PCJY, until we agree a way forward with the mask, you should proceed by implementing in the case for which neither object has a mask,

Cadair:
> I propose the following scheme

I haven't thought too much each of these individual cases, but the fact there is a list is enough to make me think we probably need a way for the user to choose. This is obviously not possible with the `__add__` operator, so we would need to have a `NDCube.add` method (and presumably `subtract`) which accepted kwargs.

Is this also not a problem for other operators as well?

DanRyanIrish:
I think this is a good idea.  As well as `add` and `subtract`, I think we would also need `multiply` and `divide` methods.

As far as I can see, this ambiguity only arises when there are masks involved.  So we could still implement the dunder methods as wrappers around the above methods, but have the raise and error/return `NotImplemented` if the non-`NDCube` operand has a mask, and require users use the `NDCube.add` method instead.

Cadair:
I am not sure how I feel about adding methods to the API for every arithmetic operation. How about functions instead? `import ndcube; ndcube.add(cube1, cube2)`?

DanRyanIrish:
> I am not sure how I feel about adding methods to the API for every arithmetic operation. How about functions instead? `import ndcube; ndcube.add(cube1, cube2)`?

I'm open to considering that.  It might go well in that analysis subpackage I've proposed.  I can also see an argument for making it a method as the first argument must always be the `NDCube` and it's/they are operation(s) valid for any cube.  It needs more thought.

Either way, a method can be converted to a function and vice versa quite easily.  So for now, I think we should proceed with a `NDCube.add` method until a longer-term design decision is made.  We should discuss this on the sunpy call.

ndcube/ndcube.py
```diff
@@ -964,24 +965,91 @@ def _new_instance(self, **kwargs):
...
+        """
+        Users are allowed to choose whether they want operation_ignores_mask to be True or False,
+        and are allowed to choose whether they want handle_mask to be AND / OR .
+        """
+        kwargs = {}
+
+        if isinstance(value, NDData) and value.wcs is None:
+            if self.unit is not None and value.unit is not None:
+                value_data = (value.data * value.unit).to_value(self.unit)
+            elif self.unit is None:
```
DanRyanIrish, line 978:
```suggestion
            elif self.unit is None and value.unit is None:
```

DanRyanIrish (changes requested):

ndcube/ndcube.py
```diff
@@ -964,24 +965,91 @@ def _new_instance(self, **kwargs):
...
+            else:
+                # TODO
+                # When there is a mask, that is when the two new added parameters (OIM and HM) come into the picture.
+                # Conditional statements to permutate the two different scenarios (when it does not ignore the mask).
+                if operation_ignores_mask is False:
+                    if handle_mask is np.logical_and:
+                        pass
+                    else:
+                        pass
+                else:
```
DanRyanIrish, line 1014:
```suggestion
```
For simplicity let's remove this code for now and just focus on the case where there are no masks.  This should make debugging the tests easier.

ndcube/tests/test_ndcube.py
```diff
@@ -1138,6 +1138,26 @@ def test_cube_arithmetic_add(ndcube_2d_ln_lt_units, value):
     check_arithmetic_value_and_units(new_cube, cube_quantity + value)
 
 
+@pytest.mark.parametrize('value', [
+    NDData(np.random.rand(10, 12),
+           unit=u.ct,
+           wcs=None,
+           uncertainty=StdDevUncertainty(np.random.rand(10, 12)),
+           mask=np.random.choice([False, False], size=(10, 12))),
```
DanRyanIrish, line 1146:
```suggestion
```

DanRyanIrish:
I'm not sure at first glance why this test is failing.  I have made some inline suggestions on chages you should make to the code.  (You can accept the changes on GitHub, and then pull them you your local branch.)

Once you've made those changes, and assuming the test is still failing, you figure out each step the code takes when the test is being run, and find where the problem is.  You can try printing out things at various points in the code and running pytest.  See this sunpy explainer: https://docs.sunpy.org/en/latest/dev_guide/contents/tests.html#id2
Alternatively, you can use the `pytest --pdb` as explained in that sunpy explainer.  This will pause the code in an interactive session where the error is raised and you can explore the values of current variables and try running specific lines of code.

DanRyanIrish (changes requested):

DanRyanIrish:
This is looking good.  The `test_cube_add_uncertainty` should also check the other attributes are as expected, i.e., the data and the unit.  For this, it might be better to not initiate your fixture with `np.random.rand`, but a known set of values, e.g. `np.ones`.

ndcube/tests/test_ndcube.py
```diff
@@ -1138,6 +1138,29 @@ def test_cube_arithmetic_add(ndcube_2d_ln_lt_units, value):
...
+    new_cube = ndcube_2d_with_uncertainty + value
+    # Check uncertainty propagation
+    expected_uncertainty = ndcube_2d_with_uncertainty.uncertainty.propagate(
+                            operation=np.add,
+                            other_nddata=value,
+                            result_data=new_cube.data * new_cube.unit,
+                            correlation=0,
+    )
+    if expected_uncertainty.unit is None:
+        expected_uncertainty = StdDevUncertainty(expected_uncertainty.array, unit=new_cube.unit)
```
PCJY, line 1158:
```suggestion
        expected_uncertainty = StdDevUncertainty(expected_uncertainty.array, unit=new_cube.unit)
    assert np.allclose(new_cube.data, ndcube_2d_with_uncertainty.data + value.data)
    assert new_cube.unit == u.ct
```

ndcube/ndcube.py
```diff
@@ -964,24 +965,85 @@ def _new_instance(self, **kwargs):
...
+                raise TypeError("Cannot add unitless NDData to a unitful NDCube.")
+
+            # check whether there is a mask.
+            # Neither self nor value has a mask
+            self_unmasked = self.mask is None or self.mask is False or not self.mask.any()
+            value_unmasked = value.mask is None or value.mask is False or not value.mask.any()
+
+            if (self_unmasked and value_unmasked) or operation_ignores_mask is True:
+                # addition
+                kwargs["data"] = (self.data + value_data) * self.unit
```
DanRyanIrish, line 990:
```suggestion
                kwargs["data"] = self.data + value_data
                kwargs["unit"] = self.unit
```

DanRyanIrish, line 990:
```suggestion
                kwargs["data"] = self.data + value_data
```

ndcube/ndcube.py
```diff
@@ -987,7 +987,7 @@ def add(self, value, operation_ignores_mask=True, handle_mask=np.logical_and):
 
             if (self_unmasked and value_unmasked) or operation_ignores_mask is True:
                 # addition
-                kwargs["data"] = self.data + value_data
+                kwargs["data"] = (self.data + value_data)*self.unit
```
DanRyanIrish, line 990:
```suggestion
                kwargs["data"] = self.data + value_data
                kwargs["unit"] = self.unit
```

PCJY:
Notes for the current stage of development:
1, The 'TypeError: None is not a valid Unit' occurs whenever an NDCube/NDData object with a unit enters the propagate method. 
Their values become None after they enter the propagate method. Their values remain correct before entering the propagate method.
It might happen due to how the unit is handled or converted inside the propagate method.

result_unit_sq is not set correctly, in _propagate_add_sub() .

TODO: 1) look a bit more into the stack trace that pytest --pdb, w, returns. 2) using pytest --pdb, u, check how result_unit_sq is assigned.

2, The test cases testing the case when neither of them has a unit, and that when only the NDData has a unit, have been implemented and passed. TODO: add the test case for when only NDCube has a unit.

DanRyanIrish:
Here is an example of including fixtures in parameterisations: https://github.com/sunpy/ndcube/blob/main/ndcube/tests/test_ndcube.py#L48
The trick is to add this `indirect` kwarg at the end of the parameterisation given the name of the input variable(s) that has to be retrieved from conftext.py

ndcube/tests/test_ndcube.py
```diff
@@ -1179,73 +1152,170 @@ def cube(request):
...
+                        ],
+                        indirect=("ndc",))
+def test_cube_add_cube_unit_unc_nddata_unit_unc(ndc, value):
+    new_cube = ndc + value # perform the addition
     # Check uncertainty propagation
-    expected_uncertainty = ndcube_2d_with_unit_uncertainty.uncertainty.propagate(
+    expected_uncertainty = ndc.uncertainty.propagate(
                             operation=np.add,
                             other_nddata=value,
                             result_data=new_cube.data*new_cube.unit,
```
DanRyanIrish, line 1180: This test is a bit circular.  You shouldn't use outputs of the test to form expected values.  Otherwise the test may pass when it shouldn't.
```suggestion
    expected_unit = u.ct
    expected_data = (ndc.data * ndc.unit).to_value(expected_unit) + (value.data * value.unit).to_value(expected_unit)
    expected_uncertainty = ndc.uncertainty.propagate(
                            operation=np.add,
                            other_nddata=value,
                            result_data=expected_data*expected_unit,
```

ndcube/tests/test_ndcube.py
```diff
@@ -1179,73 +1152,170 @@ def cube(request):
...
                             correlation=0,
     )
-    assert np.allclose(new_cube.data, ndcube_2d_with_unit_uncertainty.data + value.data)
-    assert new_cube.unit == u.ct # sometimes explicit, in case the other part is also wrong.
+    assert np.allclose(new_cube.data, ndc.data + value.data)
+    assert new_cube.unit == u.ct
```
DanRyanIrish, line 1184: Using above suggested changes:
```suggestion
    assert np.allclose(new_cube.data, expected_data)
    assert new_cube.unit == expected_unit
```

DanRyanIrish (changes requested):
The test function names are now much clearer, and it's easier to see the cases that are tested.

A couple more comments on your test writing:

1. At least one of your tests is somewhat circular, i.e. you use the result of the operation you're testing to form the expected values.  This means that an error might make it into the expected values and the test will pass when it should fail.  See inline comments for a better way to structure tests.  These are for information only.  There is an even better way to structure these tests.  See item below.

2. ndcube provides a [test helper function](https://github.com/sunpy/ndcube/blob/main/ndcube/tests/helpers.py#L125) that checks if two cubes are equal.  This means your test only needs to construct the expected cube, and then pass it and the result of the addition to that helper function.  See below for an example of how to do this for one test

```python
def test_cube_add_cube_unit_unc_nddata_unit_unc(ndc, value):
    output_cube = ndc + value # perform the addition
    # Construct expected cube
    expected_unit = u.ct
    expected_data = ((ndc.data * ndc.unit) + (value.data * value.unit)).to_value(expected_unit)
    expected_uncertainty = ndc.uncertainty.propagate(
                            operation=np.add,
                            other_nddata=value,
                            result_data=expected.data*expected.unit,
                            correlation=0,
    )
    expected_cube = NDCube(expected_data, ndc.wcs, uncertainty=expected_uncertainty, unit=expected_unit)
    # Assert output cube is same as expected cube
    assert_cubes_equal(output_cube, expected_cube)
```

All your tests in this PR should be written like this to ensure the output cubes are fully tested.  It should also make them easier to read.

ndcube/tests/test_ndcube.py
PCJY: @DanRyanIrish Thank you for the suggestion. But for this particular suggestion, if I use assert_cubes_equal(), is it still necessary to execute these two assert commands?

ndcube/tests/test_ndcube.py
PCJY: Hi @DanRyanIrish , I checked the code for the assert_cubes_equal method. I am unsure whether it checks the values, type and units of the uncertainty attributes as well. 
It looks like it only checks whether the shapes of uncertainties are the same.

PCJY: @DanRyanIrish Or maybe I can 
either: use the assert_cube_equal method together with a few more lines that checks the values, type and units of the uncertainty attributes, 
or: adding a few more lines into the assert_cube_equal method?

PCJY: I see what you mean, thank you for this suggestion! I will implement this.

ndcube/tests/helpers.py
```diff
@@ -122,13 +122,21 @@ def assert_metas_equal(test_input, expected_output):
...
     assert_wcs_are_equal(test_input.wcs, expected_cube.wcs)
-    if test_input.uncertainty:
+    if check_uncertainty_values:
+        # Check output and expected uncertainty are of same type. Remember they could be None.
+        # If the uncertainties are not None,...
+        # Check units, shape, and values of the uncertainty.
+        if (test_input.uncertainty is not None and expected_cube.uncertainty is not None):
+            assert type(test_input.uncertainty) is type(expected_cube.uncertainty)
+            assert np.allclose(test_input.uncertainty.array, expected_cube.uncertainty.array), \
+                f"Expected uncertainty: {expected_cube.uncertainty}, but got: {test_input.uncertainty.array}"
```
DanRyanIrish, line 138:
```suggestion
                f"Expected uncertainty: {expected_cube.uncertainty}, but got: {test_input.uncertainty.array}"
            else:
                assert (test_input.uncertainty is None and expected_cube.uncertainty is None)
```

DanRyanIrish:
Hi @PCJY.  The tests are looking good now.  However, could you please rename them `test_arithmetic_add_...`, rather than `test_cube_add...`, as the test actually uses the arithmetic operator `+`, and only tests the `NDCube.add` method indirectly.

ndcube/tests/helpers.py
```diff
@@ -136,8 +136,14 @@ def assert_cubes_equal(test_input, expected_cube, check_data=True, check_uncerta
...
             assert np.allclose(test_input.uncertainty.array, expected_cube.uncertainty.array), \
                 f"Expected uncertainty: {expected_cube.uncertainty}, but got: {test_input.uncertainty.array}"
+        else:
+            if test_input.uncertainty is None:
+                assert expected_cube.uncertainty is None, "Expected cube's uncertainty should also be None"
+            elif expected_cube.uncertainty is None:
+                assert test_input.uncertainty is None, "test_input should also be None"
             else:
-                assert (test_input.uncertainty is None and expected_cube.uncertainty is None)
+                pass
```
DanRyanIrish, line 145: Now the way you have it, you could do this:
```suggestion
                assert type(test_input.uncertainty) is type(expected_cube.uncertainty)
                assert np.allclose(test_input.uncertainty.array, expected_cube.uncertainty.array), \
                f"Expected uncertainty: {expected_cube.uncertainty}, but got: {test_input.uncertainty.array}"
```
and then remove lines 135-139.  Up to you though.

PCJY: @DanRyanIrish Thank you for your suggestion. I re-understood what the code aims to do, and changed it again.

PCJY:
> Hi @PCJY. I've edited your mask discussion in the PR description. I've created a table for all the scenarios for determining the result of adding the masked data. Can you fill in the table with the result of each case? Either `1`, `2`, or `3`? Once you've done that, we can look at the table and, like before, determine how to implement them, and if we have some effectively duplicated cases.

Hi @DanRyanIrish, thank you for providing the logic structures. I have filled in the results, please feel free to tell me whether I made any mistakes. (I showed my understanding process in the comments as well.)

DanRyanIrish:
Hi @PCJY Regarding you question in the PR description:
> My understanding:
> handle_mask is a function, as long as it has a value, it will be True, and does the operation on the two masks, otherwise, the new_mask value would be None, meaning:
> 1), the user did not set anything for the handle_mask kwarg, should an error be raised here?
> 2), there is no need to set any value for the handle_mask kwarg.

Yes, that's right.  Regarding your two options, `handle_mask` will have a default value.  (We can decide what that should be later.) But that means the user doesn't have to set it, so there's no need to raise an error.  The default value will trigger the default mask handling behaviour.

DanRyanIrish:
@PCJY: Regarding your comments on the distinct cases in the PR description:
> When OIM is T, the result is always 3, i.e. the actual result of the addition of the two data values.

Yes :)

> When OIM is F, the result is always the addition result of any value with its corresponding mask being F

Yes!

> (when there is not one, the result is None).

Yet to be decided

So I think a way to implement this would be:
```python
self_data, value_data = self.data, value.data # May require a copy
self_mask, value_mask = self.mask, value.mask # May require handling/converting of cases when masks aren't boolean arrays but are None, True, or False.
if not operation_ignores_mask:
    no_op_value = 0 # Value to set masked values since we are doing addition. (Would need to be 1 if we were doing multiplication.)
    self_data[self_mask] = no_op_value
    value_data[value_mask] = no_op_value
    both_masked = np.logical_and(self_mask, value_mask)
    if both_masked.any():
        # Handle case when both values are masked here, e.g.
        # self_data[both_masked], value_data[both_masked] = self.data[self_data], value.data[self_data] # to effectively ignore masks, or
        # self_data[both_masked], value_data[both_masked] = np.nan, np.nan # To make operation undefined.
        # We are yet to decide the best behaviour here.
else:
    pass # If operation ignores mask, nothing needs to be done. This line not needed in actual code.  Only here for clarity.

# Perform addition
new_data = self_data + value_data
# Calculate new mask.
new_mask = handle_mask(self.mask, value.mask) if handle_mask else None
```

PCJY:
Hi @DanRyanIrish , I have pushed four new commits. Any suggestions would be appreciated.
1, I added three different conditional blocks for the three with-mask scenarios, and to set the uncertainty and mask value for the addition result.

2, I added a new method named combine_uncertainty(), since the code for uncertainty setting is used repeatedly. But I am not sure whether I should add a new method inside the NDCube class.

3, I wrote the tests for the three different with-mask scenarios. But I think the uncertainty testing is not written in the right way. I might need some suggestions on this, thank you! 

4, To work on the __mul__ method, I wonder whether I should create a new PR or add a new method within the NDCube class in this branch, please let me know what you think, thank you!

ndcube/ndcube.py
```diff
@@ -964,24 +966,98 @@ def _new_instance(self, **kwargs):
...
+            elif (value_unmasked and not self_unmasked):
+                kwargs["mask"] = self.mask
+                # combine the uncertainty
+                kwargs["uncertainty"] = self.combine_uncertainty(value, result_data)
+
+            else:
+                kwargs["mask"] = handle_mask(self.mask,value.mask)
+                # combine the uncertainty
+                kwargs["uncertainty"] = self.combine_uncertainty(value, result_data)
+
```
DanRyanIrish, line 1010: This can all be simplified to:
```suggestion
            kwargs["uncertainty"] = self._combine_uncertainty(value, result_data)
            # Handle mask.
            self_unmasked = self.mask is None or self.mask is False or not self.mask.any()
            value_unmasked = value.mask is None or value.mask is False or not value.mask.any()
            if self_unmasked and value_unmasked:
                pass
            elif value_unmasked:
                kwargs["mask"] = value.mask
            elif self_unmasked:
                kwargs["mask"] = self.mask
            else:
                kwargs["mask"] = handle_mask(self.mask, value.mask)

```
See comment below for reason for changing `combine_uncertainty` to `_combine_uncertainty`.

ndcube/ndcube.py
```diff
@@ -964,24 +966,98 @@ def _new_instance(self, **kwargs):
...
+        self_masked = not(self.mask is None or self.mask is False or not self.mask.any()) if hasattr(self, "mask") else False
+        value_masked = not(value.mask is None or value.mask is False or not value.mask.any()) if hasattr(value, "mask") else False
+
+        if  (value_masked or (self_masked and hasattr(value,'uncertainty') and value.uncertainty is not None)): # value has a mask,
+            # let the users call the add method, since the handle_mask keyword cannot be given by users here.
+            raise TypeError('Please use the add method.')
+
+        return self.add(value) # without any mask, the add method can be called here and will work properly without needing arguments to be passed.
+
+    def combine_uncertainty(self, value, result_data):
```
DanRyanIrish, line 1047: This should be a hidden method, i.e. not intended for users.  Therefore, it should start with an `_`
```suggestion
    def _combine_uncertainty(self, value, result_data):
```

DanRyanIrish (commented):
Initial thoughts after quick glance at changes.

PCJY:
Test Scenarios with mask

- NDCube has mask but NDData does not


- NDCube does not have mask but NDData does 


- Both NDCube and NDData have mask

DanRyanIrish (approved):

PR #795, opened 2024-12-12
Add a data setter

### LINKED ISSUE #779: Add a way of setting `.data`
### Describe the feature

I frequently have the case when working with DKIST data that I want to in-place compute the dask array, i.e. this:
```python
spatial_avg = visp.rebin((-1, 1, -1), ).squeeze()
spatial_avg._data = spatial_avg.data.compute()
```
it would be nice if I didn't have to use `._data`.

### Proposed solution

Add a setter for `.data` which validates the new array is the same shape as the old array.

Cadair:
fixes #779 

Main open question is if we should allow changing the unit by setting the data with a quantity.

### DIFF
--- a/ndcube/conftest.py
+++ b/ndcube/conftest.py
@@ -575,6 +575,7 @@ def ndc(request):
 # NDCubeSequence Fixtures
 ################################################################################
 
+
 @pytest.fixture
 def ndcubesequence_4c_ln_lt_l(ndcube_3d_ln_lt_l):
     cube1 = ndcube_3d_ln_lt_l
--- a/ndcube/conftest.py
+++ b/ndcube/conftest.py
@@ -575,7 +575,6 @@ def ndc(request):
 # NDCubeSequence Fixtures
 ################################################################################
 
-
 @pytest.fixture
 def ndcubesequence_4c_ln_lt_l(ndcube_3d_ln_lt_l):
     cube1 = ndcube_3d_ln_lt_l
--- a/ndcube/ndcube.py
+++ b/ndcube/ndcube.py
@@ -401,17 +401,37 @@ class NDCubeBase(NDCubeABC, astropy.nddata.NDData, NDCubeSlicingMixin):
 
     @property
     def data(self):
+        """
+        `~numpy.ndarray`-like : The stored dataset.
+
+        Notes
+        -----
+        It is possible to set the ``.data`` attribute on a `NDCube` with an
+        array-like object of the same shape. However, this is really only
+        intended for replacing the data with a different object representing
+        the same physical data as no other properties of the cube will be
+        changed, such as uncertainty or unit.
+        """
         return super().data
 
     @data.setter
     def data(self, value):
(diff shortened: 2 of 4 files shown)


ndcube/conftest.py
```diff
@@ -575,7 +575,6 @@ def ndc(request):
 # NDCubeSequence Fixtures
 ################################################################################
 
-
 @pytest.fixture
```
Cadair, line 578:
```suggestion

@pytest.fixture
```

ndcube/ndcube.py
```diff
@@ -399,6 +399,42 @@ def __init__(self, data, wcs=None, uncertainty=None, mask=None, meta=None,
...
+        """
+        return super().data
+
+    @data.setter
+    def data(self, value):
+        # In an array-agnostic way check the shape is the same
+        if not hasattr(value, "shape") or value.shape != self.data.shape:
+            raise TypeError(f"Can only set data with an array-like object of the same shape ({self.data.shape})")
+
+        # Other masked arrays are hard to detect reliably
```
DanRyanIrish, line 423: So what happens if `value` a different type of masked array?  Does `.data` become that type of masked array?

ndcube/tests/test_ndcube.py
```diff
@@ -1238,3 +1238,64 @@ def test_ndcube_quantity(ndcube_2d_ln_lt_units):
...
+
+
+def test_quantity_data_setter(ndcube_2d_ln_lt_units):
+    cube = ndcube_2d_ln_lt_units
+    assert cube.unit
+
+    new_data = np.zeros_like(cube.data) * cube.unit
+    cube.data = new_data
+
+    assert isinstance(cube.data, np.ndarray)
```
DanRyanIrish, line 1276:
```suggestion
    assert isinstance(cube.data, np.ndarray)
    np.testing.assert_allclose(cube.data, new_data.value)
```

ndcube/ndcube.py
Cadair: yes

Cadair: Probably the same thing.

Cadair: yes, the same thing:
```
In [10]: NDCube(data=dask.array.ma.zeros_like(np.array((10,10))), wcs=WCS()).data
Out[10]: dask.array<<numpy.ma.core._convert2ma object at 0x7f8144495c5, shape=(2,), dtype=int64, chunksize=(2,), chunktype=numpy.MaskedArray>
```

this is just unsupported anywhere.

ndcube/ndcube.py
```diff
@@ -399,6 +399,42 @@ def __init__(self, data, wcs=None, uncertainty=None, mask=None, meta=None,
...
+        if isinstance(value, np.ma.MaskedArray):
```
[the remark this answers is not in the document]
DanRyanIrish: OK then.

DanRyanIrish (approved):

PR #798, opened 2024-12-17
Reduce memory consuption in axis_world_coords (again)

Cadair:
This PR works, as an alternative to #780 by not generating mesh coords for unneeded pixel axes.

i.e. if the pixel axes are not correlated with a requested world axis then don't generate the coords in the mesh.

### DIFF
--- a/ndcube/conftest.py
+++ b/ndcube/conftest.py
@@ -292,6 +292,86 @@ def wcs_3d_ln_lt_t_rotated():
     return WCS(header=h_rotated)
 
 
+@pytest.fixture
+def wcs_3d_ln_lt_l_coupled():
+    # WCS for a 3D data cube with two celestial axes and one wavelength axis.
+    # The latitudinal dimension is coupled to the third pixel dimension through
+    # a single off diagonal element in the PCij matrix
+    header = {
+        'CTYPE1': 'HPLN-TAN',
+        'CRPIX1': 5,
+        'CDELT1': 5,
+        'CUNIT1': 'arcsec',
+        'CRVAL1': 0.0,
+
+        'CTYPE2': 'HPLT-TAN',
+        'CRPIX2': 5,
+        'CDELT2': 5,
+        'CUNIT2': 'arcsec',
+        'CRVAL2': 0.0,
+
+        'CTYPE3': 'WAVE',
+        'CRPIX3': 1.0,
+        'CDELT3': 1,
+        'CUNIT3': 'Angstrom',
+        'CRVAL3': 1.0,
+
+        'PC1_1': 1,
+        'PC1_2': 0,
+        'PC1_3': 0,
+        'PC2_1': 0,
+        'PC2_2': 1,
+        'PC2_3': -1.0,
+        'PC3_1': 0.0,
+        'PC3_2': 0.0,
+        'PC3_3': 1.0,
+
+        'WCSAXES': 3,
+
+        'DATEREF': "2020-01-01T00:00:00"
+    }
+    return WCS(header=header)
+
+
+@pytest.fixture
+def wcs_3d_ln_lt_t_coupled():
+    # WCS for a 3D data cube with two celestial axes and one time axis.
+    header = {
+        'CTYPE1': 'HPLN-TAN',
+        'CRPIX1': 5,
+        'CDELT1': 5,
+        'CUNIT1': 'arcsec',
+        'CRVAL1': 0.0,
+
+        'CTYPE2': 'HPLT-TAN',
+        'CRPIX2': 5,
+        'CDELT2': 5,
+        'CUNIT2': 'arcsec',
+        'CRVAL2': 0.0,
+
+        'CTYPE3': 'UTC',
+        'CRPIX3': 1.0,
+        'CDELT3': 1,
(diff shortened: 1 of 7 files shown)


Cadair:
This PR needs tests which test at least with some DKIST or IRIS WCSes that you can get low-memory axis world coords for uncorrelated axes, and as a counterpoint that correlated axes still work properly.

For example, I have tested that `axis_world_coords("time")` works fine for [this dataset](https://dkist.data.nso.edu/datasetview/ADDMM) but not that `axis_world_coords("lat")` works for instance (which should still generate a large grid).

ndcube/ndcube.py
```diff
@@ -573,6 +533,12 @@ def _generate_dependent_world_coords(self, pixel_corners, wcs, needed_axes, unit
             # And inject 0s for those coordinates
             for idx in non_corr_axes:
                 sub_range.insert(idx, 0)
+            # If we are subsetting world axes, ignore any pixel axes which are not correlated with our requested world axis.
+            if any(world_axis in needed_axes for world_axis in world_axes_indices):
+                unneeded_pixel_axes = wcs.axis_correlation_matrix[needed_axes]
+                unneeded_pixel_axes = np.argwhere(unneeded_pixel_axes.sum(axis=0) == 0)[:, 0]
```
Cadair, line 539:
```suggestion
                needed_pixel_axes = wcs.axis_correlation_matrix[needed_axes]
                unneeded_pixel_axes = np.argwhere(needed_pixel_axes.sum(axis=0) == 0)[:, 0]
```

nabobalis:
I do question why I bother to do any programming if its always wrong.

This solution works:

Command line: /home/nabil/.mamba/envs/iris-dev/bin/memray run -f -o output.bin examples/umbral_flashes.py
Start time: Tue Dec 17 2024 11:08:16 GMT-0800 (Pacific Standard Time)
End time: Tue Dec 17 2024 11:08:42 GMT-0800 (Pacific Standard Time)
Duration: 0:00:26.707000
Total number of allocations: 6674037
Total number of frames seen: 17921
Peak memory usage: 2.3 GiB
Python allocator: pymalloc

The biggest allocations are from loading the data:

Screenshot_20241217_111209

DanRyanIrish:
This is great news!

Cadair:
@nabobalis I can't add a visp WCS without a `dkist` dependency, I assume you can't either?

nabobalis:
> @nabobalis I can't add a visp WCS without a `dkist` dependency, I assume you can't either?

I am the same

nabobalis:
Could we try Wills WCS from his issue?

ndcube/conftest.py
```diff
@@ -519,6 +560,25 @@ def ndcube_3d_rotated(wcs_3d_ln_lt_t_rotated, simple_extra_coords_3d):
     return cube
 
 
+
+@pytest.fixture
+def ndcube_3d_coupled(wcs_3d_ln_lt_l_coupled):
```
nabobalis, line 565: Not sure if this is good.

ndcube/conftest.py
```diff
@@ -292,6 +292,47 @@ def wcs_3d_ln_lt_t_rotated():
     return WCS(header=h_rotated)
 
 
+@pytest.fixture
+def wcs_3d_ln_lt_l_coupled():
+    # WCS for a 3D data cube with two celestial axes and one wavelength axis.
+    # The latitudinal dimension is coupled to the third pixel dimension through
+    # a single off diagonal element in the PCij matrix
+    header = {
```
nabobalis, line 300: Taken from Wills original issue.

ndcube/tests/test_ndcube.py
```diff
@@ -230,6 +230,40 @@ def test_axis_world_coords_wave_ec(ndcube_3d_l_ln_lt_ectime):
     assert coords[0].shape == (5,)
 
 
+def test_axis_world_coords_wave_coupled_dims(ndcube_3d_coupled):
```
nabobalis, line 233: Copied another test, not the most useful for this issue.

ndcube/tests/test_ndcube.py
```diff
@@ -230,6 +230,13 @@ def test_axis_world_coords_wave_ec(ndcube_3d_l_ln_lt_ectime):
     assert coords[0].shape == (5,)
 
 
+@pytest.mark.limit_memory("12 MB")
```
Cadair, line 233: I grafted this test into the commit history before @nabobalis original #780 and it fails (using 3GB of RAM). 

This test doesn't actually fail before this PR though, as this change fixes the case where you want to extract a single world coordinate correlated with two pixel axes where those pixel axes are also correlated with other world axes, i.e. the VISP WCS.

nabobalis (approved):

Cadair:
I would be nice to get a test which tests the 2D usecase.

nabobalis:
How does one do that?

ndcube/tests/test_ndcube.py
Cadair: I have now added a test which fails like this DKIST WCS did, but without the need to actually load the DKIST data.

nabobalis (approved):

PR #802, opened 2025-01-07
Skip plotting line data if it's all NaNs

SolarDrew:
## PR Description

Closes DKISTDC/dkist#477

Over in the DKIST python tools I discovered that trying to line plot a slice of a dataset took many times longer if there was no actual data there. This skips the attempt to actually plot a line full of NaNs which should solve that and cause no other problems anywhere, probably.


SolarDrew:
This will break if the data aren't a dask array, so that's not ideal.

ndcube/visualization/mpl_plotter.py
```diff
@@ -145,7 +145,9 @@ def _plot_1D_cube(self, wcs, axes=None, axes_coordinates=None, axes_units=None,
             # We plot against pixel coordinates
             axes.errorbar(np.arange(len(ydata)), ydata, yerr=yerror, **kwargs)
         else:
-            axes.plot(ydata, **kwargs)
+            # compute() will break on numpy arrays to catch dask arrays by casting to bool instead
```
DanRyanIrish, line 148: This comment should make more sense to those who aren't dask wonks ;)

changelog/802.trivial.rst
```diff
@@ -0,0 +1 @@
+Skip plotting line data if all values are NaN
```
DanRyanIrish, line 1:
```suggestion
Make `~ndcube.visualization.mpl_plotter.MatplotlibPlotter` only add data to line plots if at least one value of the data is unmasked and finite.
```

ndcube/visualization/mpl_plotter.py
```diff
@@ -145,7 +145,9 @@ def _plot_1D_cube(self, wcs, axes=None, axes_coordinates=None, axes_units=None,
...
+            if not bool(np.isnan(ydata).all()):
```
DanRyanIrish, line 149: Could something like
```suggestion
            if np.isfinite(ydata).any() and unmasked:
```
be used instead without causing dask problems?  This way both NaN and Inf would be checked.  In addition, it would check whether there are valid unmasked data because  `unmasked` would be a `bool` that would indicate that at least one finite data value is unmasked and default to `True`.  It would be calculated around [here](https://github.com/sunpy/ndcube/pull/802/files#diff-87dd7a503ae11f67b56b44df5c6bbdb672ba9a78281c43669bc3763ece62200bR139).

DanRyanIrish (commented):
Does an equivalent check need to be done for plotting a 2D image?

DanRyanIrish:
Hi @SolarDrew. What's the status of this PR?  Would you still like to merge it?

ndcube/visualization/mpl_plotter.py
nabobalis: We will leave this up for the future

nabobalis (approved):

PR #806, opened 2025-01-13
Fix one test for new gwcs

Cadair:


nabobalis (approved):

Cadair:
I think the rest of the fails are upstream bugs, but still investigating.

Issue #828, opened 2025-03-10
NDCube.fill


DanRyanIrish:
### Describe the feature

A method that allows users to replace masked data values with a custom value, an alter associated uncertainty and mask values.

Possible API:

```python
def fill(fill_value, unmask=False, uncertainty_fill_value=None, fill_in_place=False):
    """
    Replaces masked data values with input value.

    Returns a new instance or alters values in place.

    Parameters
    ----------
    fill_value: `numbers.Number` or scalar `astropy.unit.Quantity`
        The value to replace masked data with.
    unmask: `bool`, optional
        If True, the newly filled masked values are unmasked. If False, they remain masked
        Default=False
    uncertainty_fill_value: `numbers.Number` or scalar `astropy.unit.Quantity`, optional
        The value to replace masked uncertainties with.
    fill_in_place: `bool`, optional
        If `True`, the masked values are filled in place.  If `False`, a new instance is returned
        with masked values filled.  Default=False.
    """
    # ...code implementation here.
```

nabobalis:
We could call it fillna

ayshih:
The method should not be named just `fill()`, because users will assume that the entire array will be filled.  The analogue in `numpy.ma` is `filled()`, but that's not a sufficiently distinctive name.  I suggest `fill_masked()`.

PR #829, opened 2025-03-14
NDCube.fill_masked() method

PCJY:
## PR Description
This pull request aims to resolve the issue created by @DanRyanIrish , https://github.com/sunpy/ndcube/issues/828 .


ndcube/ndcube.py
```diff
@@ -1330,6 +1330,50 @@ def squeeze(self, axis=None):
...
+    unmask: `bool`, optional
+        If True, the newly filled masked values are unmasked. If False, they remain masked
+        Default=False
+    uncertainty_fill_value: `numbers.Number` or scalar `astropy.unit.Quantity`, optional
+        The value to replace masked uncertainties with.
+    fill_in_place: `bool`, optional
+        If `True`, the masked values are filled in place.  If `False`, a new instance is returned
+        with masked values filled.  Default=False.
+    """
+    # ...code implementation here.
```
DanRyanIrish, line 1352:
```suggestion
```

ndcube/ndcube.py
```diff
@@ -1330,6 +1330,50 @@ def squeeze(self, axis=None):
...
+    kwargs = {}
+    kwargs["data"] = self.data
+    kwargs["mask"] = self.mask
+    kwargs["uncertainty"] = self.uncertainty
```
DanRyanIrish, line 1356: We only need to generate kwargs if we want to create a new instance, i.e. if `fill_in_place=False`.  So let's set the data, mask and uncertainty to variables that can be pointers to the exisiting cube's attributes, or deepcopies of them, depending on `fill_in_place`.  So something like:
```suggestion
    if fill_in_place:
        new_data = self.data
        new_uncertainty = self.uncertainty
        new_mask = False if unmask else self.mask
    else:
        new_data = copy.deepcopy(self.data)
        new_uncertainty = copy.deepcopy(self.uncertainty)
        new_mask = False if unmask else copy.deepcopy(self.mask)
```
Variable references below need to be changed accordingly

ndcube/ndcube.py
```diff
@@ -1330,6 +1330,50 @@ def squeeze(self, axis=None):
...
+
+
+    if (fill_in_place is False):
+        # If there is a not None mask and a not None fill_value, do: change the corresponding data to fill_value.
+        if (fill_value is not None and self.mask is not None):
```
DanRyanIrish, line 1361: Given the above assignment of data variables, we no longer need to check `fill_in_place`.

Also, we should support any fill value that's compatible with the data array.  If the user supplies a non-compatible one, the code will automatically error we you try to index the data (or uncertainty) array. So we don't need to check if `fill_value is None``.

However, we can combine the 2nd if condition to check whether there are any masked values.  If there aren't, there's no need to fill them. Then you can produce a new mask that can be used to index the data array (i.e. handle cases where mask is `bool`-type rather than an array).
```suggestion
    masked = False if (self.mask is None or self.mask is False or not self.mask.any()) else True
    if masked:
            idx_mask = slice(None) is self.mask is True else self.mask # Ensure indexing mask can index the data array.
```
Then proceed with filling masked values.

ndcube/ndcube.py
```diff
@@ -1330,6 +1330,50 @@ def squeeze(self, axis=None):
...
+            kwargs["data"][self.mask] = fill_value # Boolean indexing in Python.
+        # else, do nothing to the data.
+
+        # if unmask is True, do: change the True mask values to False, otherwise, do nothing to the mask.
+        if (unmask):
+            kwargs["mask"] = False
```
DanRyanIrish, line 1367: This is done in by a suggested change above. If you accept that, you can remove this from here.
```suggestion
```

ndcube/ndcube.py
```diff
@@ -1330,6 +1330,50 @@ def squeeze(self, axis=None):
...
+
+        if (self.mask is not None and uncertainty_fill_value is not None):
```
DanRyanIrish, line 1369: If you accept the above changes, we already know that `self.mask` is not None at this point.  But we do need to check that `self.uncertainty` is not None.
```suggestion
        if uncertainty_fill_value is not None:
            if not self.uncertainty:
                raise TypeError("Cannot fill uncertainty as uncertainty is None.")
```

ndcube/ndcube.py
```diff
@@ -1330,6 +1330,50 @@ def squeeze(self, axis=None):
...
+            kwargs["uncertainty"][self.mask] = uncertainty_fill_value
```
DanRyanIrish, line 1370:
```suggestion
            new_uncertainty.array[idx_mask] = uncertainty_fill_value
```

ndcube/ndcube.py
```diff
@@ -1330,6 +1330,50 @@ def squeeze(self, axis=None):
...
+
+    # if fill_in_place is True, then change self directly? without creating kwargs?
+
+    return kwargs
```
DanRyanIrish, line 1374:
```suggestion
    if not fill_in_place:
        # Create kwargs dictionary and return a new instance.
```

DanRyanIrish (changes requested):
Hi @PCJY.  Here are some suggested improvements.

Once you've handled these, you should also consider how to handle the case where the cube (including the uncertainty) has a unit.

Then this PR will also need tests.

ndcube/ndcube.py
```diff
@@ -1330,6 +1330,50 @@ def squeeze(self, axis=None):
...
+            kwargs["data"][self.mask] = fill_value # Boolean indexing in Python.
+        # else, do nothing to the data.
+
+        # if unmask is True, do: change the True mask values to False, otherwise, do nothing to the mask.
+        if (unmask):
+            kwargs["mask"] = False
+
+        if (self.mask is not None and uncertainty_fill_value is not None):
+            kwargs["uncertainty"][self.mask] = uncertainty_fill_value
+
```
DanRyanIrish, line 1371: Fill mask in-place here, after other operations complete, but before end of above `if masked` if statement.
```suggestion
        if fill_in_place and unmask:
            self.mask = False
```
May require more subtle handling if mask cannot be directly assigned.

ndcube/ndcube.py
PCJY: Hi @DanRyanIrish, should it be:  
```
if masked:
    idx_mask = slice(None) if self.mask is None or (isinstance(self.mask, np.ndarray) and not np.any(self.mask)) or self.mask is False else self.mask
```
 , instead? 
Meaning, if self.mask is None or all values are False or is a single bool F as a whole (every mask value is T), then there is no need to index the data array at all. Otherwise, use the mask array as the index array?

ndcube/ndcube.py
PCJY: Hi @DanRyanIrish , These two statements are not logically parallel with each other, why is it okay to write '' elif unmask '' ? Should I use  '' if unmask '' instead?

ndcube/ndcube.py
DanRyanIrish: I think your version is unnecessarily long.  Because this line of code is only executed if `masked is True`, then we already know that `self.mask` is not `None` or `False`. Therefore, the only possibilities are that `self.mask` is `True` or a boolean array.  If `self.mask` is `True`, the it can't be used to directly index the data array, therefore we need to set `idx_mask` to a slicing item that captures all elements (i.e. `slice(None)`).  Otherwise, `self.mask` can be used as is.  So I think (but please correct me if I'm wrong) that my suggested line is sufficient.

On a side note inspired by your suggestion of `isinstance(self.mask, np.ndarray)`, `NDCube` does not enforce that all arrays be numpy arrays.  Other arrays exist in Python, e.g. dask arrays.  Therefore, we should never rely on an assumption that arrays are a specific type, e.g. numpy, unless we deliberately and explicitly only want to handle a specific type of array.  And this should be avoided wherever possible.

ndcube/ndcube.py
DanRyanIrish: The `elif` statement is only evaluated if `fill_in_place` is `False`.  Therefore, it is equivalent to the following `if` statement: 
`if unmask and not fill_in_place`.
But having two separate `if` statements would mean we unnecessarily check the value of `fill_in_place` twice.  Therefore, writing it in the way I've suggested is neater and more efficient.

PCJY: I see! This makes sense.

DanRyanIrish:
@PCJY Let me know when this is ready for another review.  Also, be sure to pull the latest changes from the main branch into this branch.

PCJY:
Hi @DanRyanIrish I have just implemented the method more, finished adding tests for the method. Apologies for the delay. I also handled units by keeping the unit the same (which might not be what you were trying to suggest, because you might mean: the fill_value itself has a unit? ).

ndcube/ndcube.py
```diff
@@ -1330,53 +1331,62 @@ def squeeze(self, axis=None):
...
+            If `True`, the masked values are filled in place.  If `False`, a new instance is returned
+            with masked values filled.  Default=False.
+        """
+        # variable creations for later use.
+        # If fill_in_place is true, do: assign data and uncertainty to variables.
+        if fill_in_place:
+            new_data = self.data
+            new_uncertainty = self.uncertainty
+            new_mask = False if unmask else self.mask
+            new_unit = self.unit
```
DanRyanIrish, line 1359: The unit should stay the so doesn't need to be set here or in the `else` statement below.  Instead, I meant that, after this `if...else` statement, you should check whether `fill_value` has a unit and then convert it to the same unit as the cube, as then strip the unit, i.e.
```pythion
if hasattr(fill_value, "unit")`:
    fill_value = fill_value.to_value(self.unit)
```

You might need to do something equivalent for `uncertainty_fill_value`?

ndcube/ndcube.py
```diff
@@ -1330,6 +1331,63 @@ def squeeze(self, axis=None):
...
+                    raise TypeError("Cannot fill uncertainty as uncertainty is None.")
+                new_uncertainty.array[idx_mask] = uncertainty_fill_value
+
+        if not fill_in_place:
+            # Create kwargs dictionary and return a new instance.
+            kwargs = {}
+            kwargs['data'] = new_data
+            kwargs['uncertainty'] = new_uncertainty
+            kwargs['mask'] = new_mask
+            kwargs['unit'] = new_unit
```
DanRyanIrish, line 1386: In line with my comment above about the unit, this line should not be needed here.
```suggestion
```

ndcube/ndcube.py
```diff
@@ -1355,25 +1355,26 @@ def fill_masked(self, fill_value, unmask=False, uncertainty_fill_value=None, fil
...
-                if not self.uncertainty:
+            if hasattr(fill_value, "unit"):
+                fill_value = fill_value.to_value(self.unit)
+            new_data[idx_mask] = fill_value   # python will error based on whether data array can accept the passed value.
+
+            if hasattr(uncertainty_fill_value, "unit"):
+                uncertainty_fill_value = uncertainty_fill_value.to_value(self.unit)
+            if uncertainty_fill_value is not None: # Q: can it be None?? It must be numerical.
+                if not self.uncertainty:  # or new_uncertainty
                     raise TypeError("Cannot fill uncertainty as uncertainty is None.")
```
DanRyanIrish, line 1378:
```suggestion
            if uncertainty_fill_value is not None: # Q: can it be None?? It must be numerical.
                if not self.uncertainty:  # or new_uncertainty
                    raise TypeError("Cannot fill uncertainty as uncertainty is None.")
                if hasattr(uncertainty_fill_value, "unit"):
                    uncertainty_fill_value = uncertainty_fill_value.to_value(self.unit)
```

DanRyanIrish (changes requested):

PCJY:
Hi @DanRyanIrish , I have finished debugging the tests (with the latest four commits):
1, There were some errors in the fixtures I used, so I fixed them.

2, I forgot to specify the argument "check_uncertainty_values" specifying whether it is necessary to check the uncertainty, as True, in the assert_all_cubes method. So I fixed it.

3, The coverage file reminded me that I did not test the case when uncertainty_fill_value is not None but the NDCube object's uncertainty is None. So I added an individual test method to test this case, expecting it to raise a TypeError. 
I wish to ask you a question: Do you think this is the best way to add this test case? I added it by adding an individual method because I am not sure what to put in the expected_cube parametrized argument for this scenario if I were to use the existing two test method structure.

4, Then the coverage file complains about not entering the two scenarios when one and only one of test_input and expected_cube has an uncertainty attribute. But I think the correct behaviour is that it does not enter these two blocks. So I used 'pragma no cover' to let the coverage file know it does not need to be covered by the tests.

Please let me know if there's anything you'd like me to adjust, thank you!

ndcube/tests/helpers.py
```diff
@@ -122,15 +122,28 @@ def assert_metas_equal(test_input, expected_output):
                 assert test_input[key] == expected_output[key]
 
 
-def assert_cubes_equal(test_input, expected_cube, check_data=True):
+def assert_cubes_equal(test_input, expected_cube, check_data=True, check_uncertainty_values=False):
     assert isinstance(test_input, type(expected_cube))
     assert np.all(test_input.mask == expected_cube.mask)
```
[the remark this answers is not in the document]
PCJY: Hi @DanRyanIrish , I just checked and found that: the way I defined the mask-all-true NDCube fixture's mask is using an np all-one array (using np.ones). So that's why it is not complaining. Should I change the fixture to a single Boolean True?

PCJY:
Hi @DanRyanIrish , thanks for your review and bringing up the single-Boolean case issue. In my latest commit, there a few new changes addressing the single-Boolean-True mask value test case, I think the test is more robust now. 
Please let me know if you have any further feedback, thanks.

ndcube/tests/helpers.py
```diff
@@ -124,7 +124,10 @@ def assert_metas_equal(test_input, expected_output):
 
 def assert_cubes_equal(test_input, expected_cube, check_data=True, check_uncertainty_values=False):
     assert isinstance(test_input, type(expected_cube))
-    assert np.all(test_input.mask == expected_cube.mask)
+    if isinstance(test_input.mask, bool) and isinstance(expected_cube.mask, bool):
+        assert test_input.mask == expected_cube.mask
```
DanRyanIrish, line 128:
```suggestion
    if isinstance(test_input.mask, bool):
        if not isinstance(expected_cube.mask, bool):
            raise AssertionError("Masks not of same type.")
        assert test_input.mask == expected_cube.mask
```

Cadair:
I've encountered various astropy masked things recently, they use `filled()`: https://docs.astropy.org/en/stable/api/astropy.utils.masked.Masked.html maybe we want to consider that?

ayshih:
Both `astropy.utils.masked.Masked` and `numpy.ma.MaskedArray` name their analogous method `filled()`.  The distinction I'd draw is that both of those classes are expressly for masked use – the word is in their names! – so including "masked" in a method name would be superfluous.  In contrast, I don't feel like masking is at the forefront of all `NDCube` usage, so I think users need to be alerted that the method is relevant only for masked `NDCube` objects.

DanRyanIrish (approved):

PR #833, opened 2025-04-30
Remove physical-type equality check in reproject_to

CyclingNinja:
Removing the quantity check in reproject to that prevents transforming HPC to HGC

Partially this is a PR to have a discussion about this, but this is also proving a blocker elsewhere.


Cadair:
@DanRyanIrish Do you remember what this was trying to do? I don't fully understand why we would have such a restrictive check in place?

DanRyanIrish:
> @DanRyanIrish Do you remember what this was trying to do? I don't fully understand why we would have such a restrictive check in place?

I don't.  But if it's blocking HPC -> HGC, I'm open to removing the check if we can't see any adverse consequences.

CyclingNinja:
Having spoken to @astrofrog about the handling at astropy level, we are happy to go ahead with removing this, knowing that mismatched systems will throw an error

changelog/833.bugfix.rst
```diff
@@ -0,0 +1,6 @@
+Trivial/Internat Changes
+------------------------
+
+- Remove the check for consistent coordinate systems as behaviour controlled in astropy
+
+-
```
nabobalis, line 6:
```
Removed the check for compatible coordinate systems within ``reproject_to`` as it was redundant.
```

nabobalis, line 6:
```suggestion
Removed the check for compatible coordinate systems within ``reproject_to`` as it was redundant.
```

DanRyanIrish: Or even:
```suggestion
Removed the check for compatible coordinate systems within `~ndcube.NDCube.reproject_to` as it was redundant.
```

DanRyanIrish (dismissed):

PR #834, opened 2025-05-01
Avoid using deprecated `astropy.utils.isiterable()`

eerovaher:
## PR Description

[`astropy.utils.isiterable()`](https://docs.astropy.org/en/latest/api/astropy.utils.misc.isiterable.html) is being deprecated because it duplicates [`numpy.iterable()`](https://numpy.org/doc/1.23/reference/generated/numpy.iterable.html), which can be used as a drop-in replacement.


nabobalis (approved):

nabobalis:
Thanks @eerovaher

PR #840, opened 2025-05-13
Enable multiplying NDCube and NDData

PCJY:
## PR Description

Enable NDCube to be multiplied with wcs-less NDData.

### Adapting `NDCube.add` to support addition and multiplication.
- [x] Rename `NDCube.add` to `NDCube._operate_arithmetic` and add an `operation` arg.  `operation` should be a string of either `"add"` or `"multiply"`.
- [x] Create a new `NDCube.add` method that simply returns `self._operate_arithmetic` with `operation` equal to `'add'`.
- [x] Rerun tests to ensure everything still works as expected.
- [x] Include if statements in `NDCube._operate_arithmetic` to check whether `operation == 'add'`.  (The `'multiply'` case can be supported later.)  There are three places where if statements are needed:
  - Managing the unit
  - Performing addition of the data
  - Calculating the uncertainties
  - Don't forget that these may need to be done in the `NDData` and `Quantity` sections of the code.
- [x] Rerun tests to make sure they all still pass.
- [x] Add an `operation` arg to `NDCube._combine_uncertainty`, i.e. `def _combine_uncertainty(self, value, result_data, operation)`. Replace `np.add` with `operation` in the code of this method.  Make sure you add `np.add` where you call `self._combine_uncertainty` in `NDCube._operation_arithmetic`.
- [x] Rerun tests to make sure everything is still working as expected.

### Adding support for multiplication.
- [x] Add `elif` statements to `NDCube._operation_arithmetic` to support multiplication.  Remember, there are three scenarios where multiplication differs from addition:
  - Handling the unit
  - Calculating the new data (`+`-> `*`)
  - Calculating the new uncertainties (`np.add` -> `np.multiply`)
  - Again, remember changes might be needed in the `Quantity` section of the method, as well as the `NDData` section.
- [x] Make the `NDCube.multiply` method return `self._operate_arithmetic` with `operation` set to `'multiply'`.
- [x] Make `NDCube.__mul__` call `NDCube.multiply` just like `NDCube.__add__` calls `NDCube.add`.
- [ ] Write a simple test for `NDCube.__mul__`.
- [x] Add a changelog file.


ndcube/ndcube.py
```diff
@@ -963,7 +963,7 @@ def _new_instance(self, **kwargs):
     def __neg__(self):
         return self._new_instance(data=-self.data)
 
-    def add(self, value, handle_mask=np.logical_and):
+    def _arithmetic(self, value, op, handle_mask=np.logical_and):
```
DanRyanIrish, line 966: The args need to be in the order that you call them with below.  Also, do keep in-line with methods starting with verbs, let's rename this method.  So:
```suggestion
    def _operate_arithmetic(self, operation, value, handle_mask=np.logical_and):
```

ndcube/ndcube.py
```diff
@@ -1016,6 +1016,9 @@ def add(self, value, handle_mask=np.logical_and):
         # return the new NDCube instance
         return self._new_instance(**kwargs)
 
+    def add(self, value, handle_mask = np.logical_and):
```
DanRyanIrish, line 1019: This method needs a proper docstring.  For an example, see the `NDCubeBase.reproject_to` method's docstring.

DanRyanIrish, line 1019:
```suggestion
    def add(self, value, handle_mask=np.logical_and):
```

DanRyanIrish:
Hi @PCJY.  On reflection, I think it would be simpler if the `_operate_arithmetic` only handled the `NDData` case.  This is because there isn't a lot of overlap in the code with the `Quantity` and other cases.  So this part of the code can remain the `add()` method like this:
```python
def _arithmetic_operate_with_nddata(self, operation, value, handle_mask):
    # Do stuff

def add(self, value, handle_mask=np.logical_and):
    if isinstance(value, NDData):
        kwargs = self._arithmetic_operate_with_nddata("add", value, handle_mask)
    elif hasattr(value, 'unit'):
        # Put non-NDData code from self.add here...

    # return the new NDCube instance
    return self._new_instance(**kwargs)
```

ndcube/ndcube.py
PCJY: Do the _operate_arithmetic and the multiply method also need docstring?

DanRyanIrish:
Another idea for making this simpler, is to create a default `handle_mask` function.  See the code below.

```python
def _arithmetic_handle_mask(self, self_mask, value_mask):
    if self_mask is None and value_mask is None:
        new_mask = None
    elif self_mask is None:
        new_mask = value_mask
    elif value_mask is None:
        mask = self_mask
    else:
        new_mask = np.logical_and(self_mask, value_mask)
    return new_mask
```

Then in `NDCube._arithmetic_operate_with_nddata` we can have the default `handle_mask=None` and a line of code such as:
```python
if handle_mask is None:
    handle_mask = self._arithmetic_handle_mask
kwargs["mask"] = handle_mask(self.mask, value.mask)
```
This would mean that if the user provides a `handle_mask` function, it is always used.  Whereas at the moment, it is only sometimes used, which is confusing.

DanRyanIrish:
> Do you mean, _operate_arithmetic() method handles the NDData case, and then, the add/mul method handles the Quantity case?

Yes.  But add/mul calls also `_operate_arithmetic` if `value` is an `NDData` instance.  So from the user-perspective it's the same as now.  But how we organise the code is simpler and easier to maintain.

ndcube/ndcube.py
```diff
@@ -1016,6 +1016,9 @@ def add(self, value, handle_mask=np.logical_and):
         # return the new NDCube instance
         return self._new_instance(**kwargs)
 
+    def add(self, operation, value, handle_mask = np.logical_and):
+        return self._operate_arithmetic(operation, value, handle_mask)
```
DanRyanIrish, line 1020: We already know what the operation is here, i.e. `"add"`, so we don't need to expose this arg to the user.
```suggestion
    def add(self, value, handle_mask=np.logical_and):
        return self._operate_arithmetic("add", value, handle_mask)
```

Same below for `multiply`.

DanRyanIrish:
Apologies for the delay in getting these comments to you @PCJY.  I wrote them but realised I didn't actually post them to the PR!

DanRyanIrish:
Also, we need a changelog file this PR too.

PCJY:
Hi @DanRyanIrish , here is my summary for all the methods we have now for scaling and addition for NDCube and wcs-less NDData. Please let me know if you have any feedback on this, thanks.

1, The ndcube.fill\_masked() method 
   This method can be used by the users to specify the masked values, and to apply (or not) on the data and uncertainty, etc.

2, The default ndcube.\_arithmetic\_handle\_mask() method
   This method is used when handle\_mask is not provided by users, as the default behavior of handling masks.
   The method essentially combines the mask using logical OR when both objects have masks.

3, The ndcube.\_combine\_uncertainty() method

4, The ndcube.\_arithmetic\_operate\_with\_nddata() method
   Raises a TypeError if the wcs attribute of the 'value' object is not None.
   Performs addition or multiplication when the 'value' object is an NDData instance, handles masks, units, and uncertainty (using the \_combine\_uncertainty() method). 
   Set to private and should only get called by the add method instead of by users (although they can but not encouraged to).

5, The ndcube.add() method
   Performs addition for the scenarios:
   when the 'value' object is an NDData instance (wcs-less, has mask, unit and uncertainty), (calls the                                                                                             ndcube.\_arithmetic\_operate\_with\_nddata() method)
   when the 'value' object is a Quantity object (has unit and data).

6, The ndcube.\_\_add\_\_() method, the + operator

Previously it was necessary because the handle\_mask and operation\_ignore\_mask arguments must be set by users, in order to deal with masked elements before using the uncertainty propagation method.

Now that we have many methods to deal with masks, we assume when users come to arithmetic operations, they have already got the masked data. We would not specify the masks' impact on the data inside the arithmetic operation methods. 

Therefore, this method just calls the ndcube.add() method.

7, The ndcube.multiply() method
   Performs multiplication for the scenarios:
   when the 'value' object is an NDData instance (wcs-less, has mask, unit and uncertainty), (calls the ndcube.\_arithmetic\_operate\_with\_nddata() method)
   when the 'value' object is a Quantity object (has unit and data).

8, The ndcube.\_\_mul\_\_() method, the * operator
This method calls the ndcube.multiply() method. (for the same reasons for the ndcube.\_\_add\_\_() method)

DanRyanIrish:
Hi @PCJY.  Thanks for the summary of our progress.  I concur that the next steps are to change the `__add__` method so that it only calls the `add` method (as you've already done to `__mul__` and `multiply`), and then write the tests for the multiply method.

Keep up the good work!

DanRyanIrish:
@PCJY: Here is a suggestion to discuss that I have reflected on as the project has evolved.  The `add` and `multiply` methods do not affect how the data or uncertainties are calculated.  Instead, their justification is now solely to handle the case where `value` is an `NDData` object with a mask and the user does not want the new mask to be calculated via `np.logical_or`.  This is likely to be a small subset of the cases.  Because the `+`/`*` operators will usually be used, users are just as likely to be far more familiar with them than the `add`/`multiple` methods, and may choose to manually change the mask themselves after the operation. This only actually requires one extra line of code:
```python
new_cube = cube1 + nddata1
new_cube.mask = np.logical_and(cube1.mask, nddata1.mask)
```
Therefore, the maintenance, documentation, and publicisation of the `add` and `multiply` methods are a lot of work to save users one line of code in a small subset of cases, and many users may choose not to use them anyway.

Therefore, should we consider removing the `add` and `multiply` methods from the API?  All the code you've written can instead live under the `__add__` and `__mul__` methods.  This would reduce how much documentation you have to write, and also simplify the tutorials.  Furthermore, in that tutorial, we can include an example like above, where the user alters the mask after the operation.  What are your thoughts?

PCJY:
> > @PCJY: Here is a suggestion to discuss that I have reflected on as the project has evolved. The `add` and `multiply` methods do not affect how the data or uncertainties are calculated. Instead, their justification is now solely to handle the case where `value` is an `NDData` object with a mask and the user does not want the new mask to be calculated via `np.logical_or`. This is likely to be a small subset of the cases. Because the `+`/`*` operators will usually be used, users are just as likely to be far more familiar with them than the `add`/`multiple` methods, and may choose to manually change the mask themselves after the operation. This only actually requires one extra line of code:
> > ```python
> > new_cube = cube1 + nddata1
> > new_cube.mask = np.logical_and(cube1.mask, nddata1.mask)
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Therefore, the maintenance, documentation, and publicisation of the `add` and `multiply` methods are a lot of work to save users one line of code in a small subset of cases, and many users may choose not to use them anyway.
> > Therefore, should we consider removing the `add` and `multiply` methods from the API? All the code you've written can instead live under the `__add__` and `__mul__` methods. This would reduce how much documentation you have to write, and also simplify the tutorials. Furthermore, in that tutorial, we can include an example like above, where the user alters the mask after the operation. What are your thoughts?
> 
> Hi @DanRyanIrish , thanks for sharing your new ideas. I am thinking, then what about the code components in the add and multiply methods that handle the uncertainty and units? Without the add and multiply methods, how can the users combine these two attributes when it comes to the arithmetic operations for NDCube and wcs-less NDData? Please let me know your thoughts on this and what I might be missing, thank you.

Or is it because, all the unit handling and stuffs are handled by the _arithmetic_operate_with_nddata() method, and by changing the add and multiply methods to \_\_add\_\_ and \_\_mul\_\_ methods, it is just removing redundant wrappers but not contents?

PCJY:
> > Or is it because, all the unit handling and stuff is handled by the _arithmetic_operate_with_nddata() method, and by changing the add and multiply methods to `__add__` and `__mul__` methods, it is just removing redundant wrappers but not contents?
> 
> This is exactly right.
> 
> > But when adding NDCube with wcs-less NDData, if the users use the + and * operator, they would automatically use the _arithmetic_operate_with_nddata() method, without needing to add the one extra line new_cube.mask = np.logical_and(cube1.mask, nddata1.mask) ?
> 
> Yes, that's right. Using `+` or `*` operators to combine an `NDCube` with a WCS-less `NDData` will handle the masks according to your `_arithmetic_handle_mask` method. The extra line explicitly setting the mask is only needed if the user wants to over-ride this.

Then what should we do about the handle\_mask argument? The operator cannot have an argument, but the argument is present in the add method, which calls the \_arithmetic_operate_with_nddata, which calls the \_arithmetic_handle_mask method when handle\_mask is None.

DanRyanIrish:
> Then what should we do about the handle_mask argument? The operator cannot have an argument, but the argument is present in the add method, which calls the _arithmetic_operate_with_nddata, which calls the _arithmetic_handle_mask method when handle_mask is None.

We can simply delete the `handle_mask` kwarg.  Then the mask will always be handled by `_arithmetic_handle_mask`.  If the user wants something different, they can overwrite the mask after the operation as we discussed before, e.g.
```python
new_cube.mask = np.logical_and(cube1.mask, nddata1.mask)
```

ndcube/tests/test_ndcube.py
```diff
@@ -1266,7 +1266,7 @@ def test_arithmetic_add_cube_unc_nddata_unc(ndc, value):
     output_cube = ndc + value # perform the addition
     # Construct expected cube
     expected_data = ndc.data + value.data
-    expected_uncertainty = ndc.uncertainty.propagate(
+    expected_uncertainty = ndc.uncertainty.propagate(  # is this the correct way to test uncertainty? no need to test propagate
```
DanRyanIrish, line 1269: This is OK in this case because, as you say, we aren't testing `propagate`, and if we are using multiple parameterizations, we might not be able to hard-code the expected result as it may be different each time.  In this case, though, we only have one test case, so strictly speaking, it's better practice to explicitly set the expected value.  But I don't think it's important to change here.

ndcube/tests/test_ndcube.py
```diff
@@ -1453,6 +1453,56 @@ def test_cube_arithmetic_multiply(ndcube_2d_ln_lt_units, value):
...
+                            ),
+                            ("ndcube_2d_ln_lt_unit_unc_mask", NDData(np.ones((2, 3)),
+                                                                                wcs=None,
+                                                                                uncertainty=StdDevUncertainty(np.ones((2, 3))*0.1),
+                                                                                mask=np.ones((2, 3), dtype=bool),
+                                                                                unit=u.ct) # both of them have uncertainty and mask and unit.
+                            )
+                        ],
+                        indirect=("ndc",))
+def test_cube_arithmetic_multiply_ndcube_nddata(ndc, value):
```
DanRyanIrish, line 1475: It would probably be better to simplify this test by including the `expected_cube` in the parameterization.  Then the test only has to calculate the `output_cube` and compare it to the `expected_cube`.

PCJY:
Hi @DanRyanIrish , I think I have fixed the test for \_\_multiply\_\_. When you have a moment, could you please review it? I’d appreciate any feedback you might have, thank you.

ndcube/tests/test_ndcube.py
```diff
@@ -1473,6 +1453,35 @@ def test_cube_arithmetic_multiply(ndcube_2d_ln_lt_units, value):
     # TODO: test that uncertainties scale correctly
 
 
+@pytest.mark.parametrize(("ndc", "value", "expected_cube"),
+                        [
+                            ("ndcube_2d_ln_lt_no_unit_no_unc_no_mask_2", NDData(np.ones((2, 3)),
+                                                                                wcs=None),
+                             "ndcube_2d_ln_lt_res_ndc_mask_unc_unit" # neither of the two has uncertainty or mask or unit.
```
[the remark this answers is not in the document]
PCJY: But I think to initialise NDCube with wcs in it, it cannot know what 'wcs_2d_lt_ln' is within the parametrize() scope, if wcs_2d_lt_ln is only defined later in the test function.

DanRyanIrish (approved):

DanRyanIrish:
Congratulations and thank you, @PCJY , for this PR :partying_face:

PR #844, opened 2025-05-27
Tidy NDCube tests

DanRyanIrish:
## PR Description

This PR breaks test_ndcube.py into multiple files based on the functionalities being tested, as the file was becoming unwieldy.

DO NOT merge until after #840 

## To Do
- [ ] Move arithmetic tests to own module (not until #844 merged).
- [ ] Should slicing tests be their own module in the [mixin subpackage](https://github.com/sunpy/ndcube/tree/main/ndcube/mixins/tests)?


ndcube/tests/test_ndcube_arithmetic.py
```diff
@@ -0,0 +1,295 @@
...
+                            ),
+                        ],
+                        indirect=("ndc",))
+def test_arithmetic_add_one_unit(ndc, value):
+    assert isinstance(ndc, NDCube)
+    with pytest.raises(TypeError, match="Adding objects requires that both have a unit or neither has a unit."):
+        ndc + value
+
+
+@pytest.mark.parametrize(("ndc", "value", "expected_kwargs"),
```
DanRyanIrish, line 52: @PCJY Will you give this function and the `multiply` test below a quick review?  Just take a look and ask questions about anything you don't understand.  If you happen to find any problems, please point those out too.

In summary, I replaced all the `__add__` tests (except `test_arithmetic_add_one_unit`) with this one, which is similar to your `multiply` test below.  That's because the code and logic tree is much simpler than when you first wrote the tests.

I also added an extra case in the parameterizations of your multiply test to catch all the cases.

nabobalis:
I see no issue with splitting up test files to make it clear what each file is testing.

changelog/844.bugfix.rst
```diff
@@ -0,0 +1 @@
+Fix tests of resultant uncertainty when adding/multiplying `~ndcube.ndcube.NDCube` with `~astropy.nddata.NDData`.
```
PCJY, line 1: This is not fixing the resultant uncertainty but changing the fixture and results so that it covers all possible Boolean permutations for higher test coverage.

Cadair (approved):

PR #851, opened 2025-06-12
stole mailmap from sunpy

nabobalis:


Cadair (commented):
I think there are some contributors in here which aren't in core?

nabobalis:
I added those and checked the shortlog. I will triple check but I think I was able to clean it up fully.

Cadair (approved):

PR #869, opened 2025-07-01
Made error message more helpful when cropping to a single pixel

ayshih:
The error message when cropping to a single pixel does not currently make it clear that one can get the single pixel by setting `keepdims=True`.  This PR adds that helpful information.  This PR also fixes a typo in the utility function's docstring.


ndcube/utils/cube.py
```diff
@@ -127,7 +127,7 @@ def get_crop_item_from_points(points, wcs, crop_by_values, keepdims):
         Denotes whether cropping is done using high-level objects or "values",
         i.e. low-level objects.
 
-    keep_dims : `bool`
+    keepdims : `bool`
```
nabobalis, line 130: why is keepdims, not keep_dims?

nabobalis (approved):

Issue #871, opened 2025-07-03
Error when cropping a 1D NDCube


ayshih:
### Describe the bug

Attempting to crop a 1D NDcube errors.

### To Reproduce

```python
import numpy as np

import astropy.wcs
import astropy.units as u

from ndcube import NDCube

data_1d = np.arange(200)

wcs_1d = astropy.wcs.WCS(naxis=1)
wcs_1d.wcs.ctype = 'WAVE',
wcs_1d.wcs.cunit = 'nm',
wcs_1d.wcs.cdelt = 4,
wcs_1d.wcs.crpix = 1,
wcs_1d.wcs.crval = 3,

cube_1d = NDCube(data_1d, wcs=wcs_1d)
cube_1d.crop((7*u.nm,), (15*u.nm,))
```
gives
```
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
Cell In[78], line 1
----> 1 cube_1d.crop((7*u.nm,), (15*u.nm,))

File ~\AppData\Local\mambaforge\envs\py312\Lib\site-packages\ndcube\ndcube.py:622, in NDCubeBase.crop(self, wcs, keepdims, *points)
    619 def crop(self, *points, wcs=None, keepdims=False):
    620     # The docstring is defined in NDCubeABC
    621     # Calculate the array slice item corresponding to bounding box and return sliced cube.
--> 622     item = self._get_crop_item(*points, wcs=wcs, keepdims=keepdims)
    623     return self[item]

File ~\AppData\Local\mambaforge\envs\py312\Lib\site-packages\ndcube\utils\cube.py:57, in sanitize_wcs.<locals>.wcs_wrapper(*args, **kwargs)
     53     raise TypeError("wcs argument must be a High Level WCS or an ExtraCoords object.")
     55 params.arguments['wcs'] = wcs
---> 57 return func(*params.args, **params.kwargs)

File ~\AppData\Local\mambaforge\envs\py312\Lib\site-packages\ndcube\ndcube.py:648, in NDCubeBase._get_crop_item(self, wcs, keepdims, *points)
    644         if not (value is None or isinstance(value, classes[j])):
    645             raise TypeError(f"{type(value)} of component {j} in point {i} is "
    646                             f"incompatible with WCS component {comp[j]} "
    647                             f"{classes[j]}.")
--> 648 return utils.cube.get_crop_item_from_points(points, wcs, False, keepdims=keepdims)

[...]

PR #872, opened 2025-07-04
Fix bug in NDCube.crop

DanRyanIrish:
## PR Description
This PR fixed a bug outlined in #871 revealed by trying to crop a 1-D NDCube.

### DIFF
--- a/ndcube/utils/cube.py
+++ b/ndcube/utils/cube.py
@@ -141,17 +141,17 @@ def get_crop_item_from_points(points, wcs, crop_by_values, keepdims):
     # where each inner list gives the index of all points for that array axis.
     combined_points_array_idx = [[]] * wcs.pixel_n_dim
     high_level_wcs = HighLevelWCSWrapper(wcs) if isinstance(wcs, BaseLowLevelWCS) else wcs
-    wcs = high_level_wcs.low_level_wcs
+    low_level_wcs = high_level_wcs.low_level_wcs
     # For each point compute the corresponding array indices.
     for point in points:
         # Get the arrays axes associated with each element in point.
         if crop_by_values:
             point_inputs_array_axes = []
-            for i in range(wcs.world_n_dim):
+            for i in range(low_level_wcs.world_n_dim):
                 pix_axes = np.array(
-                    wcs_utils.world_axis_to_pixel_axes(i, wcs.axis_correlation_matrix))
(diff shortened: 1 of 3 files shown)


DanRyanIrish:
Hi @ayshih.  Here's a PR to fix the bug you found in #871.  Could you give it a review?

ndcube/utils/cube.py
```diff
@@ -164,14 +164,16 @@ def get_crop_item_from_points(points, wcs, crop_by_values, keepdims):
...
-        if len(array_axes_without_input):
+        if len(array_axes_without_input) > 0:
+            wcs_slice = np.array([slice(None)] * low_level_wcs.pixel_n_dim)
             wcs_slice[np.array(list(array_axes_without_input))] = 0
-        sliced_wcs = SlicedLowLevelWCS(wcs, slices=tuple(wcs_slice))
-        sliced_point = np.array(point, dtype=object)[np.array(point_indices_with_inputs)]
+            sliced_wcs = SlicedLowLevelWCS(low_level_wcs, slices=tuple(wcs_slice))
+            sliced_point = np.array(point, dtype=object)[np.array(point_indices_with_inputs)]
+        else:
+            sliced_wcs, sliced_point = low_level_wcs, np.array(point, dtype=object)
```
ayshih, line 176: I suggest adding a comment here that the variable `sliced_wcs` won't actually be a sliced WCS in this logic branch.  Just saving the potential confusion for a future maintainer.

ayshih (approved):
I confirm that this resolves #871, but I'll note that the successfully cropped cube will unavoidably have a WCS instance with a broken world->pixel conversion for versions of astropy without astropy/astropy#18394.

DanRyanIrish:
OK thanks. Do you think we should therefore set the lowest supported version of astropy for the next release of ndcube to the one with that fix?

ayshih:
> Do you think we should therefore set the lowest supported version of astropy for the next release of ndcube to the one with that fix?

No, I think that'd be too restrictive of a pin.  With the workaround in this PR, everything works in the ndcube API.  If a user encounters this bug with an astropy object, they simply need to update astropy.

ndcube/utils/cube.py
DanRyanIrish, line 176:
```suggestion
            # Else, if all axes have at least one crop input, no need to slice the WCS.
            sliced_wcs, sliced_point = low_level_wcs, np.array(point, dtype=object)
```

Issue #860, opened 2025-06-20
Crop mishandles negative array indices


DanRyanIrish:
### Describe the bug

`NDCube.crop` uses `wcs.world_to_array_index` to convert world coordinates to array indices. This function uses negative array indices to denotes points to the left/below the extent of the data array.  `NDCube.crop` naively uses the returned array indices to construct the required slice item.  This can result in a slice item of `slice(-1, 1)` which is intended to be equivalent to `slice(0, 1)`.  However, Python's slicing notation interprets this can slicing forwards from the last array index to the first. This results in slicing the cube to a shape of `(0,)`

## Proposed solution:
* `ndcube.utils.wcs.get_crop_item_from_points` should convert any negative array index to 0, so that it's output is always `slice(0, 1)`, and never `slice(-1, 1)`
* `NDCube.__getitem__` should add to its slice input sanitization so that it errors for cases like `slice(-1, 1)` where the result will be shape `(0,)`

Thanks to @hayesla for revealing this bug.

### To Reproduce

import sunpy

sunpy.map.Map(...)
etc

DanRyanIrish:
The `crop` related part of this issue is addressed by PR #874.  The `__getitem__` related part can be done in a separate PR.

Issue #858, opened 2025-06-19
Add some documentation about how crop is inclusive in terms of pixels


hayesla:
### Describe the feature

Following https://github.com/sunpy/sunpy/issues/8276, we should add some defining boundaries in pixel space on which ones are included/not. i.e. the fact that a point on the boundary, it includes that pixel in crop, rather than exclude it, like submap currently does.

DanRyanIrish:
## Current Rules for `NDCube.crop` (unless there's a bug)
* Array elements represent an interval in world space in each dimension that is closed on the lower end (includes the lower boundary) and open at the upper end (excludes the boundary).
  * This definition is in keeping with how Python indexing is defined.  (Note this is not the same how WCS defines integer pixel coordinates to be in the centre of the pixel.)
  * This definition ensures that any world coordinate only belongs within one array, and enables robust, unambiguous cropping behaviour to be implemented.
* `NDCube.crop` slices the cube to the smallest bounding box in array index space such that all locations in the cube defined by the input world coordinate points are included in the returned cube.  This makes  `NDCube.crop` a *closed-closed* interval operation, i.e. the lower and upper-most array elements are retained.
  * Note that this is different to slicing where all array elements from and including the lower array index and up to but not including the upper array index are kept, i.e. it is a *closed-open* interval operation.
  * The justifications for making `crop` a closed-closed operation are:
    * it is usually not possible to guarantee that the upper world coordinate will be the open boundary.  This is because array elements have a fixed non-zero width in all world dimensions.  If we tried to make `NDCube.crop` closed-closed, and the provided upper world coordinate was somewhere in the middle of an array element, then the actual upper boundary in world space would be an arbitrary distance below the provided world coordinate depending on where it lands in the pixel.  By contrast, the closed-open logic of slicing always returns a predictable shape and extent.
    * In the case where both the lower and upper boundaries fall within a single array element, e.g. the 1st, a closed-open logic would mean that the cube would have to both include and exclude that array element.  This could result in `NDCube` trying to crop itself to a scalar, which is not allowed.
    * If the world coordinate was rounded to the nearest array element, we would end up with a behaviour whereby sometimes the provided world coordinate would be included in the cropped cube, and other times not.  And it may be hard to predict which way it will go ahead of time with inspected `NDCube.wcs.world_to_array_index` before the operation.
  * Hence guaranteeing that all locations provided by the user will be in the cropped cube, is a more intuitive and predictable behaviour for users.

[...]

PR #874, opened 2025-07-08
Fixed cropping when a world point is exactly on a pixel edge

ayshih:
This PR fixes NDCube cropping so that it no longer treats the lower edges of a pixel as belonging to that pixel.  A world point exactly on a pixel edge does not immediately imply that any pixel touching that edge is included in the crop; it depends on the other world points specified for the crop.

TODO:
- [x] Update documentation
- [x] Add tests

This PR now also fixes the `crop` related part of the bug outlined by #860

### DIFF
--- a/ndcube/utils/cube.py
+++ b/ndcube/utils/cube.py
@@ -234,10 +234,15 @@ def get_crop_item_from_points(points, wcs, crop_by_values, keepdims, original_sh
             if min_array_idx == max_array_idx:
                 ambiguous = True
                 max_array_idx += 1
-                message += (f"All input points corresponding to array axis {array_axis} lie on "
-                            f"the boundary between array elements {min_array_idx} and "
-                            f"{max_array_idx}. The cropped NDCube will only include array "
-                            f"element {max_array_idx}.\n")
+                if min_array_idx == 0:
+                    message += (f"All input points corresponding to array axis {array_axis} lie on "
+                                "the lower boundary of array element 0 (the first element). "
+                                "The cropped NDCube will only include array element 0.\n")
+                else:
+                    message += (f"All input points corresponding to array axis {array_axis} lie on "
+                                f"the boundary between array elements {min_array_idx - 1} and "
+                                f"{min_array_idx}. The cropped NDCube will only include array "
+                                f"element {min_array_idx}.\n")
             if max_array_idx - min_array_idx == 1 and not keepdims:
                 item.append(min_array_idx)
             else:


ayshih:
I'm fixing the behavior only when the world point is exactly on a pixel edge.  None of your examples have world points that are exactly on a pixel edge (a multiple of 10).

> * `cube.crop(16A, 54A)` goes across [20, 50)A, (slicing equivalent 1:5), so neither input given to crop is in the cropped cube

No, the crop should be [10, 60)A

> * `cube.crop(17A, 23A)` (slicing equivalent 1:1), is not allowed and raises an error.

No, the crop should be [10, 30)A

ayshih:
> > A world point exactly on a pixel edge does not immediately imply that any pixel touching that edge is included in the crop; it depends on the other world points specified for the crop.
> 
> Does this mean that the intended behaviour is that crop is undefined if all points are in pixel corners? (Based on the code changes it doesn't look like that's what you're doing, but is it what you intend?)

There is nothing special about pixel corners.  The question is what pixels (not counting the pixel edges) are spanned between the specified world points.  If there are pixels between the specified pixel corners, then it is a valid crop.

ayshih:
This PR does result in the following behavior, which some may find objectionable:
* `cube.crop(15A, 15A)` returns [10, 20)A, because 15A is undeniably in that pixel
* `cube.crop(20A, 20A)` errors, because 20A is not defined to be in any pixel

DanRyanIrish:
> In your notation, we agree that:
> 
> * `cube.crop(11A, 19A)` returns [10, 20)A
> 
> * `cube.crop(11A, 21A)` returns [10, 30)A
> 
> 
> 
> Before this PR:
> 
> * `cube.crop(11A, 20A)` returns [10, 30)A
> 
> 
> 
> After this PR:
> 
> * `cube.crop(11A, 20A)` returns [10, 20)A

I need to think about this a bit more, but now I feel I better understand your vision and don't have the same level of concern as before.

DanRyanIrish:
> This PR does result in the following behavior, which some may find objectionable:
> * `cube.crop(20A, 20A)` errors, because 20A is not defined to be in any pixel

That said, this does seem to be an unreasonable limitation of the definitions of the algorithm. If we were to move to this scheme, I think the most intuitive thing to do would be to special case this example so it returns a length-1 cube spanning the interval [20, 30)A.

There may be other issues with that though that I haven't thought through enough though.

ayshih:
I'll explicitly note that this approach for handling pixel edges when converting a range of world coordinates to a range of pixels is precisely how we do it in `sunpy` core for submapping and also for auto-extent reprojection.  I don't see any reason why NDCube should behave differently.

> > * `cube.crop(20A, 20A)` errors, because 20A is not defined to be in any pixel
> 
> That said, this does seem to be an unreasonable limitation of the definitions of the algorithm. If we were to move to this scheme, I think the most intuitive thing to do would be to special case this example so it returns a length-1 cube spanning the interval [20, 30)A.

I think returning an error is the Right Thing™ to do.  Heck, I'm tempted to return an error for `cube.crop(15A, 15A)` because there's no space between the world coordinates, but at least there the pixel choice is unambiguous.

I do not think it should be special-cased.  Consider a mirror image of the cube and issuing the same crop call.  The same special-case logic would presumably return what is now the [20, 10)A interval, and thus the same crop call retrieves different pixels.

DanRyanIrish:
> Heck, I'm tempted to return an error for `cube.crop(15A, 15A)` because there's no space between the world coordinates, but at least there the pixel choice is unambiguous.

Recall that current mental map for `NDCube.crop` is to return the smallest cube in array-space that includes the input points. It is not about giving `crop` ranges in world coordinates. This mental map was decided upon because we don't know how many world axes correspond to any given array axis, nor whether any of those axes are rotated relative to the pixel grid. Given that, the current `NDCube.crop` mental map is the only one that is simple and intuitive all possible scenarios, even if it's less intuitive in specific simple scenarios.

There are 2 separable issues raised by this PR:
1. inputs considered points or intervals;
2. whether pixels are considered closed-open or closed-closed intervals in world-space.

Regarding the first, as stated above, I think we should keep the input points API as it's unambiguous in all possible cubes and WCSs.

However, I am open to considering switching to a closed-closed pixel model, if it doesn't make currently valid `NDCube.crop` inputs invalid. This brings us to the error cases.


> I do not think it should be special-cased.  Consider a mirror image of the cube and issuing the same crop call.  The same special-case logic would presumably return what is now the [20, 10)A interval, and thus the same crop call retrieves different pixels.

Agreed. If remember correctly, this was raised when we were trying to make `crop` work originally. The decision to consider array elements closed-open intervals in world space resolved this and other issues. The ambiguity you raise is a consequence of moving to a closed-closed interval model. That said, I don't think the specific case you mention is a problem. I think it's reasonable for the same crop call applied to a cube with a different WCS to return a different result.

More generally regarding the error, I think it's unreasonable for users to potentially have to think about the world values at specific pixel boundaries. After all, if they already know that, it's easier for them to use the array-slicing API. Therefore so long as there's at least one input point given within the FOV of the `NDCube`, it should return a valid cube. This would require a solution like the special-case I mentioned. If we can't support this or an equivalent solution, I would be against changing `NDCube.crop` from what we currently have.

DanRyanIrish:
> (I'll point out the above abstract examples use angstroms, but if you actually tried to run the examples in real code, it may be impossible to specify a world point that is exactly on a pixel edge, due to floating-point precision.  That's why my example [here](https://github.com/sunpy/ndcube/issues/858#issuecomment-3049064267) uses meters.)

I think this concern could be addressed with an `edge_tolerance` kwarg whereby if the resulting pixel coord is within, e.g. 1e-5, of the pixel edge, the input is considered to correspond to the edge.

ayshih:
> Recall that current mental map for `NDCube.crop` is to return the smallest cube in array-space that includes the input points.

Okay, I'll adjust my phrasing.  In your parlance, I want pixels to be considered "closed-closed" intervals for `crop()`.  In other words, a world point on a pixel edge should not be treated as belonging to a *specific* pixel, but rather should be treated as contained by *every* pixel that shares that pixel edge.

For example, 20A is contained in both the [10, 20]A pixel and the [20, 30]A pixel, and the following are the minimal crops:
* `cube.crop(20A, 29A)` should return [20, 30]A
* `cube.crop(11A, 20A)` should return [10, 20]A (rather than [10, 30]A)

> I think it's reasonable for the same crop call applied to a cube with a different WCS to return a different result.

While an arbitrarily different WCS is allowed to return different results, it's by considering mirror flips and rotations that we can evaluate biases in the output.  Any biases could lead to unexpected results and user confusion.  One should not choose biased output when there is an alternative – here, a tighter crop – that is stable to mirror flips and rotations.

For example:
* `cube.crop(20A, 30A)` should return [20, 30]A (rather than [20, 40]A)
* `cube_reflected.crop(20A, 30A)` should return [30, 20]A (rather than [30, 10]A)

> Therefore so long as there's at least one input point given within the FOV of the `NDCube`, it should return a valid cube. This would require a solution like the special-case I mentioned.

I think the crux here is that I feel that it is okay for there to be ambiguity given certain sets of world points – e.g., there are two crops of the same size that contain all of the world points – and in those cases, it is okay for `crop()` to error.  I'd prefer `crop()` to error when it's unclear what the user is asking for, rather than insisting on a certain (biased) solution that may be "wrong" as far as the user is concerned.

For example, I think `cube.crop(20A, 20A)` should error because both [10, 20]A and [20, 30]A are valid solutions, so the minimal crop is not well defined.  Upon seeing the error message, the user can adjust the world points to assert their preference, e.g., either `cube.crop(19A, 20A)` or `cube.crop(20A, 21A)`.

> > (I'll point out the above abstract examples use angstroms, but if you actually tried to run the examples in real code, it may be impossible to specify a world point that is exactly on a pixel edge, due to floating-point precision.  That's why my example [here](https://github.com/sunpy/ndcube/issues/858#issuecomment-3049064267) uses meters.)
> 
> I think this concern could be addressed with an `edge_tolerance` kwarg whereby if the resulting pixel coord is within, e.g. 1e-5, of the pixel edge, the input is considered to correspond to the edge.

Yes, of course.  I just wanted people to be aware that the above thought examples may not actually behave as advertised under this PR if using angstroms with wcslib.

DanRyanIrish:
I see the sense in what your saying and am open to your perspective.  However,...

> ...and in those cases, it is okay for `crop()` to error. I'd prefer `crop()` to error when it's unclear what the user is asking for,...

This is a major stumbling block for me.  I can imagine a case where a user might have a workflow where they are pulling the world coordinates of various features, e.g. contour boundaries, from some database, or from the output of a previous part of their pipeline, and want to crop a cube to that feature.  In the case of a single feature study, I agree that the user can manually adjust one of their coordinates, but I feel this may unnecessarily annoy the user and/or lead them to think that `ndcube` is broken.  In the case of a pipeline, this could become prohibitively annoying or cumbersome.

Ultimately, I see this limitation as making `NDCube.crop` less user-friendly and less robust.

> rather than insisting on a certain (biased) solution that may be "wrong" as far as the user is concerned.

I'd draw a distinction here between "wrong" and "not what a user wants/finds intuitive".  Wrong implies a bug or not self-consistent.  But so long as a behaviour is well-defined and bug-free, I would argue it cannot be "wrong".  But it may not be the same as some users' preference.

I think the same arguments you're making could be applied to the Python slicing API.  Below my understanding of how Python indexing works:  (Note that I am not referring to WCS pixel indexing where integer pixel indices refer to the centre of the pixel.)
In Python indexing, integer indices refer to the lower edge of the array element (see below)
```
| _ | _ | _ | _ |   # This line represents the array elements as _ and the edges of the elements as |
0   1   2   3   4   # This line represents the indices of the array elements. Note they are aligned with the edges of the elements.
```
And array elements are considered closed-open intervals, i.e. each index belongs only to the array element right-ward of it.  Hence, Python slicing also works as an open-closed interval, i.e.
```python
my_array[1:3]
| _ | _ |  # Note the output is length 2.
1   2   3
```
goes from the lower edge of the 1st (0-based counting) element to the lower edge of the 3rd element, i.e. it does not include the 3rd element.  Some may argue that this is "wrong", as it some users will "obviously" want to include the 3rd element if they enter `3` in the slicing.  (This was something I had to adjust to when I started using Python.)  However, I don't consider this behaviour "wrong" as it's well defined and unambiguous on its own terms.  It's simply a valid, if not unique, choice of definitions.

This means it solves various problems compared a closed-closed framework where array indices belong to both array elements it touches.  For example, the above closed-open paradigm the following is allowed and unambiguous:
```python
my_array[1]
_  # The contents of the 1st element, extracted from the array.
```
as it slices out the 1st array element.  However, in your closed-closed paradigm, this would be undefined as it could refer to the 0th or 1st element.  This is one reason why I feel that
```python
cube.crop(20A)
```
should be a valid call and return the pixel right-ward of 20A.

I agree with you that handling world values on pixel edges is the crux of the issue.  And unfortunately, for the above reasons, I think the current behaviour is better than your proposed behaviour because it is unambiguous and much closer to how Python slicing actually works.

That said, I am still open to adjusting the `NDCube.crop` behaviour so that if a world value that represents the upper/rightmost extent of the region of interest falls exactly on a pixel edge, that we do not include the upper/right-ward pixel associated with that world value.  I think this represents a compromise between the current and your proposed behaviour and makes `NDCube.crop` actually behave slightly more like Python indexing than it already does.

ayshih:
> * **Compromise solution**: Keep current paradigm with the following change: exclude rightmost pixel if uppermost world value lands on a pixel edge.  If all world values land on the same pixel edge, this is special cased so that the right-ward pixel is returned.

On that last point, how about instead of raising an error, `crop()` emits a warning about the ambiguity?  That way, `crop()` successfully returns a result that will probably be adequate for the majority of users, and the minority of users will be notified that they may want to adjust their world points.

DanRyanIrish:
I think a warning in the case that all world coordinates corresponding to a single pixel edge for a given array axis/axes, is a good idea.

ayshih:
I've revised this PR to emit a warning instead of raise an error for the edge case (ha ha)

DanRyanIrish:
Hi @ayshih.  Thanks for updating the PR.  At first glance, the changes you made look correct.  That said, I found it hard to decipher what each variable was because some of the variables now have pixel indices stored in array order.  I've made a PR to your branch with changes that make the `get_crop_item_from_points` function work explicitly in pixel space (including pixel axis ordering) until it comes time to calculate the min and max array indices for each array axis.  This also allows a slight simplification of some of the code higher up in the function.

While I was at it, I also shifted where the warning is raised so that a separate one is issued for each axis.  This allows more explicit info in the warning that I think will help users.  Technically, this means it's possible that multiple warnings can be raised for one crop operation.  However, this should be almost vanishingly small given the chances that any, let alone all, world values corresponds exactly to a pixel edge, and the still smaller chance that this happens for multiple axes.

However, if we don't want to risk even this chance, we could compile the warning message axis by axis as I've done, but issue the warning after iterating through all the axes using the `ambiguous` variable, as you've done.

DanRyanIrish:
On second glance, you were raising a warning whenever the min and max pixel coord was in the same pixel, not just the same pixel edge.  I've fixed with in the PR to your branch.

ayshih:
> While I was at it, I also shifted where the warning is raised so that a separate one is issued for each axis. This allows more explicit info in the warning that I think will help users. Technically, this means it's possible that multiple warnings can be raised for one crop operation.

I implemented it as a single warning covering all axes because with default warning settings, I believe only the first warning will get emitted.

> On second glance, you were raising a warning whenever the min and max pixel coord was in the same pixel, not just the same pixel edge.

Nope.  If the world points are in the pixel not at the edges, max will be min +1 because of how they are calculated.

DanRyanIrish:
> I implemented it as a single warning covering all axes because with default warning settings, I believe only the first warning will get emitted.

Ah I see.  In that case we should build up the warning message axis by axis to give the user the clearest information and then raise it as one warning like you did originally.

> Nope. If the world points are in the same pixel not at the edges, max will be min +1 because of how they are calculated.

I stand corrected :) I'll revert that change in my PR.

ndcube/utils/cube.py
```diff
@@ -137,81 +138,98 @@ def get_crop_item_from_points(points, wcs, crop_by_values, keepdims):
...
                 result_is_scalar = False
+    # Raise warning if all world values for any array axes correspond to a pixel edge.
+    if ambiguous:
+        warn_user(message)
     # If item will result in a scalar cube, raise an error as this is not currently supported.
     if result_is_scalar:
         raise ValueError("Input points causes cube to be cropped to a single pixel. "
-                         "This is not supported when keepdims=False.")
+                         "This is not supported when keepdims=False. "
+                         "Please set keepdims=True and try again.")
```
ayshih, line 232: I feel that adding this statement is a bit silly.  I feel that it is already implied by the preceding sentence, and it implicitly assumes that the user intended to crop to a single pixel, when it may actually be in error (e.g., specifying only a single world point)

DanRyanIrish:
Hi @ayshih.  I've made some updates to your PR.  Would you care to review them?

ayshih:
I've fixed a couple of bugs with the code

DanRyanIrish:
Thanks!

PR #884, opened 2025-10-02
Added specific cropping unit tests

nabobalis:


DanRyanIrish:
Hi @nabobalis.  It looks to me that the time world axis is also coupled to the celestial world axes via a common pixel axis.  So in order to unambiguously determine a point in the cube associated with a time, you also have to provide a `SkyCoord`, which you haven't done.  That's why the error message is saying that it expects 2 inputs (`(SkyCoord, Time)`) when you have only given it one (`Time`).

nabobalis:
With that change, I get:
`test_crop_by_extra_coords_using_combined_wcs - ValueError: 3 components in point 0 do not match WCS with 4 components.`

PR #887, opened 2025-10-06
Add NDCube method to convert to NDData instance.

DanRyanIrish:
## PR Description

This PR introduces a new method `NDCube.to_nddata`, which allows for an `NDCube` to be converted to an `NDData` (or subclass of same) instance.  The motivating use case is to enable users to easily drop coordinate-awareness from their data and thereby perform arithmetic operations with NDCubes.

### DIFF
--- a/ndcube/ndcube.py
+++ b/ndcube/ndcube.py
@@ -1540,27 +1540,23 @@ class NDCube(NDCubeBase):
                        "psf": psf,
                        "extra_coords": extra_coords,
                        "global_coords": global_coords}
-        extra_coords = self._extra_coords if user_kwargs["extra_coords"] is COPY else None
-        global_coords = self._global_coords if user_kwargs["global_coords"] is COPY else None
         user_kwargs = {key: value for key, value in user_kwargs.items() if value is not COPY}
         user_kwargs.update(kwargs)
         all_kwargs = {key.strip("_"): value for key, value in self.__dict__.items()}
         all_kwargs.update(user_kwargs)
         # Inspect call signature of new_nddata class and
         # remove unsupported items from new_kwargs.
-        nddata_sig = inspect.signature(nddata_type).parameters.keys()
-        all_kwarg_keys = list(all_kwargs.keys())
-        for key in all_kwarg_keys:
-            if key not in nddata_sig:
-                del all_kwargs[key]
+        all_kwargs = {key: value for key, value in all_kwargs.items()
+                      if key in inspect.signature(nddata_type).parameters.keys()}
         # Construct and return new instance.
         new_nddata = nddata_type(**all_kwargs)
         if isinstance(new_nddata, NDCubeBase):
-            if extra_coords:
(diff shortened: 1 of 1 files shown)


DanRyanIrish:
@Cadair: I found time to PR this.  Whatever we agree on the call signature, I think this captures the core of the implementation.

ndcube/ndcube.py
```diff
@@ -1461,6 +1461,59 @@ def fill_masked(self, fill_value, uncertainty_fill_value=None, unmask=False, fil
...
+
+        Parameters
+        ----------
+        nddata_type:
+            The type of the returned object. Must be a subclass of `~astropy.nddata.NDData`
+            or a class that behaves like one.  Default=`~astropy.nddata.NDData`.
+
+        kwargs:
+            Attributes to change on output object. For example, to change the data on the
+            returned object compare to this instance, set a kwarg ``data=new_data``.
```
DanRyanIrish, line 1480:
```suggestion
            returned object compared to this instance, set a kwarg ``data=new_data``.
```

ndcube/ndcube.py
```diff
@@ -1461,6 +1461,59 @@ def fill_masked(self, fill_value, uncertainty_fill_value=None, unmask=False, fil
...
+            Note that kwargs set to ``None`` will not be passed to the constructor
```
DanRyanIrish, line 1481:
```suggestion
            Inputs to the return type's constructor that should differ from the values
            stored in attributes of this instance. For example, to change the data on the
            returned object compared to this instance, set a kwarg ``data=new_data``.
            Note that kwargs set to ``None`` will not be passed to the constructor
```

ndcube/ndcube.py
```diff
@@ -1461,6 +1461,59 @@ def fill_masked(self, fill_value, uncertainty_fill_value=None, unmask=False, fil
...
+            of the ``nddata_type``. Therefore, if that attribute exists on the new type,
+            it will be given the default value defined by ``nddata_type``'s call signature.
```
DanRyanIrish, line 1483:
```suggestion
            it will be given the default value defined by ``nddata_type``'s call signature.
            Note that kwargs given by the user and attributes on this instance that are
            not supported by the ``nddata_type`` constructor are ignored.
```

DanRyanIrish, line 1483:
```suggestion
            Inputs to the ``nddata_type`` constructor that should differ from the values
            stored in attributes of this instance. For example, to set different data values
            on the returned object, set a kwarg ``data=new_data``, where ``new_data`` is an
            an array of compatible shape and dtype.  Note that kwargs set to ``None`` will
            not be passed to the constructor of the ``nddata_type``. Therefore, if that
            attribute exists on the new type, it will be given the default value defined by
            ``nddata_type``'s call signature.  Note that kwargs given by the user and
            attributes on this instance that are not supported by the ``nddata_type``
            constructor are ignored.
```

PR #892, opened 2025-10-09
Simplify NDCube.to_nddata

Cadair:
I couldn't take it.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -28,7 +28,7 @@ jobs:
       coverage: codecov
       toxdeps: tox-pypi-filter
       envs: |
-        - linux: py313
+        - linux: py313-minimal
     secrets:
       CODECOV_TOKEN: ${{ secrets.CODECOV_TOKEN }}
 
--- a/ndcube/conftest.py
+++ b/ndcube/conftest.py
@@ -4,7 +4,6 @@ predictable NDCube objects.
 """
 import logging
 
-import dask.array
 import numpy as np
 import pytest
 from gwcs import coordinate_frames as cf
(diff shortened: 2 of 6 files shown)


Cadair:
> I know you're joking, but...no.

I wasn't, but I've changed it to "copy".

> we are trying to use the call signature as documentation. That's what the docstring is for.

I strongly and fundamentally disagree with this. The function signature is the most important source of documentation, it is machine and human readable, it's the thing you actually write in your code.

---

My main issue with your suggestion is the automatic extraction of attributes from the instance `__dict__`, I think that's prone to failure and generally makes me feel uncomfortable. 

I think this PR makes things clearer for the two main use cases:

* Making a `NDData` object: things like tab-completion and editor hints will work because the function signature is populated.
* Making a new instance of a user-specified different type: There's no magic here, if the user wants to copy a new attribute they can specify it.

I sat down and tried to write a test which on a subclass of NDCube this method gets overwritten and made to make a new type. I realised in the process of doing this that it's then a different method, it would be a `to_ndcube` or `to_myclass` method, which could then use this underneath, but it's not a `to_nddata` method at that point. So I'm not worried about not easily addressing that use case.

ndcube/ndcube.py
```diff
@@ -1465,24 +1465,28 @@ def fill_masked(self, fill_value, uncertainty_fill_value=None, unmask=False, fil
...
-        Attribute values can be altered on the output object by setting a kwarg with the new
-        value, e.g. ``data=new_data``.
-        Any attributes not supported by the new class (``nddata_type``), will be discarded.
+        Constructs a new `~astropy.nddata.NDData` instance from this object.
+
+        By default all known ``NDData`` attributes are copied (by reference) from
+        this object, values can be altered on the output object by
+        setting a kwarg with the new value, e.g. ``data=new_data``.
+        Custom attributes on this class can be passed by setting that
+        keyword to `"copy"`, for example ``mycube.to_nddata(spam="copy")``
```
```suggestion
        keyword to ``"copy"``, for example ``mycube.to_nddata(spam="copy")``
```

PR #893, opened 2025-10-30
docs: add example using reproject to change wcs resolution

alasdairwilson:
Thoughts on this very simple upscaling example?

I also added a new piece of error handling to rebin as the current message when you use scale factors > 1 is misleading 

(e.g. 2 is an integer factor)


alasdairwilson:
Thanks for the reminder that this existed stale-bot

I think I have tackled the core issue, the language is now much more precise that we are changing resolution and not upscaling and ive added a warning to that sort and pointing users to rebin for integer changes.

Cadair (approved):

PR #931, opened 2026-05-20
Added unit tests that mimics a dynamic spectra

nabobalis:
Comes with a bugfix(?).

SPOILER: AI WAS USED but code was human reviewed and edited.


ndcube/utils/cube.py
```diff
@@ -181,8 +181,20 @@ def get_crop_item_from_points(points, wcs, crop_by_values, keepdims, original_sh
         # Derive the pixel indices of the input point and place each index
         # in the list corresponding to its axis.
         # Use the to_pixel methods to preserve fractional indices for future rounding.
-        point_pixel_indices = (sliced_wcs.world_to_pixel_values(*sliced_point) if crop_by_values
-                               else HighLevelWCSWrapper(sliced_wcs).world_to_pixel(*sliced_point))
+        if crop_by_values:
```
nabobalis, line 184: I am not sure if this is quite true but it did break a unit test.

nabobalis: This is a real regression in the oldest deps, gwcs looks like it used to handle this internally?!

Cadair: By bumping gwcs to 0.24 it has fixed this issue. That version is ~16 months old at this point so I'm happy with that.

Cadair (approved):

PR #942, opened 2026-06-11
Drop Python 3.11 and updgrade dep

nabobalis:
Also dropped 3.11. 

Updated all packages using spec0-action with a 2 year cadence. 

Maybe you don't want to do this but I think for 2.5.0, I think it's time.

TODO:
- [ ] Changelog?


Cadair:
Thanks. I think I'll sit on this until we are closer to 2.5 to make backports a bit easier.

We are keeping numpy at 1.26 (over all the sunpy packages) until it causes us trouble just for people who might still be stuck.

PR #950, opened 2026-07-23
Typing NDCube

samaloney:
## PR Description

Add types and type checking to NDCube

* Some of the types could probably use astropy's `QuantityLike`
* Some of the types weren't clear so had to guess
* Found some real or at least potential bugs (commit e1808f41809899cd04fac01fe685f44e27092191)

## AI Assistance Disclosure

AI tools were used for:
- [x ] Code generation (e.g., when writing an implementation or fixing a bug)
- [  ] Test/benchmark generation
- [  ] Documentation (including examples)
- [x ] Research and understanding
- [  ] No AI tools were used

> Regardless of AI use, the human contributor remains fully responsible for correctness, design choices, licensing compatibility, and long-term maintainability.

### DIFF
--- a/changelog/950.feature.rst
+++ b/changelog/950.feature.rst
@@ -1,0 +1,1 @@
+Add type and type checking to NDCube
--- a/changelog/950.feature.rst
+++ b/changelog/950.feature.rst
@@ -1,1 +1,1 @@
-Add type and type checking to NDCube
+Add type and type checking to package
--- a/ndcube/extra_coords/extra_coords.py
+++ b/ndcube/extra_coords/extra_coords.py
@@ -1,5 +1,5 @@
 import abc
-from typing import TYPE_CHECKING, Any
+from typing import TYPE_CHECKING, Any, TypeAlias
 from numbers import Integral
 from functools import reduce, partial
 from collections.abc import Iterable, Sequence
@@ -30,6 +30,14 @@ if TYPE_CHECKING:
 
 __all__ = ['ExtraCoords', 'ExtraCoordsABC']
 
+# The set of types ExtraCoords.add actually accepts, per its exhaustive
+# isinstance/raise chain: a pre-built BaseTableCoordinate, or anything it
+# knows how to build one from (Time, SkyCoord, a single Quantity, or a
+# sequence of 1-D Quantities for a multi-dimensional coordinate).
+LookupTableInput: TypeAlias = (
+    BaseTableCoordinate | Time | SkyCoord | u.Quantity | list[u.Quantity] | tuple[u.Quantity, ...]
+)
+
 
 class ExtraCoordsABC(abc.ABC):
     """
@@ -53,7 +61,7 @@ class ExtraCoordsABC(abc.ABC):
     def add(self,
             name: NamesType,
             array_dimension: int | Iterable[int],
-            lookup_table: Any,
+            lookup_table: LookupTableInput,
             physical_types: NamesType = None,
             **kwargs: Any) -> None:
         """
@@ -134,7 +142,7 @@ class ExtraCoordsABC(abc.ABC):
 
     @abc.abstractmethod
     def resample(self, factor: float | Iterable[float], offset: float | Iterable[float] = 0,
-                ndcube: Any = None, **kwargs: Any) -> "ExtraCoordsABC":
+                ndcube: "NDCubeABC | None" = None, **kwargs: Any) -> "ExtraCoordsABC":
         """
         Resample all extra coords by given factors in array-index-space.
 
@@ -170,16 +178,16 @@ class ExtraCoords(ExtraCoordsABC):
 
         # Lookup tables is a list of (pixel_dim, LookupTableCoord) to allow for
         # one pixel dimension having more than one lookup coord.
-        self._lookup_tables: list[tuple[Any, BaseTableCoordinate]] = []
+        self._lookup_tables: list[tuple[int | Iterable[int], BaseTableCoordinate]] = []
         self._dropped_tables: list[BaseTableCoordinate] = []
 
         # We need a reference to the parent NDCube
         self._ndcube = ndcube
 
     @classmethod
-    def from_lookup_tables(cls, names: Sequence[str], pixel_dimensions: Sequence[Any],
-                           lookup_tables: Sequence[Any],
-                           physical_types: Sequence[Any] | None = None) -> "ExtraCoords":
+    def from_lookup_tables(cls, names: Sequence[str], pixel_dimensions: Sequence[int | Iterable[int]],
+                           lookup_tables: Sequence[LookupTableInput],
+                           physical_types: Sequence[NamesType] | None = None) -> "ExtraCoords":
         """
         Construct a new ExtraCoords instance from lookup tables.
 
@@ -227,7 +235,7 @@ class ExtraCoords(ExtraCoordsABC):
 
         return extra_coords
 
-    def add(self, name: NamesType, array_dimension: int | Iterable[int], lookup_table: Any,
+    def add(self, name: NamesType, array_dimension: int | Iterable[int], lookup_table: LookupTableInput,
             physical_types: NamesType = None, **kwargs: Any) -> None:
         # docstring in ABC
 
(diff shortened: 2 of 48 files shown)


samaloney:
Not sure how to resolve the doc build issue I think its because there is more then one `slice` object so it can be uniquely resolved?

ndcube/extra_coords/extra_coords.py
```diff
@@ -521,16 +549,16 @@ def cube_wcs(self):
...
+    def _cube_array_axes_without_extra_coords(self) -> set[int]:
         """Return the array axes not associated with any extra coord."""
-        return set(range(len(self._ndcube.shape))) - set(self.mapping)
+        return set(range(len(self._ndcube.shape))) - set(self.mapping)  # type: ignore[union-attr]
 
-    def __str__(self):
+    def __str__(self) -> str:
         classname = self.__class__.__name__
-        elements = [f"{', '.join(table.names)} ({axes}) {table.physical_types}: {table}"
+        elements = [f"{', '.join(table.names or ())} ({axes}) {table.physical_types}: {table}"
```
DanRyanIrish:
```suggestion
        elements = [f"{', '.join(table.names) if table.names else ()} ({axes}) {table.physical_types}: {table}"
```

changelog/950.feature.rst
```diff
@@ -0,0 +1 @@
+Add type and type checking to NDCube
```
DanRyanIrish:
```suggestion
Add type and type checking to package
```
This PR had greater scope than just the NDCube class, which is great!

ndcube/extra_coords/extra_coords.py
```diff
@@ -120,6 +132,15 @@ def __getitem__(self, item: str | int | slice | Iterable[str | int | slice]) ->
         not supported.
         """
 
+    @abc.abstractmethod
+    def resample(self, factor: float | Iterable[float], offset: float | Iterable[float] = 0,
```
DanRyanIrish, line 144: I don't think we've decided that resample should live in the ABC.  My initial reaction is that I'm not against it.  But we shouldn't add this without making an active decision.

ndcube/tests/test_ndcollection.py
```diff
@@ -112,13 +112,45 @@ def test_collection_pop(collection, popped_key, expected_popped, expected_collec
 
 @pytest.mark.parametrize(('collection', 'key', 'expected'), [
     (cube_collection, "cube0", NDCollection([("cube1", cube1), ("cube2", cube2)],
-                                            aligned_axes=aligned_axes[1:]))])
+                                            aligned_axes=aligned_axes[1:])),
+    # Regression test: __delitem__ used to call self.aligned_axes.__delitem__(key)
+    # unconditionally, crashing with AttributeError on a collection with no
+    # aligned axes since self.aligned_axes is None in that case.
+    (unaligned_collection, "cube0", NDCollection([("cube1", cube1), ("cube2", cube2)]))])
```
DanRyanIrish: Would it not be more readable to have it this way around?
```suggestion
    (unaligned_collection, "cube0", NDCollection([("cube1", cube1), ("cube2", cube2)]))])
    # Last regression test: __delitem__ used to call self.aligned_axes.__delitem__(key)
    # unconditionally, crashing with AttributeError on a collection with no
    # aligned axes since self.aligned_axes is None in that case.
```

ndcube/tests/test_ndcollection.py
```diff
@@ -112,13 +112,45 @@ def test_collection_pop(collection, popped_key, expected_popped, expected_collec
...
 def test_del_collection(collection, key, expected):
     del_collection = collection.copy()
     del del_collection[key]
     helpers.assert_collections_equal(del_collection, expected)
 
 
+def test_slice_by_keys_unaligned_collection():
```
DanRyanIrish, line 132: Can you explain what this test is doing?  Is doesn't appear to be testing slicing as its comments suggest.

ndcube/ndcollection.py
```diff
@@ -49,7 +50,7 @@ class NDCollection(dict):
     axis 1 of cube0 is aligned with axis 1 of cube1.
     """
 
-    def __init__(self, key_data_pairs, aligned_axes=None, meta=None, **kwargs):
+    def __init__(self, key_data_pairs: Any, aligned_axes: Any = None, meta: Any = None, **kwargs: Any) -> None:
```
DanRyanIrish: Should `key_data_pairs` not be `Iterable[Any, Any]`? Or whatever the correct equivalent is?

ndcube/ndcollection.py
```diff
@@ -64,11 +65,12 @@ def __init__(self, key_data_pairs, aligned_axes=None, meta=None, **kwargs):
 
         # Convert aligned axes to required format.
         sanitize_inputs = kwargs.pop("sanitize_inputs", True)
+        keys: tuple[Any, ...] = ()
```
DanRyanIrish, line 78: Why is this needed?  `keys` is defined a couple line below.  And `key_data_pairs` must contain at least one valid key so `keys` cannot be empty.  Seems to me like it isn't needed.  But am I missing something?

ndcube/ndcollection.py
```diff
@@ -80,20 +82,20 @@ def __init__(self, key_data_pairs, aligned_axes=None, meta=None, **kwargs):
         if self._aligned_axes is None:
             self.n_aligned_axes = 0
         else:
-            self.n_aligned_axes = len(self.aligned_axes[keys[0]])
+            self.n_aligned_axes = len(self._aligned_axes[keys[0]])
 
     @property
-    def aligned_axes(self):
+    def aligned_axes(self) -> dict[Any, tuple[Any, ...]] | None:
```
DanRyanIrish: This should return `None` or a `dict` where the value assigned to each key is `int` or `tuple[int]`.  Can this be captured in typing here?

DanRyanIrish (changes requested):
More comments to follow

ndcube/extra_coords/extra_coords.py
samaloney: Ok yea can pull it from this PR and update the type hints if needed

ndcube/ndcollection.py
samaloney: Yea I think `dict[Any, int] | `dict[Any, tuple[int, ...]] | None` 

Also where typing could potentially help design - as consumer of this I need to check if the value is an int or tuple then maybe the length of the tuple but if it was always returned a tuple `(), (1,), (1, 2, 3) can always use `len` - not suggestion this should be changed by the way just a thought.

ndcube/tests/test_ndcollection.py
samaloney: Yea so `unaligned_collection` looks like this and then it sliced with `unaligned_collection[("cube0", "cube1")]`  
```
<ndcube.ndcollection.NDCollection object at 0x10ffbc4d0>
NDCollection
------------
Cube keys: ('cube0', 'cube1', 'cube2')
Number of Cubes: 3
Aligned dimensions: None
Aligned physical types: None

ndcube/ndcollection.py
samaloney: Yep so this should accept a dict of type `dict[str, NDCube|NDSequence]` or tuple on the tuple is is a tuple of tuple e.g `tuple[tuple[str, NDCube|NDSequence]` or a long list of key, data pair e.g `tuple[any, ...]`. I see if I can worked it out from the code.

ndcube/ndcollection.py
samaloney: So this is because pyright raises and possibly unbound warning because `aligned_axes` is `None` the variable isn't defined - could add local comment disable that warning

ndcube/ndcollection.py
samaloney: So if I change the type annotation to it raises an additional type errors as in the current code later on this properties value is subscripted so if were an int this would raise an error.

https://github.com/sunpy/ndcube/blob/9befb9b43c8829c8f341f23a91e84255d94f9384/ndcube/ndcollection.py#L229-L255

Internally its always converted to a tuple the only place I can see where it can be an int is in `_sanitize_user_aligned_axes` also `tuple(int, ...)` cover

